Issue #1085, opened 2014-10-09
VAM CP x-axis bug


matthieudelcourt:
The CP VAM y-axis goes up to 7,000+ m/hour, it could be rescaled to the actual data :-)

PR #1135, opened 2014-11-16
VAM CP x-axis bug #1085

Joern-R:
... fixed by not considering Y-Axis values in the 0 to 4.993 minutes range which are not shown for VAM anyway
... both in standard view and in compare view / intervals and date ranges
... handle intervals below 4.993 minutes properly (since they are not shown in CP VAM)


matthieudelcourt:
Hi, I downloaded the GC version 3.4 (the latest available) and it seems that the problem is back. I don't see the x-axis before 5min on the Critical VAM. Also, comparing the data to Strava it seems that the values are quite off (100m/h to 200m/h difference). I'm happy to give you an example on my own data.
Cheers :-)
Matt

PR #1220, opened 2015-01-05
Train View - add TE/PS/LR Balance - Real Time Metrics

Joern-R:
... Torque Effectiveness
... Pedal Smoothness
... L/R Balance
(mainly for Vector - but implemented following standard ANT+ message rules)


liversedge:
That was quick !

PR #1224, opened 2015-01-06
Time in pace zone metrics for swim activities

amtriathlon:
Shown in activity summary
computed if there are swim pace zones defined and the activity isSwim
isSwim() looks just at Sport="Swim" for now

### DIFF
--- a/src/Athlete.cpp
+++ b/src/Athlete.cpp
@@ -110,12 +110,14 @@ Athlete::Athlete(Context *context, const QDir &homeDir)
             QMessageBox::warning(context->mainWindow, tr("Reading HR Zones File"), hrzones_->warningString());
     }
 
-    // Pace Zones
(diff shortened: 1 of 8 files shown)


liversedge:
isSwim needs to be a first class variable in RideItem and stored in RideDB.json so it can be used in filters  etc. I can look at that for you :)

amtriathlon:
Excellent, thanks!

2015-01-07 4:28 GMT-03:00 Mark Liversedge <notifications@github.com>:

> isSwim needs to be a first class variable in RideItem and stored in
> RideDB.json so it can be used in filters etc. I can look at that for you :)
>
> —
> <https://github.com/GoldenCheetah/GoldenCheetah/pull/1224#issuecomment-68987532>
> .
>



-- 
Alejandro.

PR #1232, opened 2015-01-10
Fixed NaN in xPower when there are no samples

amtriathlon:
It propagates to Bike/SwimScore and plays havoc with PMC


liversedge:
Ha! I /literally/ just fixed in a different way (cleaning any Nan/Inf values after they are computed).
But happy to apply this too :)

liversedge:
If you have values cached in rideDB.json you might need to delete it !

PR #1238, opened 2015-01-11
Changes to build on OS X with spaces in path

WillerZ:
Needed to escape the spaces in $$(OUT_PWD) and $$quote(…) did not work.

Also qmake failed to complete initially because the translation files I have now removed from src/application.qrc are not present.

### DIFF
--- a/qwt/src/src.pro
+++ b/qwt/src/src.pro
@@ -17,7 +17,7 @@ include( $${QWT_ROOT}/qwtfunctions.pri )
 TEMPLATE          = lib
 TARGET            = $$qwtLibraryTarget(qwt)
 
-DESTDIR           = $${OUT_PWD}/../lib
+DESTDIR           = $$replace($${OUT_PWD}/../lib," ","\\ ")
 
 contains(QWT_CONFIG, QwtDll) {
 
--- a/src/application.qrc
+++ b/src/application.qrc
@@ -75,15 +75,6 @@
         <file>images/settings.png</file>
         <file>images/addchart.png</file>
         <file>images/library.png</file>
-        <file>translations/gc_fr.qm</file>
-        <file>translations/gc_ja.qm</file>
-        <file>translations/gc_it.qm</file>
-        <file>translations/gc_pt-br.qm</file>
-        <file>translations/gc_de.qm</file>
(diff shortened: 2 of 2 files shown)


liversedge:
You need to install the QT translation tools or point to lrelease in gcconfig.pri, there is a bit of history behind the way we default to qt4 versions that I must confess I cannot for the life of me remember.

But the replace " " thing is rather ingenious. 
I can't merge with the translation files change, but if you tidy it up I can :)

liversedge:
screenshot from 2015-01-11 16 35 17

WillerZ:
@liversedge fixt

liversedge:
That was quick !
Thanks

Joern-R:
Hi Phil,

sorry - I removed the $$replace as it created problems both on Mac and Windows if the path name contained upper case - see https://github.com/GoldenCheetah/GoldenCheetah/issues/1247

I added some debug messages to src.pro and got this output.

message ("OLD")
message($${OUT_PWD}/../lib)
message ("NEW")
message($$replace($${OUT_PWD}/../lib," ","\\ "))
message ("NEW with lower case conversion")
new = $$lower($${OUT_PWD}/../lib)
message($$replace(new," ","\\ "))

Output was:
>OLD
> E:/50_Dokumente/GitHub/Debug-64Bit-5.4.0/qwt/src/../lib
> NEW
> 
> NEW with lower case conversion
> e:/50_dokumente/github/debug-64bit-5.4.0/qwt/src/../lib

So the replace has problems with " " when text not in "lower case". With this DESTDIR was just empty and GC could not be build as the linker did not find the qwtlib.

I searched for a solution on that - but just found a thread in the QT mailing list describing the replace with space / upper/lower case solution. But not specifically for directory pathes and other QT variables.

When trying to convert the path to lowercase first and the apply $$replace, QMake generated absolute pathes into the 

- object_script.libqwt.Release 
- object_script.libqwt.Debug 

files to build the libs - and not relative any more. With this GC src.pro does not work any more as it points to the same release lib in both debug and release builds of GC.

If you have a solution how to overcome the $$replace problem - happy to test it on windows. Just push a commit and I will cross-check.

Joern

PR #1239, opened 2015-01-11
Fix - .JSON not updated with man.l changes directly after Import or Download

Joern-R:
... correct the path in RideItem after updating RideCache to the final
path (since RideItem has a local buffer for the path - which is e.g.
used at "Save Ride") any manual changes after the import will create a
new .JSON in /tmpActivities and not update the original .JSON in
/activities

(Problem only occurs when using the in-memory data after an Import or
Download)

### DIFF
--- a/src/DownloadRideDialog.cpp
+++ b/src/DownloadRideDialog.cpp
@@ -25,6 +25,7 @@
 #include "Settings.h"
 #include "JsonRideFile.h"
 #include "HelpWhatsThis.h"
+#include "RideItem.h"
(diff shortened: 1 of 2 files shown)


liversedge:
I literally JUST noticed this bug with my ride from this afternoon.
You're too fast !!!

:)

Joern-R:
I imported today's ride just a little earlier - (and had caused the problem) - wasn't aware that we still have pathes stored/used outside the new directory structure object.

PR #1244, opened 2015-01-13
LTM - Curves Settings - Add Resorting of Curves

Joern-R:
... add "Up" - "Down" function for Curves Table Items in LTM Settings

(particular useful when adding a curve in Stacked View and wanting to have
that curve shown at first position)


liversedge:
NICE!

PR #1252, opened 2015-01-16
Add "GC_HAS_INTERVALS" for Route Window

Joern-R:
... do not allow Route Window if not demanded


Joern-R:
Thanks - good to know - I tried to add a "Route Segment" chart yesterday - but since all the of the code is not available it directly end with a dump.

PR #1273, opened 2015-01-23
Fixed Speed units in manual workouts and Pace units in Ride Summary

amtriathlon:
Ride summary is ugly, it should be simpler with toString method but it would need a context (or similar) parameter.


liversedge:
I think we need to move away from html and towards a more 'interactive' widget .. it has got to the point where its just too complicated to maintain -- and the original idea was to use html to make things easier !

Maybe we can fix this in 4.0

PR #1301, opened 2015-02-04
Simplified IntervalSummaryWindow removing Pace special cases

amtriathlon:
toString FTW ;-)


liversedge:
cool!

amtriathlon:
Mark, please check the second commit, the change in RideSummaryWindow is straightforward but I am no so sure about RideCache modification. It is working in my tests but it could have side effects which I don't know.

liversedge:
its fine.

PR #1303, opened 2015-02-05
Use Duration when Time Moving is zero in Session RPE and Daniels EqP

### LINKED ISSUE #1265: Daniels EqP zero when time riding is zero
When traning on a stationary trainer with GPS off in my Garmin Edge 500, I get time riding 00:00 on all my workouts (only the duration "works"), and I guess that's because I'm not "moving". Avg power, NP works fine, but Daniels EqP just becomes 0 on such rides.

 Is it a bug or can't Daniels EqP be calculated without a time riding value? (which I guess is the "problem")

amtriathlon:
Important for Session RPE, similar to TRIMP points, also fixes #1265


liversedge:
good idea !

PR #1315, opened 2015-02-12
Release 3.1.0

DirkVC:
Add max HR, max Speed and max. Cadence in the manual entry of the workout


Joern-R:
The pull request only contains "old" - already merged commits. Please remove and create a new clean pull request with your suggested enhancements.

Thanks.

liversedge:
The manual workout is overloaded, adding hardly used metrics is not a great idea.
We would do better to add a ridemetadata page that has the "details" screen .. or similar.

PR #1327, opened 2015-02-21
Error handling - if XML file is "write-protected"

Joern-R:
... report error in case any of the "*-layout.xml" files cannot be written to (e.g. due to write protection)

### DIFF
--- a/src/HomeWindow.cpp
+++ b/src/HomeWindow.cpp
@@ -1170,7 +1170,10 @@ HomeWindow::saveState()
 
     QString filename = context->athlete->home->config().canonicalPath() + "/" + name + "-layout.xml";
     QFile file(filename);
-    file.open(QFile::WriteOnly);
+    if (!file.open(QFile::WriteOnly)) {
+        QMessageBox::critical(this, tr("Problem Saving Charts Configuration"), tr("File: %1 cannot be opened for 'Writing' ! Please check file properties.").arg(filename));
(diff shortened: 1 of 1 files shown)


liversedge:
This will throw a dialog for each view so there will be 4 dialogs on exit .. is that a good idea?
But then, it shouldn't fail to write... same applies for all the config files though?
I wonder if a warning should be issued after checking if any/all the config files are writable instead?

Joern-R:
This should be a very exceptional case - but it recently happened to a user in German Forum - who was noticing that his Chart changes were not stored. After having opened the file with Notepad++ and saving it - all was fine.

That's why I proposed the "simple" way. Expecting only very few occurences - but if it happens - users need to investigate the root-cause in their file system.

"Charts.XML", "Autoimport.xml", and some more  have the same concept to write the file - so adding the same check for these will make sense. Adding huge logic to check first, collect the problems and through a combined message I consider to much effort.

Yes - there will be a lot of message if e.g. the whole directory is "read only" - just until the User fixes the problem.

So - if you agree - I would just add a similar check to the other ".XML" config writing locations to be consistent.

liversedge:
I think you are right, its an edge case, so this is good enough !

Joern-R:
Ok - and I will add it to all other places where Config is Written with the same fileOpen logic. So that we help users to determine such problems.

PR #1355, opened 2015-04-04
Adding a class that enables indending QwtPlotMarkers.

Deadolus:
This commit is adding a new class QwtIndPlotMarker that will enable us
to indend Labels. The usage is pretty much the same,
please look at the changes in LTMPlot.cpp to see what will
change for using the new class.
So far the functionality is implemented for horizontally written labels only.

This solves the issue of all Labels (e.g. Seasons and events in LTMPlot case)
being on one line and thus being unreadable.
According to [https://groups.google.com/forum/#!topic/golden-cheetah-users/-v6lhGZ5B-o](https://groups.google.com/forum/#!topic/golden-cheetah-users/-v6lhGZ5B-o)

Signed-off-by: Simon Egli (Silzo) <smn.egli@gmail.com>

### DIFF
--- a/src/IndendPlotMarker.cpp
+++ b/src/IndendPlotMarker.cpp
@@ -1,0 +1,437 @@
+
+/* -*- mode: C++ ; c-file-style: "stroustrup" -*- *****************************
+ * Qwt Widget Library
+ * Copyright (C) 1997   Josef Wilgen
+ * Copyright (C) 2002   Uwe Rathmann
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the Qwt License, Version 1.0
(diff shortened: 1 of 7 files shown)


liversedge:
Nice job ! Would be interesting to apply similar to AllPlot.

Deadolus:
Hi Mark, thanks for pulling. 
I wanted to push this one out as soon as possible - as I announced it already a long time ago. 
I will have a look at AllPlot (having a bit of time over easter) and apply the same. 
If anyone wants to do it themselves - you're welcome to beat me to it. :)
Also: there might be even more plots where we might use the new class.

Cheers, 

Simon

PR #1356, opened 2015-04-05
Adding Indended Plot Markers for AllPlot

Deadolus:
I intentionally didn't add the IndendPlotMarkers to all markings,
but only to the interval markers at the top of the plot.

NOTE: the height of the spacing is actually what Qwt tells us
is the textheight.
I think this spacing is rather big.
We might want to reduce it somehow in the future.

Signed-off-by: Simon Egli (Silzo) <smn.egli@gmail.com>

### DIFF
--- a/src/AllPlot.cpp
+++ b/src/AllPlot.cpp
@@ -32,6 +32,7 @@
 #include "Zones.h"
 #include "Colors.h"
 #include "WPrime.h"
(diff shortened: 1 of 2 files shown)


liversedge:
Blimey that looked easy !
Am pulling to test and play, can't see any problems in the code, the changes are so trivial.

Nice job again.

PR #1367, opened 2015-05-10
Added Peak Pace metrics for running and swimming

amtriathlon:
Durations are the same as for power except for 1 sec and 5 sec.


liversedge:
Hi Ale,

I will add peak pace auto discovery too; applies for swimming and running only.
Using the same durations.

Cheers
Mark

PR #1392, opened 2015-06-03
Discover Peak Pace intervals for running and swimming

amtriathlon:


liversedge:
Thanks Ale.

Any thoughts on identifying sustained efforts for running / swimming as we do for cycling power ?

amtriathlon:
Mark, at first glance it seems we could use the same scheme with CV&D' for
sustained efforts and TTE, I will try it without the searching for sprints.

2015-06-03 12:15 GMT-03:00 Mark Liversedge <notifications@github.com>:

> Thanks Ale.
>
> Any thoughts on identifying sustained efforts for running / swimming as we
> do for cycling power ?
>
> —
> <https://github.com/GoldenCheetah/GoldenCheetah/pull/1392#issuecomment-108479160>
> .
>



-- 
Alejandro.

PR #1397, opened 2015-06-08
Import multisport FIT files

alex-hhh:
Hi Mark,

I added support for importing multi-sport FIT files, as generated by Garmin Foreruner devices.  This works by importing each session as a separate activity (Ride).

Can you please have a look at the changes and let me know if you are interested in including them into the code base?  I am happy to make any changes you think are necessary (I'm not familiar with the GC contribution policy).

Thanks,
Alex.

### DIFF
--- a/src/FitRideFile.cpp
+++ b/src/FitRideFile.cpp
@@ -76,6 +76,18 @@ struct FitFileReaderState
     QFile &file;
     QStringList &errors;
     RideFile *rideFile;
+
+    // List of rides that we imported. Most FIT files will have one ride, but
+    // multisport ones will have multiple rides, one for each sport.
+    //
+    // NOTE: this is an output parameter received from
+    // FitFileReader::openRideFile(), and may be NULL.  In that case, the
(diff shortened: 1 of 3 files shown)


liversedge:
Hey Alex,

Thank you for keeping on this, of course I am very interested to include it, and will take a look.
I wonder how many folks record data this way?
Is it something you do regularly ?

alex-hhh:
Hi Mark,

This is how I record triathlons and some of the brick workouts (bike + transition + run).  Last season I had 5 triathlon races and about 10 brick workouts.  In general during a race it is more convenient to use this feature, since advancing from one activity to the next is done by pressing the "Lap" button instead of having to reset the watch and start a new activity.

This has been a feature on Garmin Forerunners since at least the 310XT, I suspect a lot of people know and use it.

If you would like to test the feature, I'm happy to send you a multi-sport FIT file.

Cheers,
Alex.

liversedge:
Yes please. It sounds like we need some cool way of associating the separate sections into a single 'race' or 'workout' in the future !

PR #1400, opened 2015-06-13
Set Pace Zone Shading and Pace units according to sport in CV chart

amtriathlon:
In activities it depends on selected activity, bests are automatically filtered.
In range mode it can be set by sidebar or home filters for single sport

a) Examples in Trends:
a1) Swims:
image
a2) No sport filter (mixed activities or only rides):
image
a3) Runs:
image
b) Examples in Activities:
b1) Swim selected:
image
b2) Ride selected:
image
b3) Run selected:
image

### DIFF
--- a/src/CPPlot.cpp
+++ b/src/CPPlot.cpp
@@ -57,7 +57,8 @@ CPPlot::CPPlot(QWidget *parent, Context *context, bool rangemode) : QwtPlot(pare
     model(0), modelVariant(0),
 
     // state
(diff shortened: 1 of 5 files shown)


liversedge:
We may be overusing the isRun/isSwim concept in place of a straight 'Sport' code?
Given we are unlikely to have other sports for a long time this is no biggie though !

Thanks for continuing to develop the multisport stuff, its really really appreciated :+1:

amtriathlon:
Agree, I would prefer Sport code if starting again, but we are in the middle of the river, or even near the other bank!

PR #1405, opened 2015-06-16
Updated Russian translation

prokoudine:
This is ca. +1500 translated messages. There's still a lot of work to do, but GC should look a lot more usable to bolshevik triathletes now :)

There's a number of places where having proper plural forms implementation (https://doc.qt.io/archives/qq/qq19-plurals.html) would benefit users. I guess I could patch those (I only can do my best, not being a real programmer), but I have no idea if you have any sort of a strings freeze practice, and whether one is in effect currently.

Also, now that GoldenCheetah is multisport software, referring to every kind of an activity as a 'Ride' looks a bit awkward. Perhaps you might want a patch for that as well that would use a less sport-specific term such as 'Activity'?

### DIFF
--- a/src/translations/gc_ru.ts
+++ b/src/translations/gc_ru.ts
@@ -1,18 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="ru">
+<TS version="2.0" language="ru">
 <context>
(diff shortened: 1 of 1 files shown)


liversedge:
Thanks.

I thought I had eradicated 'ride' in favour or 'activity' ...

PR #1409, opened 2015-06-18
Added error message for unsupported Lap Swimming FIT files

amtriathlon:
Until proper handling is added better to give an error message,
otherwise an activity with no data is created causing confusion.

I think it is a good idea to try to prevent a flow error reports due to this.

### DIFF
--- a/src/FitRideFile.cpp
+++ b/src/FitRideFile.cpp
@@ -960,6 +960,11 @@ struct FitFileReaderState
                 case 23: /*decodeDeviceInfo(def, time_offset, values);*/ break; /* device info */
                 case 18: decodeSession(def, time_offset, values); break; /* session */
 
+                case 101: /* lap swimming length */
+                    errors << "Unsupported Lap Swimming FIT File - Use .tcx or .pwx formats";
+                    stop = true;
+                    break;
+
                 case 2: /* DEVICE_SETTINGS */
                 case 3: /* USER_PROFILE */
                 case 7: /* ZONES_TARGET12 */


liversedge:
How should these files be handled ?

amtriathlon:
IMO in the same way we currently handle TCX and PWX lap swimming files: expanding each length record and rest interval to 1 per sec samples with the same speed and cadence (0 speed and cadence for pauses).

liversedge:
That sounds quite easy to achieve ?
Have I missed something ?

amtriathlon:
>That sounds quite easy to achieve ?
Yes, it does and it was the case for tcx and pwx formats.
>Have I missed something ?
I don't know, I didn't looked to much at the code nor the protocol yet,
since there was another developer working on this feature.

glonnon:
I just ran into this issue.

Is anyone activitly working on this issue?  If not, I am willing to
implement this feature.

Thanks,
Greg

liversedge:
Hey Greg,

You are _always_ welcome, how long you reckon it will take?
If we can sneak a fix through before Stage 21 of the Tour that'd be neat ...............

Mark

amtriathlon:
Not that I know Greg, please, go ahead!

2015-07-05 16:18 GMT-03:00 Mark Liversedge <notifications@github.com>:

> Hey Greg,
>
> You are *always* welcome, how long you reckon it will take?
> If we can sneak a fix through before Stage 21 of the Tour that'd be neat
> ...............
>
> Mark
>
> —
> <https://github.com/GoldenCheetah/GoldenCheetah/pull/1409#issuecomment-118653264>
> .
>



-- 
Alejandro.

PR #1422, opened 2015-06-27
Small fix for C -> F temperature conversion

dresco:
Looks like the metric to imperial conversion was missing one step..


liversedge:
thanks !

PR #1435, opened 2015-07-07
Fixed default CV value when pace units change

amtriathlon:
There was a commentary indicating this was pending that I was not aware.


liversedge:
Will test the interval editing, not really sure at present if this is good or not !

PR #1442, opened 2015-07-10
Minor Fixes WPrime::summarize units and Spanish Translation

amtriathlon:


liversedge:
Oops. Good spot on the units.

PR #1450, opened 2015-07-15
Minor Translation Related Fixes

amtriathlon:
"Metric Pace" translatable
Format for time columns in Navigator

Actually sorting by time columns doesn't work because hours are absent when value < 1h, perhaps it can be temporarily solved formating time as hh:mm:ss even when it is < 1h


liversedge:
I think formatting by hh:mm:ss is a smart move!

amtriathlon:
On second thought, It would probably be better to return a QTime...

PR #1451, opened 2015-07-15
Fixes ordering of time columns in Navigator

amtriathlon:
Using a QTime for isTime metrics

Navigator works as expected, but I don't know if this change may have unwanted side effects. The alternative would be to format a HH:MM:SS string.

### DIFF
--- a/src/RideCacheModel.cpp
+++ b/src/RideCacheModel.cpp
@@ -88,10 +88,13 @@ RideCacheModel::data(const QModelIndex &index, int role) const
                 // version using the right metric/imperial conversion
                 RideMetric *m = const_cast<RideMetric*>(factory->rideMetric(factory->metricName(i)));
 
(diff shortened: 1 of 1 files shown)


liversedge:
Nice fix .. but yeah, not sure either. Just don't ride for more that 24 hrs!

amtriathlon:
OTOH the string method should work up to 100hs but I'm not sure if it is
worth...


2015-07-15 16:19 GMT-03:00 Mark Liversedge <notifications@github.com>:

> Nice fix .. but yeah, not sure either. Just don't ride for more that 24
> hrs!
>
> —
> <https://github.com/GoldenCheetah/GoldenCheetah/pull/1451#issuecomment-121719347>
> .
>



-- 
Alejandro.

liversedge:
completely agree, that's one hell of an edge case !

PR #1457, opened 2015-07-20
Updated Metrics translation in LTM

amtriathlon:
- Code cleanup, removing deprecated code and refactoring
- Metrics translation is now done in a method for LTMSettings
- User defined metric name (uname) is translated if it matches english name
- User defined units (uunit) is only translated if it matches saved units


liversedge:
Thanks!

PR #1467, opened 2015-07-24
Adjust Video Playback speed on cyclist speed

vlcvboyer:
This patch set adjusts Video Playback speed on cyclist speed in training mode with VLC.
Needs GPX file to be used (same way than street view)
Valid when using libVLC 
see feature 382 linked to this amendment:
https://groups.google.com/forum/#!topic/golden-cheetah-developers/t9JFkSFLuv8


liversedge:
Am waiting till v3.2 is released to pull this into master.

Thanks for the patch !

vlcvboyer:
Hi !
I'm pleased to see that my patch is appreciated !
I have few question in regard with training feature:
Is there any plan to change training view, etc. ?
I saw that a dedicated training program was under progress but as far as I see it is empty since long time.
At present my vision of it is that it's not as intuitive as it could be (for example streetview needs to select existing GPX in activities list before going to training view... In addition we have to import an activity which may not be ours but will be taken into account for statistics, we cannot name it, etc. )
It will be very interesting to be able to use gpx directly and controling trainer dynamic brake based on GPX elevation + wind heading computation.
I'll be pleased to contribute much more but I don't want to write some code which will not match with future versions or other in-progress projects.

liversedge:
The training view doesn't get enough attention, the problem is there is a very small number of users and there is a lot of new ideas and interest in the other aspects.

It has received a fair amount of updates in 3.2 W'bal as you ride, vector support, moxy and Tacx support was added.

There are a LOT of updates pending for v4.0 too - check out the github issues tagged "train view".

You are more than welcome to start working on new ideas -- the UX is terrible and anything is likely to be an improvement .. do check it "fits in with general direction" before starting, but I have quite an open mind on this.

If you want to look at decoupling it from the main app and getting "OpenTrainer" started you are very welcome to have a go....

vlcvboyer:
I've closed this pul request as I've created a mess on my github fork when syncing with goldenCheetah master...
Cleanning is in progress...

PR #1477, opened 2015-07-30
Poor man's fixes for plural forms + updated Russian translation

prokoudine:
Reasoning behind changes: http://doc.qt.io/qt-4.8/i18n-source-translation.html

I'm not really a programmer, and I feel that this could be done more elegantly, but it seems to work for me.

### DIFF
--- a/src/IntervalSummaryWindow.cpp
+++ b/src/IntervalSummaryWindow.cpp
@@ -238,7 +238,8 @@ QString IntervalSummaryWindow::summary(QList<IntervalItem*> intervals, QString &
             temp.metrics()[j] = 0.00f;
 
     // set name
-    temp.name = QString(tr("%1 selected intervals")).arg(intervals.count());
+    int n = intervals.count();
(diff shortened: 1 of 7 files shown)


amtriathlon:
I think it is an interesting improvement, but perhaps would be better to review this after v3.2 release since applying it now would invalidate existing translations, some of which are already finished.
Is this strictly necessary for the Russian translation or just an enhancement?

prokoudine:
"Is this strictly necessary for the Russian translation or just an enhancement?" — well, without the patch part of the translation look like a half-assed job done in terms of grammar :)

liversedge:
I've put into the 4.0 queue since its really not that important.

prokoudine:
Should I re-update my next translation based on upstream master branch then?

liversedge:
If you have combined a translation update with a small code fix then sadly yes.
Its better to keep things like that separate :)

Thanks !

PR #1496, opened 2015-08-06
Make default PMC TRIMP chart to real TRIMP

MaksVasilev:
Now it's use only 2 TRIMP (HR based) metrics and 1 Skiba (Power based) metric. This is incorrect.


liversedge:
BTW, thanks for the fix !

PR #1498, opened 2015-08-06
Fixed translate charts when config file is missing

amtriathlon:
Default config due to new user or deleted config should be translated
Updated chart titles according to charts.xml for translation purposes
Updated Spanish Translation


amtriathlon:
Agree, besides we are not translating user defined metric names and units, which should be even more important when we have user defined formulas... Perhaps we could add language dependent repositories for localized versions so users can contribute also the translations, and the option for non-english to access the localized repository or the english one, or just select any available repository for everyone...

PR #1501, opened 2015-08-08
Added Sigma Rox 10.0 .slf file import support (it may also work with …

cgiraldo:
…Rox 9.0 but it is not tested).

Supported data:
 - speed, cadence, hearthrate, power, latitude, longitude, altitude, temperature.

Laps are not yet supported. (laps in Rox 10.0 slf are encoded at the end of the file as <Markers>).


liversedge:
Thats great, thank you.
Do you have a test file for the test/rides directory ?

cgiraldo:
I uploaded the test file in my fork.

PR #1510, opened 2015-08-12
Updated Russian translation

prokoudine:
Hopefully good enough for v3.2, although fixes could follow.


liversedge:
Other fixes will need to go into v4.0, we are releasing ! :)

PR #1517, opened 2015-08-17
Fixed POW precedence and dangling else ambiguity

amtriathlon:
Now 2*2^3+1 gives 17 instead of 65
expr1 ? expr2 ? expr3 : expr 4 : expr 5 is disambiguated to expr1 ? (expr2 ? expr3 : expr 4) : expr 5
Lowers shift/reduce conflitcs to 21


liversedge:
thanks.. just also pushed a fix for a terrible bug that broke config(), const(), best() et al.
It was introduced by the daterange() commit.

PR #1518, opened 2015-08-18
Fixed comparison operator precedence

amtriathlon:
Adding a comparison expression cexpr and separate productions
for each operator allows to define precedence rules to remove
all shift-reduce conflicts, except one due to the -probably
harmless- (expr) and (lexpr) ambiguity.

It seems to work, but please review and test before applying to be sure it dosen't broke anything previously working!

### DIFF
--- a/src/DataFilter.y
+++ b/src/DataFilter.y
@@ -60,7 +60,7 @@ extern Leaf *root; // root node for parsed statement
 %token <op> EQ NEQ LT LTE GT GTE
 %token <op> ADD SUBTRACT DIVIDE MULTIPLY POW
 %token <op> MATCHES ENDSWITH BEGINSWITH CONTAINS
-%type <op> AND OR cop;
+%type <op> AND OR;
 
 %union {
    Leaf *leaf;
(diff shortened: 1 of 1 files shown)


liversedge:
e.g. TSS > 100 ? 1 : 0 
no longer parses.

amtriathlon:
Doing so re-introduces ambiguity in the grammar (i.e. several shift-reduce conflicts), but I thing it is not necessary: both side of the comparison are expr and the result is a (kind of) lexpr, I've tried that example and it works as expected since both symbol (TSS) and literal (100, 1, 0) are expr.
I will doe more testing anyway.
test

liversedge:
Currently (as in master)

TSS > 100 ? 0 : 1

Evaluates as TSS > (100 ? 0 : 1) !!!

amtriathlon:
Weird, I'm on master head + this commit and that formula works as expected in my build:
test1
Do you mean current head without this patch?

liversedge:
Are you sure .. use it as a data filter in the search box.
Do TSS > 100
then TSS <= 100
then TSS > 100 ? 0 : 1 ...

do the numbers align ?

liversedge:
MY BAD. I HAD NOT APPLIED YOUR PATCH.

Have now done so and it works.
Will pull and then test it.

SORRY.

PR #1532, opened 2015-08-23
fix ANT speed and cadence: value was still at last computed value

vlcvboyer:
When using training mode with speed or cadence sensor or speed/cadence sensor the speed remain at previous value when we stop the wheel.
This is due to speed telemetry update done only when we receive a signal from ANT device indicating a magnet detection.
My patch decreases the speed step by step when we receive ANT message from sensor without magnet detection. Formula is : rpm cannot be more than duration from previous magnet detection.
Tested with ANT Speed sensor and Cadence sensor on Windows

### DIFF
--- a/src/ANT.h
+++ b/src/ANT.h
@@ -414,6 +414,7 @@ public:
         lastCadenceMessage = QDateTime(QDateTime::currentDateTime());
         telemetry.setCadence(x);
     }
+    float getCadence(void) { return telemetry.getCadence(); }
     void setSecondaryCadence(float x) {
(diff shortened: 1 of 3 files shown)


vlcvboyer:
can you please check #1532 and merge?
It fixes a speed error (says that we are still riding when we stopped as it was updating the speed value only when the magnet passes in front of sensor but when wheel is stopped the magnet will no more go to sensor...)
Thanks

liversedge:
It has conflicts !

vlcvboyer:
Yes it was quite an old fix.
I've just synchronized it.
Compilation in progress... to ensure everything is ok. Then I'll push it again.

vlcvboyer:
Conflicts solved and compilation successful!

PR #1543, opened 2015-08-28
Improve on import from polar hrm files

leifwar:
I've made some changes to ease the import of polar hrm file into Goldencheeta. There are two main contributions:

1: Add polar created gpx file if it exist (always with the same name as the .hrm file). 
2: Handle R-R data if this is the only data that has been recorded. I sometimes turn of everything but the R-R recording in which case it will store this instead of the normal heart beats per minutes.

### DIFF
--- a/src/GpxParser.cpp
+++ b/src/GpxParser.cpp
@@ -147,12 +147,19 @@ bool
 
     else if (qName == "trkpt")
     {
+        // Time from beginning of activity
+        double secs = start_time.secsTo(time);
+
         if(lastLon == 0)
         {
             // update the "lasts" and find the next point
             last_time = time;
             lastLon = lon;
             lastLat = lat;
+	    // first point
+            rideFile->appendPoint(secs, cad, hr, 0, 0, 0, watts, alt, lon, lat, 0, 0.0, temp, 0.0, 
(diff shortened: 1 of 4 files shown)


liversedge:
Would it be better to store the R-R data in its own data series ?

liversedge:
We can add a new series quite easily.... then with the formula work I am doing it could be plotted without any code changes.

Do let me know what you think.

leifwar:
I agree with you that it would be much better to store the R-R data in its own data series. I did think about doing that but didn't know how or how that would fit with the rest of the code.

liversedge:
Its ok, I will commit this and then come back and add RR as a data series later.

THANKS FOR THE CONTRIBUTION :)

liversedge:
Can you email me your name so I can add you to the contributors list ?

Ta
Mark

leifwar:
My name is Leif Warland. 

Kopi: "leifwar" <leif.warland@getmail.no> 
Sendt: 28. august 2015 20:15:31 
Emne: Re: [GoldenCheetah] Improve on import from polar hrm files (#1543) 



Can you email me your name so I can add you to the contributors list ? 

Ta 
Mark

PR #1545, opened 2015-08-29
SrmRideFile: hacked up v9 support

rclasen:
it reads the data fine - but when I try to use the knowledge to write files, srmx refuses to read them or shows junk. So... something is likely still wrong.


liversedge:
Thanks Rainer. You're faster than me !

rclasen:
let's see what's missing. although it spits out the data as expected it's still feeling strange. I simply don't get what's missing to make srmx happy.

Issue #939, opened 2014-06-26
Train View Redesign


liversedge:
I want to redesign the Train View completely and was one of the drivers behind looking to build 'opentrainer' as a separate application. The refactoring of the views in 3.1 means this is not needed, so can be incorporated into the main application.

Key principles of the redesign;
1. work on running a workout that comprises a device setup, a workout script and optional video or other visuals
2. work on a main view that uses a graphscene and is dynamic, layered, responsive and provides 'tv quality' visuals
3. Remove the need to drag/configure contents and positioning -- it is tedious !
4. avoid being constrained by re-using the approach in the 'tile mode' for the other GC views

vlcvboyer:
I've made few tests which were successful:

![test speedometer overlay](https://cloud.githubusercontent.com/assets/13483855/9294869/3f1df7da-445a-11e5-8ab2-cb4e959d9549.jpg)

I have to clean the code then I'll put it on github.
It should be fully cross platform but I've tested only on Windows.

liversedge:
What have you added?

Is it those overlays ?

If so, then GREAT JOB :)

vlcvboyer:
Yes: overlays using Qt

vlcvboyer:
You can see/check/test/... my code here:
https://github.com/vlcvboyer/GoldenCheetah/tree/trainview

Do not hesitate to send comments.

vlcvboyer:
Hi,
I do not plan to raise a pull request as (from my point of view) it still have to be amended to allow customization (which meter to use and define size/position : XML file?).
Thereby beeing able, as a minimum, to disable it !
Otherwise it will always be part of Video playback but everyone do not want it...
What is your point of view?

liversedge:
If the selected media has data to control playback it should be honoured!
otherwise don't adjust video speed.

What customisation is needed ?

vlcvboyer:
Meters layout should be configurable from "none" to whatever the user want to see (speed, cadence, HR, power, distance, time, elevation profile, race map.....) it can be needle meters, digits, curves.... and placed wherever the user what to place them.
XML seems to be the better file format for this kind of job. The user will be able to select within the several preset layout (several XML files) the one he want to use.

This feature is not linked to video playback synchronization. But it works with it and do not interfere !

vlcvboyer:
I've created XML file for configuring my video playback overlays.
Feel free to test/check and send comments.
[...]

PR #1549, opened 2015-09-07
Train view: add meters as overlays

vlcvboyer:
linked to feature request Train View Redesign #939


liversedge:
I have consolidated into a single patch and pulled into the master repository.
I was not able to fully test, but will work through that shortly.
See commit dc402f12046e92e3a1f29c5cf194d083f0b72f50

PR #1552, opened 2015-09-08
Add new language support: Chinese(Traditional)

NicholasFengTW:
The initial version of Chinese(Traditional) support, I also update the line number in each .ts file because Pages.cpp needs to be modified to add new language option. 
Please help review and commit to trunk, thanks.


amtriathlon:
Hi Nicholas, I have no idea about chinese, but I could test change of languages works as expected and other languages are not affected, so I will merge it to master, thanks for your contribution!

PR #1565, opened 2015-09-24
Autoimport - new configuration option(s)

Joern-R:
... select only files for import which are CREATED or LAST_CHANGED in the file system during last 90/180/360 days
... thus limiting the number of files which are considered for import in the source directory
   (without changing/deleting the source files - and without the need to track indexes,...
    of what has already been imported).

### DIFF
--- a/src/Pages.cpp
+++ b/src/Pages.cpp
@@ -5593,17 +5593,18 @@ AutoImportPage::saveClicked()
 
     // re-read
     context->athlete->autoImportConfig->readConfig();
-
     return 0;
 }
 
 void
 AutoImportPage::addRuleTypes(QComboBox *p) {
 
+    RideAutoImportRule* config = new RideAutoImportRule();
+    QList<QString> descriptions = config->getRuleDescriptions();
 
-    p->addItem(tr("No autoimport"));
-    p->addItem(tr("Autoimport with dialog"));
-
+    foreach(QString description, descriptions) {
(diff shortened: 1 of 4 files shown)


liversedge:
wouldn't it be better just to keep a record of the imported files and ignore files that have been previously imported ?

Joern-R:
Agreed - was thinking on a record of imported files as well. But this may get more complicated when e.g. trying to substitute a file with a new version of the same training. What would be the "object" to track to find out if an activity is already imported. And has to be added in various place - e.g. also when deleting,... - what happens if a user delete via OS - is the record self-reparing,...

Obviously easiest could be "date/time" of the activities - which would provide a good match for most cases. So this solution may only be an iterim one - until such index is available.

liversedge:
Given it would only affect automatic importing there is no harm in not autoimporting a file once it has been imported before. the user can always select it manually as they do today ?

I think the list would be global though, since once it has been imported for one rider it should be ignored for all others ?

amtriathlon:
What about to put an additional field for the original file name in the .json to avoid having another file to maintain? The ride cache could generate the list or test the existence.

liversedge:
Yes. We could add a metadata field "Import Filename" and get it from the metadata tags stored in the RideItem. When constructing the ride cache we could extract these into a stringlist too.

liversedge:
mind you, that wouldn't be global...........

PR #1571, opened 2015-09-26
Segregate General & Athlete Settings on preferences main page

Joern-R:
On the Preferences Main page - with Config Migration - is a mix of General and Athlete specific settings. To visualize the grouping for the user, the settings have been re-grouped and separated like in the screen shot below. Still the question is if it would be better to move all Athlete settings to the "Athlete" page. (Which would then get rather big - and some fields which have been on the main pref. page for a long time would appear to have disappeared).
athlete

### DIFF
--- a/src/Pages.cpp
+++ b/src/Pages.cpp
@@ -90,6 +90,121 @@ GeneralPage::GeneralPage(Context *context) : context(context)
     configLayout->addWidget(langCombo, 0,1, Qt::AlignLeft);
 
     //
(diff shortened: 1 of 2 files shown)


liversedge:
I think they should move to the Athlete Page ?

Joern-R:
Agreed and done - Moved the whole athlete specific field section to Athlete preferences
athlete2

liversedge:
PERFECT ! 

Thank you :)

PR #1576, opened 2015-10-01
FixElevation bug fixes

nickb937:
- support user-supplied MapQuest API key to workaround the per-day API
  request limit on the pre-existing key.
- terminate elevation processing when Bad Request and Developer Limit
  errors are received to prevent the possible destruction of elevation
  data in other files that are waiting to be processed
- stop a memory leak and file descriptor leak by deleting the
  QNetworkAccessManager() object.  The descriptor leak can cause a
  crash if many files are needing processing.


liversedge:
Hi Nick, 

I missed this and just stumbled across it. It has a merge conflict due to my fix for a SEGV.
Is there /any/ chance you could fix up and then I'll merge it in ??????

nickb937:
Hmm, this pull request has turned into a mess.  I'll close and re-submit a new one.

PR #1577, opened 2015-10-01
Move "Unit" config from "Athlete" to "General" level

Joern-R:
... Unit Field is moved from Athlete to General preferences
... first Athlete opened with new logic determines the general settings
... if configuration  is already migrated (based on one of the previous commits) -
    migration can be forced to happen again by removing the INI files or
    just the preferences pages has to be opened and Saved with the
    unit settings you want

### DIFF
--- a/src/Athlete.cpp
+++ b/src/Athlete.cpp
@@ -79,11 +79,11 @@ Athlete::Athlete(Context *context, const QDir &homeDir)
     if (returnCode != 0) return;
 
     // metric / non-metric
-    QVariant unit = appsettings->cvalue(cyclist, GC_UNIT);
+    QVariant unit = appsettings->value(NULL, GC_UNIT, GC_UNIT_METRIC);
     if (unit == 0) {
         // Default to system locale
         unit = QLocale::system().measurementSystem() == QLocale::MetricSystem ? GC_UNIT_METRIC : GC_UNIT_IMPERIAL;
-        appsettings->setCValue(cyclist, GC_UNIT, unit);
+        appsettings->setValue(GC_UNIT, unit);
     }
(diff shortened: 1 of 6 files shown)


amtriathlon:
Good idea, units seems more like a preference for the user than for the athlete, I think we should also move Pace / Swim Pace metric option to General tab since they are already global.

Joern-R:
Ale - would you be so kind and do the move of Pace/Swim Pace option also to
General Page - you better know potential side effects.

Since the Settings themselves are already "Global" there is not need to
adjust anything in Settings.h/Settings.CPP or in the Migration Tool for
this.
Thanks
Joern

2015-10-02 15:42 GMT+02:00 Mark Liversedge <notifications@github.com>:

amtriathlon:
No problem Joern, I will do.

2015-10-02 10:46 GMT-03:00 Joern <notifications@github.com>:

> Ale - would you be so kind and do the move of Pace/Swim Pace option also to
> General Page - you better know potential side effects.
>
> Since the Settings themselves are already "Global" there is not need to
> adjust anything in Settings.h/Settings.CPP or in the Migration Tool for
> this.
> Thanks
> Joern
>
> 2015-10-02 15:42 GMT+02:00 Mark Liversedge <notifications@github.com>:
>
> > Merged #1577 <https://github.com/GoldenCheetah/GoldenCheetah/pull/1577>.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/GoldenCheetah/GoldenCheetah/pull/1577#event-425260855>
> > .
> >
>
> —
> <https://github.com/GoldenCheetah/GoldenCheetah/pull/1577#issuecomment-145025654>
> .
>



-- 
Alejandro.

PR #1578, opened 2015-10-02
DropboxSupport - mintor text corrections

Joern-R:
... Use different Shorcut in Menu (Crtl+B already in use)
... Exchange text reference to TP by generic approach


liversedge:
Thanks ! attention to detail never my strong point !

PR #1580, opened 2015-10-04
Add "Shared Network Folder" as new "File Store" implementation

Joern-R:
... use a shared network drive (e.g. mounted via WebDAV) to synchronize activity data
   (similar to Dropbox,...)

Open/TBD:
... where to best put the Preferences for the different FileStores
   (currently add the Network folder configuration on Preferences->Passwords,
    even though no PW is needed - with growing number of FileStores a
    dedicated Preferences Page may be the solution)
... where to put the MainWindow Menu entries if the number of FileStores increases

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -80,6 +80,7 @@
 #if QT_VERSION > 0x050000
 #include "Dropbox.h"
 #endif
+#include "NetworkFileStore.h"
 #include "FileStore.h"
 
 // GUI Widgets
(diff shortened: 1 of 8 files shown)


liversedge:
Is there a site I can use to test this ?

Joern-R:
It's not related to a site - the files are stored e.g. on a network drive.
I tested it against my NAS storage which is mounted a a network drive.

I will look for a Cloud Site which provides WebDAV access we
can do a cross sync. test. Will let you know via mail when ready.

2015-10-04 20:06 GMT+02:00 Mark Liversedge <notifications@github.com>:

liversedge:
The webdav config should ask for a hostname to connect to ?
Then the browse button should make calls on the webdav server via the FileStore readdir method.

PR #1582, opened 2015-10-05
Local File Store

Joern-R:
... wrong status text (".") missing


liversedge:
yeah, this is a bit crappy. there should be a more reliable way of indicating sucess!

PR #1584, opened 2015-10-08
Fixed Pace Zones Config

amtriathlon:
Modifications were lost without notification if the user changed sport before saving.

Now we could move the pace metric option to general, but I'm not sure because:
- on the positive side units settings would stay together and it should be clear they are not athlete dependent.
- on the negative side the code would be uglier and the setting effect not as immediately visible for the user as now
Comments are welcome.


liversedge:
yeah i think keep the pace units on the pace config screen.

PR #1585, opened 2015-10-08
Fix Speed from Distance

### LINKED ISSUE #1568: Pace of running intervals
Hi,

I'm using GC 3.2 on Ubuntu 15.04 and have noticed a bug in the detection of the peak pace for running intervals. The detected pace for the intervals seems to be different from the actual pace, as you can see in the screenshots in attachment. The 20 min peak pace is detected as 6:17 min/km, while the actual pace is 5:48 min/km.
 
Interval detection - peak pace for running activities

I've seen differences in pace up to 0:30 min/km, where the detected pace can be either faster or slower than the actual pace.
It seems that the incorrect value of 6.3 (= 6:17 min/km) is used for the charts in the long term trends as you can see below.
 
Use of incorrect peak pace in long term trends

The same behaviour occurs when running GC 3.2 under Windows 7. I'm recording the activities with a Garmin Edge 800 and 1s recording. I tried selecting "Smart recording" in the GoldenCheetah settings, but this does not make any difference.

Any idea if this is a (known) bug or if it has something to do with my hardware and/or settings?

Kind regards,

Nigel

amtriathlon:
Optionally smoothed by configurable moving average parameter
Fixes #1568

### DIFF
--- a/src/FixSpeed.cpp
+++ b/src/FixSpeed.cpp
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) 2015 Alejandro Martinez (amtriathlon@gmail.com)
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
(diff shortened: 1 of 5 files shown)


liversedge:
Nice !

PR #1591, opened 2015-10-13
Ant-fec protocol : additional features (slope simulation mode...)

vlcvboyer:


liversedge:
Superb, thank you so much for doing this.
Which trainer did you test with ?

vlcvboyer:
tested on my smart Vortex

PR #1596, opened 2015-10-17
AutoBackup

Joern-R:
... Abort does also reset the counter (so that user is not asked every further close - but only after the next period)

### DIFF
--- a/src/AthleteBackup.cpp
+++ b/src/AthleteBackup.cpp
@@ -129,7 +129,7 @@ AthleteBackup::backupOnClose()
     zipFile.close();
     ZipWriter writer(zipFile.fileName());
 
-    QProgressDialog progress(tr("Adding files to backup %1 for athlete %2 ...").arg(targetFileName).arg(cyclist), tr("Abort Backup"), 0, fileCount, NULL);
+    QProgressDialog progress(tr("Adding files to backup %1 for athlete %2 ...").arg(targetFileName).arg(cyclist), tr("Abort Backup and Reset Counter"), 0, fileCount, NULL);
     progress.setWindowModality(Qt::WindowModal);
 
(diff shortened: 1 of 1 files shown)


liversedge:
I wonder if we should add a menu option to take a backup and reset the counter then too ?

Joern-R:
I was thinking about that - but was not sure if we can guarantee a
consistent snapshot then.

Advantage of doing it when closing is that all files are in-sync and
already saved and
closed - so non of the files taken should in in an undefined state (e.g.
after manual changes, but
before saving them to disc).

Only thing which is with the "at close" logic not in sync yet in the
snapshot is the counter
itself - its not yet reset - since I zip the .INI file first and then reset
the counter as last step.

Do you know if there is a way to check for "everything" stored - so that a
snapshot of the
data can be taken ? If we add that to the menu - where should it reside ?

2015-10-17 19:06 GMT+02:00 Mark Liversedge <notifications@github.com>:

liversedge:
To find rides that need saving:

foreach (RideItem *rideItem, context->athlete->rideCache->rides())
        if (rideItem->isDirty() == true)
             dirtyList.append(rideItem);

There is a a dialog in SaveDialogs.cpp.

Personally, I would just warn unsaved changes will not be backed up.
Best menu to add it to would be the Athlete menu ?

Joern-R:
Good - will add this for 3.3 soon.
- in the Athlete Menu (Same structure like "open" - so that any Athlete can
be backup up, independent of open or not)
- sync all INI files via QSettings to disk (if Athlete is open)
- checking for "dirty" (if Athlete is open) and giving a warning


2015-10-18 1:09 GMT+02:00 Mark Liversedge <notifications@github.com>:

PR #1597, opened 2015-10-17
TrainMode - Use Multimedia Keys

Joern-R:
... add Multimedia Keys to control the train mode
... Start/Stop/Forward/Rewind/Pause working on MCE (Windows Media Center compatible remote controls) without any problems
    ... tested with 2 MCE remotes under Windows (Mac and Unix still need to be tested/verified)
... For "new Lap" none of the special media keys worked - so choose "0" since this is available even on pure remotes

Since in earlier versions of QT not all Keys are defined, just start with 5.4 as pre-requisite.

### DIFF
--- a/src/TrainSidebar.cpp
+++ b/src/TrainSidebar.cpp
@@ -233,6 +233,9 @@ TrainSidebar::TrainSidebar(Context *context) : GcWindow(context), context(contex
     rewind->setStyleSheet("background-color: rgba( 255, 255, 255, 0% ); border: 0px;");
     rewind->setAutoRepeat(true);
     rewind->setAutoRepeatDelay(200);
+#if QT_VERSION > 0x050400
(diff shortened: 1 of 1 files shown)


liversedge:
NICE!

So you can control train view with a media center remote on windows ?

Joern-R:
Yes - only nit still is that after the first "Start" the main window gets
control again. You have to put focus back to the "toolbox window".
(I suspect it relates to starting the video - but not sure - can be any
other widget as well).

But this happens only once. So you can "Start" - "Pause", put the focus to
the "tools windows" and then pause, stop, ff, rw, new lap.
From the remote - I actually had one which I never used (used the diNovo
MCE Keyboard instead). Just stumbled over it and gave it a try today.

Would be curious if the Media Keys on a Mac Bluetooth keyboard can do the
some job.

2015-10-17 19:08 GMT+02:00 Mark Liversedge <notifications@github.com>:

PR #1599, opened 2015-10-18
video-sync: add "None" menu entry and disable gpx from activities

vlcvboyer:
"None" entry in RLV list allow user to disable this feature


Joern-R:
I have just tested this - works fine - just one thing I noticed is when you switch from RLV to NONE and do not change the media, the playback speed is taken from RLV (seems that VLC does not reset to media FPS when starting again. This attached patch fixes it.

[patch.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/16201/patch.txt)

PR #1616, opened 2015-10-30
Added vdottime(VDOT, distance) to formulas

amtriathlon:
Computes the equivalent time (in seconds) for distance (in km)
at VDOT, tipically used with VDOT metric as first parameter.


liversedge:
nice !

PR #1617, opened 2015-10-31
Take wind into account during power estimation

vlcvboyer:
... collect weather broadcast included in garmin FIT files (when used with smartphone)
   ... take into accound wind speed and heading during power estimation
             (for those like me who don't have power sensors !)
   ... manual entry of wind speed and heading proposed in dialog box
             (for those who don't have garmin...)

in any case this remain an inaccurate estimate !

### DIFF
--- a/src/FixDerivePower.cpp
+++ b/src/FixDerivePower.cpp
@@ -24,6 +24,10 @@
 #include <algorithm>
 #include <QVector>
 
+#ifndef MATHCONST_PI
+#define MATHCONST_PI 		    3.141592653589793238462643383279502884L /* pi */
+#endif
+
 // Config widget used by the Preferences/Options config panes
 class FixDerivePower;
 class FixDerivePowerConfig : public DataProcessorConfig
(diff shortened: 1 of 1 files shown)


liversedge:
Is windspeed and windheading a single value that applies for the entire ride ?

vlcvboyer:
In regard with wind speed and wind heading:
NO: there are multiple values when extrated from garmin FIT file: as we have several weather broadcast data during whole ride (closest weather station or fresh data incoming). I use them as soon as available (we can do a weighted average, etc. if we want to be more accurate but this will never be accurate! It will be more useful to indicate if the ride was done alone or behind someonelse)
YES: only one value at present when entered manually
In the future we will be able to use a weather history API like weatherunderground or openweatherdata. I've tried it but the free version has theroreticaly limitations... I'll investigate further.

But of course headwind (not windheading!) is computed based on cyclist direction for each track record.
Thereby when you had wind in front of your face at the beginning of ride the power is higher but when back at home it is less..

I saw that headwind was already part of activity records so I just populate it with garmin FIT file weather broadcast data and then used it for power estimation (wind was assumed 0).

vlcvboyer:
Hi,
Were you able to review this patch?

grauser:
Warning in GC headwind series is wind speed. If you ride at 30km/h with no wind headwind should be 30km/h too...

vlcvboyer:
Noted!
Patch updated accordingly!

vlcvboyer:
I've just resync it with current master branch as a recent commit created conflicts

vlcvboyer:
Hi,
Is it ok? Do you have any additional comment?

Best regards

vlcvboyer:
resync again in order to resolve conflicts with current master branch!

liversedge:
will merge when checks complete.

PR #1619, opened 2015-11-03
Basic Stryd (Power Meter for Running) support

amtriathlon:
Enables power metrics for non cycling activities when power data is present.

### DIFF
--- a/src/BasicRideMetrics.cpp
+++ b/src/BasicRideMetrics.cpp
@@ -719,7 +719,7 @@ class TotalWork : public RideMetric {
         }
         setValue(joules/1000);
     }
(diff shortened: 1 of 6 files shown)


jonbev:
Nice one. I have had mine delivered but haven't had a chance to get it going. I connected it to my watch (Fenix 3) with no problem but couldn't seem to get it to register any power reading. HR was fine.

jonbev:
I have mine working now and have done a few runs. It would be great to have running specific power zones. Is this possible?

amtriathlon:
Possible yes, we already have sport specific Pace Zones working as a model to follow, but I'm afraid it affects too much of the code to make the cut for v3.3, I would wait up to 4.0.
In the meantime you can use a separate athlete for this purpose.

PR #1620, opened 2015-11-03
Fix SEGV when using "Library/GoldenCheetah" athlete directory (in 3.3… only

Joern-R:
... problem relates to introduction of new .INI files
... old /gc ini file approach is still in place - now again working


liversedge:
is this right? It won't do a migration unless we have already migrated !?

PR #1624, opened 2015-11-05
ANT FE-C

dresco:
Detect FE-C trainer at channel connection (based on the kickr code), 
and make an initial request for device capabilities at this time.

Check erg mode is supported before requesting a target load.

Works well for me, but will ask both Mark and Vianney to review prior to merging..


liversedge:
Does it do the right thing with FE-C vs PM ?
Seems trivial enough to merge for me.
Will see if Vianney has any thoughts too.

vlcvboyer:
Ok for me.
FYI I've tested it on Win64 / debug / Qt5 / libvlc / ANTUSB-m stick / Smart Vortex (setup as FE-C) / slope simulation mode.

PR #1625, opened 2015-11-06
french translation partial update

vlcvboyer:
I changed directly in gc_fr.ts file after updating it with lupdate
Is it the expected way to do it or is it mandatory to use qt software to edit this file?


vlcvboyer:
Hi
Can you please review and merge?
...there is still a lot of text to translate but I want to ne sure that I'm doing it in a suitable way!

grauser:
Could you contact me directly ?
Thank you.

vlcvboyer:
... yes... how to do that? can't see your email !

vlcvboyer:
Does my new commit match your expectations?

grauser:
Yes, thank you.

Issue #980, opened 2014-07-20
Migrate to Unified LibUsb 1.1.18


liversedge:
http://libusbx.1081486.n5.nabble.com/Libusbx-devel-Announcing-libusb-1-0-18-as-well-as-libusbx-1-0-18-FINAL-td2375.html

See here: https://github.com/libusb/libusb
And: http://libusb.info/

lurdan:
It seems GC not to be going to support newer libusb versions (otherwise assumes to use bundled tarball), right?

Ramoonus:
v 1.0.22 is out now
https://github.com/libusb/libusb/releases

PR #1641, opened 2015-11-08
Libusb 1.1.X support (bug #980)

nickb937:
Bug #980

Adds support for Libusb 1.1.X as a compile-time option (set GC_HAVE_LIBUSB1).  The existing behaviour of compiling against libusb 0.1 is retained.   Libusb 1.1 makes Train mode stable on Linux and allows for receiving ANT+ data for longer than 3-5 minutes.

The code is known not to work correctly on Windows and I lack a suitable development environment to debug/test it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 branches:
 - master
-osx_image: xcode7.2 # will also get OSX El Capitan
+osx_image: xcode7.2b1 # will also get OSX El Capitan
 language:
 - objective-c
 compiler:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 branches:
 - master
-osx_image: xcode7.2b1 # will also get OSX El Capitan
 language:
 - objective-c
 compiler:
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,15 +96,11 @@ script:
 - CC=clang CXX=clang++ make -j4 sub-src
 after_success:
 - cd src
-#fix for libcocoa.dylib issue with qt5.4.2
-#not much of a fix, but will checkout macdeployqtfix shortly...
-##- mkdir -p GoldenCheetah.app/Contents/MacOS/platforms
-##- cp /usr/local/opt/$QT_PATH/plugins/platforms/libqcocoa.dylib GoldenCheetah.app/Contents/MacOS/platforms/
-## fix up the bundle with macdeployqtfix
-- python ../travis/macdeployqtfix.py GoldenCheetah.app /usr/local/opt/$QT_PATH
 - ls -laR GoldenCheetah.app
 - sh ../travis/check-gc.sh
 - /usr/local/opt/$QT_PATH/bin/macdeployqt GoldenCheetah.app -verbose=2 -dmg
+## fix up the bundle with macdeployqtfix
+- python ../travis/macdeployqtfix.py GoldenCheetah.app /usr/local/opt/$QT_PATH
 - brew remove $QT
 - mv GoldenCheetah.dmg GoldenCheetah_$QT.dmg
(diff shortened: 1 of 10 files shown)


liversedge:
This one is gonna need some thought !
Will need to put into the v4 queue, so likely to need rebasing and merging in the new year.
Might be worth consolidating into a single patch or local branch at your end .

liversedge:
This is to fix problems with a Fortius, not ANT, right?

nickb937:
It fixes problems with all ANT reception on Linux (I've tried multiple machines and kernels).  Train mode for any of my ANT devices (speed/cadence, vector, hrm) fail to receive data after a couple of minutes due to hangs in libusb.

liversedge:
Strange I have no issues, with Ubuntu 14.04 and libusb 0.1.12 from the contrib directory.

liversedge:
The travis build failure is coz -lusb1 is added to src.pro unconditionally I suspect ?

ibooij:
You could also try to use [libusb-compat](http://www.libusb.org/wiki/libusb-compat-0.1). You can then use libusb-1.x on Linux and libusb-win32 on Windows almost without changing any code.

I've used this in my own cycling program, [Big Ring](https://github.com/ibooij/big-ring)

arcanericky:
Be aware. This pull request contains a modification to `src/src.pro` that removes:
```
HEADERS +=  APIWebService.h
SOURCES +=  APIWebService.cpp
```

This breaks builds when uncommenting (enabling):
```
#HTPATH = ../httpserver
```

The error emitted is:
```
main.o: In function `main':
main.cpp:(.text.startup+0x1de2): undefined reference to `vtable for APIWebService'
RideDB_yacc.o: In function `RideDBparse(RideDBContext*)':
RideDB_yacc.cpp:(.text+0x1111): undefined reference to `APIWebService::writeRideLine(RideItem&, HttpRequest*, HttpResponse*)'
```

Otherwise, these changes fix problems and are working great for me for 1 hour + rides on Ubuntu 15.10.

I also looked into the errors on the Travis build, and Mark is correct. It's failing because it's still trying to use the old `libusb`.

WaterworthD:
I've checked this with the Fortius VR and libusb-1.0.20. It works but but some work needs to be done in the read method to ensure it's the same as before and the Fortius discovery method appears to have been changed (doesn't differentiate between the VR and pre-VR models)

Slightly off topic but do we still need to read from the usb device into a 64 byte buffer then return 1 byte per ANT read from this buffer?

flyingflo:
I can confirm the ANT reception problems as described by @nickb937 on ubuntu 15.04. After some minutes no more packets can be read from libusb, it always return error 110 (iirc).

I worked around it experimentally by linking libusb 1.0 and libusb-compat instead of libusb0.1. It's quite easy, and works well.

lurdan:
What blocks this issue? (at least conflicted PR?)
( If there are something I can do, I will try..)

antoniusriha:
I'd like to pick this up but need some directions as to how to implement it. Currently win32 code loads libusb0.dll dynamically on-demand, whereas other OS code links to libusb 0.1. For Linux it would be best to port to libusb 1.0. For Windows the easiest would be to leave it untouched (since everything is working fine). For MaxOSX - I've no idea. So my question is: Should we port to libusb 1.0 for Linux (and MacOSX?) only and let win32 use libusb 0.1? Or should we port everything to libusb 1.0?

antoniusriha:
Since porting to libusb-1.0 on Windows seems to be a bit of a PITA [1], I'd suggest to only port Linux/MacOSX code paths to libusb-1.0. @liversedge: Will such a patch be accepted?

[1] https://sourceforge.net/p/libusb/mailman/libusb-devel/thread/54D91E86.1040002@probo.com

PR #1642, opened 2015-11-08
fix few compiler warning and related minor bugs

vlcvboyer:
few fix based on compiler warning
as per my understanding web pages will also have to be updated (new or missing features => new entries in HelpWhatsThis.cpp)


liversedge:
Thanks .. I wonder if that overlay code error is actually a real bug you just fixed....

liversedge:
YUP! AWESOME you just fixed a really annoying bug related to the mouse cursor and state when working with the overlay box in the CP window.

THANK YOU! That one has driven me nuts for over a year

PR #1643, opened 2015-11-09
SportPlusHelp Cloud Integration

LucaRa177:
Hello, 
I developed a simple integration (based on the already existing ones) so that GC is able to send specific trainings to SportPlusHealth servers to be included into user’s training list.
I hope everything’s fine and you will accept it. 
Best regards, 
Luca

### DIFF
--- a/src/Pages.cpp
+++ b/src/Pages.cpp
@@ -720,6 +720,30 @@ CredentialsPage::CredentialsPage(QWidget *parent, Context *context) : QScrollAre
 
 
     //////////////////////////////////////////////////
+    // SportPlusHealth
+
+    QLabel *sph = new QLabel(tr("SportPlusHealth"));
+    sph->setFont(current);
+
+    QLabel *sphUserLabel = new QLabel(tr("Username"));
+    QLabel *sphPassLabel = new QLabel(tr("Password"));
+
+    sphUser = new QLineEdit(this);
+    sphUser->setText(appsettings->cvalue(context->athlete->cyclist, GC_SPORTPLUSHEALTHUSER, "").toString());
+
+    sphPass = new QLineEdit(this);
+    sphPass->setEchoMode(QLineEdit::Password);
+    sphPass->setText(appsettings->cvalue(context->athlete->cyclist, GC_SPORTPLUSHEALTHPASS, "").toString());
+
+    grid->addWidget(sph, ++row, 0);
+
+    grid->addWidget(sphUserLabel, ++row, 0);
+    grid->addWidget(sphUser, row, 1, Qt::AlignLeft | Qt::AlignVCenter);
+
+    grid->addWidget(sphPassLabel, ++row, 0);
+    grid->addWidget(sphPass, row, 1, Qt::AlignLeft | Qt::AlignVCenter);
+
+    //////////////////////////////////////////////////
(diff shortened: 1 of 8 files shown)


liversedge:
The integration test failed with "SportPlusHealthUploader.cpp:26:10: fatal error: 'QMessageLogger' file not found"

liversedge:
Currently we build on QT4.8.6 and QT5.5, I think QMessageLogger is QT 5.3 or higher.

LucaRa177:
Sorry, you are right I developed using QT5. Anyway I removed that library, as I needed it just for logging during the development.

liversedge:
Perfect. Will wait for the integration test then check through and merge.
I assume there are no API tokens/secrets needed?

liversedge:
Oh, QJsonDocument is not in QT4.8 either. You are welcome to make the whole thing conditional on QT5. We did that with the recent Dropbox support; we will likely drop support for QT4 in January...

LucaRa177:
I see that in ShareDialog it's been used the MVJSONReader. Is it compatible with QT4? I could fastly switch to this library and commit the changes.

liversedge:
Sure. Its a very simple library that doesn't handle arrays well, but good for small snippets.

LucaRa177:
Everything should be OK. How does it work now? Is there some step I still miss?
Sorry for the questions, but I never took part to a collaborative opensource project until now.
Could you also tell me when these changes will appear in the downloadable release of the app?

liversedge:
Yes it compiles fine. 

I need to check the code through and then merge probably today or tomorrow.
We have not started the release cycle for v3.3 yet, although a first public development build for it was published yesterday. I am sure we will publish a release candidate shortly and this will be in there.

The full release of v3.3 is hoped to be Jan 1st 2016, assuming tests go well.

LucaRa177:
OK thank you for the info ;)

liversedge:
Sorry Luca but if I open the Share Dialog without configuring name/pass the checkbox is enabled and checked !  It should be disabled until the user settings have been made. I looked at the code and it seemed ok, do you know why this happens?

LucaRa177:
Hello, I just checked deleting the credentials already inserted and that checkbox was ok (unchecked and not enabled). But I'm not in the use case "I enter the first time without ever inserting the credentials", as I think you are. Is there a way to "reset" the application (I tried with a "Rebuild all" but it was not enough)?

liversedge:
Lets get it into the repo, we can fix up nits later.
The release candidates are out soon so best not to miss that bus !

LucaRa177:
Great thank you. I saw you edited the method " SportPlusHealthUploader::canUpload", was that missing empty parameter causing our problem?

LucaRa177:
Sorry, I've just read the commit comment, everything's clear now. Thanks for the fix.

LucaRa177:
Hello,
not sure if this is the right place to post about, but I noticed a minor error. On Golden Cheetah website, page "http://www.goldencheetah.org/#section-features", it is written "Sports Plus Health" while we should see "SportPlusHealth" (no "s", and no spaces).
How could we fix this?
Thank you,
Luca

liversedge:
Have fixed it.

LucaRa177:
Thank you

PR #1649, opened 2015-11-11
Disable Average Speed Override on ManualRideDialog

### LINKED ISSUE #1645: Swim pace calculation error
Hi,

I'm manually entering my swim data, so just distance and time for the whole swim. The last few months, after 15 August, something seems wrong with the calculation of the average swim pace (min/100m). I believe this is around the date where I started using GC 3.2 on Ubuntu.  

These are some data before 15 August:
image

GC seems to calculate pace correctly

And here are some data after 15 August:
image

GC pace is off by 1-3 s/100m

Could this be a bug in GC 3.2?

Kind regards,

Nigel

amtriathlon:
It creates problems with pace metric due to limited precision,
better changed via Distance/Duration to preserve consistency.
Fixes #1645

If the user REALLY want to override Average Speed to be different from Distance/Duration it always can do that in the details tab, but I don't think it is a common use case.

### DIFF
--- a/src/ManualRideDialog.cpp
+++ b/src/ManualRideDialog.cpp
@@ -218,6 +218,8 @@ ManualRideDialog::ManualRideDialog(Context *context) : context(context)
     avgKPH->setDecimals(1);
     avgKPH->setMinimum(0);
     avgKPH->setMaximum(100);
+    // Average Speed depends on Distance and Duration, only for display
+    avgKPH->setEnabled(false);
 
     // how to estimate stress
     QLabel *modeLabel = new QLabel(tr("Estimate Stress by:"), this);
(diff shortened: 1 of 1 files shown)


amtriathlon:
I think this is not the better idea, perhaps we should go back to the behavior before this commit https://github.com/GoldenCheetah/GoldenCheetah/commit/efd3859d2e876af97be6e848435bda32a5be3dc3 WRT Average Speed.

liversedge:
I'll take your advice on this Ale.

amtriathlon:
Mark, there is part of that commit that should preserved: the one that remembers when manual estimation method was last used. I will send a new PR.

PR #1651, opened 2015-11-11
Add more "sport" tags + read weight from FIT file.

leifwar:
I've added some more sport tags when reading garmin FIT files based on what I found in http://pub.ks-and-ks.ne.jp/cycling/GarminFIT.shtml.  I also found and added reading of weight data stored in the FIT file. 

I've only tested this on FIT files generated with a garmin fenix 3 watch.

### DIFF
--- a/src/Charts/AllPlot.cpp
+++ b/src/Charts/AllPlot.cpp
@@ -350,6 +350,10 @@ AllPlotObject::AllPlotObject(AllPlot *plot, QList<UserData*> user) : plot(plot)
     hrCurve->setPaintAttribute(QwtPlotCurve::FilterPoints, true);
     hrCurve->setYAxis(QwtAxisId(QwtAxis::yLeft, 1));
 
+    hrvCurve = new QwtPlotCurve(tr("R-R Rate"));
+    hrvCurve->setPaintAttribute(QwtPlotCurve::FilterPoints, true);
+    hrvCurve->setYAxis(QwtAxisId(QwtAxis::yLeft, 1));
+
     tcoreCurve = new QwtPlotCurve(tr("Core Temp"));
     tcoreCurve->setPaintAttribute(QwtPlotCurve::FilterPoints, true);
     tcoreCurve->setYAxis(QwtAxisId(QwtAxis::yLeft, 1));
@@ -582,7 +586,7 @@ AllPlotObject::setColor(QColor color)
     QList<QwtPlotCurve*> worklist;
     worklist << mCurve << wCurve << wattsCurve << atissCurve << antissCurve << npCurve << xpCurve << speedCurve << accelCurve
              << wattsDCurve << cadDCurve << nmDCurve << hrDCurve
-             << apCurve << cadCurve << tempCurve << hrCurve << tcoreCurve << torqueCurve << balanceLCurve
+             << apCurve << cadCurve << tempCurve << hrCurve << hrvCurve <<tcoreCurve << torqueCurve << balanceLCurve
              << balanceRCurve << lteCurve << rteCurve << lpsCurve << rpsCurve
              << lpcoCurve << rpcoCurve
              << altCurve << slopeCurve << altSlopeCurve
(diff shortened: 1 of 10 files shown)


liversedge:
thanks, sub-sport might be a problem as we want Run or Bike or Swim, so will probably remove that and put into a Sub-Sport tag instead.

liversedge:
gonna put in the queue for v4 simply because it could break a few things that rely upon Sport being a particular value and also some multisport FIT files will break current behaviour with this code.

leifwar:
Sounds like a good idea to make a sub-sport tag. I would like to see other sports included though, such as cross country skiing (very popular around here).

liversedge:
Any joy with sub sport ?

leifwar:
I have been a little busy but will dig into to it.

liversedge:
Actually Vianney sent through an update, which is similar to yours but didn't set the sub-sport field.
There are likely a gazillion conflicts ! Sorry about that.

leifwar:
That's ok. I will try to expand on what has been done so far.

vlcvboyer:
Yes, my patch creates minor conflicts with yours. Sorry about that.
I code few fields that I use at present but most are just comments placed there in order to share this information with other contributors and avoid loss of time when one of those fields will be used in the future. FIT files documentation is not really easy to find...

leifwar:
I have made the SupSport tag and tried to include it in the Details->Workout tab. It took some time to figure out how these fields where added. Correct me if I'm wrong but It seems to me the src/xml/metadata.xml is used for this purpose, though I can't figure out how to force upgrade of users xml/metadata.xml file. I still have other sports than Run, Bike and Swim.

liversedge:
Apologies for the delay in merging. I missed it.

leifwar:
No problem. I've been tracking your development and doing a lot of rebasing to keep it up to date. Hope you don't mind that. 

I don't know if the quality of the R-R data is actually good enough to be used for anything. For a while the data was clearly wrong but that was fixed by an upgrade of the watch's software. 

Best regards, 
Leif 

Kopi: "Leif Warland" <leif.warland@getmail.no>, "Author" <author@noreply.github.com> 
Sendt: 29. januar 2017 10:28:22 
Emne: Re: [GoldenCheetah/GoldenCheetah] Add more "sport" tags + read weight from FIT file. (#1651) 



Apologies for the delay in merging. I missed it.

liversedge:
Ack - I didn't notice the change to the ridefile format.
Will have to remove that, its a massive regression.

liversedge:
I've reverted the commits that change the ridefile format.

If you can make it fit into XDATA series then I can commit, but any changes to the JSON ridefile format need to be carefully considered as they need to be backwards compatible, which is tricky since we use a custom parser.

leifwar:
The R-R data does not match the time vector (it is the time it self) as all the other variables so I'm not sure if it can fit into XDATA. Is it possible to have different XDATA of different lengths? If so it can be done. 

Kopi: "Leif Warland" <leif.warland@getmail.no>, "Author" <author@noreply.github.com> 
Sendt: 29. januar 2017 14:07:30 
Emne: Re: [GoldenCheetah/GoldenCheetah] Add more "sport" tags + read weight from FIT file. (#1651) 



I've reverted the commits that change the ridefile format. 

If you can make it fit into XDATA series then I can commit, but any changes to the JSON ridefile format need to be carefully considered as they need to be backwards compatible, which is tricky since we use a custom parser.

liversedge:
Yes, XDATA has a separate distance and time value. Also, you don't need MSECS since SECS is a double.

leifwar:
So if we take out b47cb73 and d432602 and then re-implement the ladder to read into XDATA we should be fine? I guess there is no need to implement plotting functionality for XDATA. 

Kopi: "Leif Warland" <leif.warland@getmail.no>, "Author" <author@noreply.github.com> 
Sendt: 29. januar 2017 14:37:15 
Emne: Re: [GoldenCheetah/GoldenCheetah] Add more "sport" tags + read weight from FIT file. (#1651) 



Yes, XDATA has a separate distance and time value. Also, you don't need MSECS since SECS is a double.

liversedge:
Exactly -- extract the data from the FIT file into XDATA and then you can use it in user data series or user metrics using the XDATA() function.

leifwar:
I've reverted the two commits in my repository (sent a pull request). Will start on importing R-R to XDATA. 

Kopi: "Leif Warland" <leif.warland@getmail.no>, "Author" <author@noreply.github.com> 
Sendt: 29. januar 2017 15:11:29 
Emne: Re: [GoldenCheetah/GoldenCheetah] Add more "sport" tags + read weight from FIT file. (#1651) 



Exactly -- extract the data from the FIT file into XDATA and then you can use it in user data series or user metrics using the XDATA() function.

PR #1659, opened 2015-11-13
Add support for Lap Swimming FIT files

amtriathlon:
The length message is used to generate 1 sample per second with speed,
cadence and distance data.
One test file is included.

It doesn't change the processing of other messages so it should be a pretty safe change for activities different from lap swimming, which don't use the length message.
I added a single user contributed test file since I don't have a device to generate more files, but I think we could add this patch to RC1 for user testing.


jonbev:
I have added a test file to my fork that doesn't seem to work very well. Its not far off but just not right. Your test file seemed to work well.

https://github.com/jonbev/GoldenCheetah/blob/93671977a7c9e3bdf19100f385133f9acabd6201/test/rides/lap_swim_test.FIT

amtriathlon:
Jon, it seems the 920xt also generates a record message previous to each length message which my code doesn't handle well, skipping them gives reasonable results but I need to do this only for lap swimming files. I will try to push a fix soon. Thanks for the testing and the file.

amtriathlon:
Jon, I think I fixed the problem with the 920xt files, if you want to give it a look it is here: https://github.com/GoldenCheetah/GoldenCheetah/pull/1664

jonbev:
That looks much better, thanks.

PR #1662, opened 2015-11-13
Monarksupport

erikboto:
This add support for a variety of Monark stationary bikes and should work with all LTx/LCx models, http://monarkexercise.se/productgroups/?lang=en&main_cat=sports. Since I do not have a LC-model myself I haven't implemented the support for controlling load yet, but I intend to do that with the help of other Monark users.

Since I started out using the Computrainer*.* I left the original copyright lines there, in case you wonder how those names ended up there.


liversedge:
WOW!

If we can get this into the v3.3 release candidate then that would be ace, since we can validate it doesn't break the other functionality.

The CI build fails on QSerialPort on QT4. I want to drop QT4 support soon but if you can make the support conditional on QT5 then that would resolve this for now ?

erikboto:
Right, forgot that QtSerialPort is Qt5 only. Too long since I did something Qt4.

I'll make it enabled on Qt5 only, would be awesome if it got into a RC-build so that I can point some other Monark users to it to test it out a bit more and with other models. Currently my windows builds are not supporting much of the optional libraries and I don't build nice installers etc.

liversedge:
If you look at Dropbox you will see we only add the source and headers on QT5 in src.pro

Joern-R:
Erik,

if you want to build the libraries on Window - I can give you infos for the libs - just drop me a note via mail  joern.rm at gmail.com

Joern

liversedge:
Thanks for the quick fixup.

PR #1663, opened 2015-11-13
fix bug: potential [out of range]

vlcvboyer:
we cannot have "curPosition = 0" as we will use [curPosition - 1] later.
Depending on the file which is used it creates a program termination based on faulty ASSERT in qVector::operator[] as i<0


liversedge:
will wait for the RC1 release to get built and announced then will merge for RC2

PR #1666, opened 2015-11-14
Add some basic up/down buttons to train mode

dresco:
Not sure what your thoughts will be on this one? 

I suspect there is not a huge appetite for adding to the current train mode toolbar. However, this should be low risk, and with FE-C making it into 3.3, at least provides basic support for altering workout intensity.. 

The icons are just a hack on one of the existing workout controls, but I think look passable at 64x64?


liversedge:
We need a slider and to fix the crappy window -- better that the controls are embedded/appear on the train view. There is no reason we cannot do this now since Views.cpp has a specific view for train view.

Will close this as don't think it is the right approach.

Sorry.

PR #1667, opened 2015-11-14
Add missing 'text' entry to parser

vlcvboyer:
Allows units to be displayed when overlay widgets are used on video training

This was part of initial video overlays development but unfortunately was not included in my commits.
It will be used later for messages such as trainer status (calibration required, trainer failure, trainer over its power limits...)

### DIFF
--- a/src/MeterWidget.cpp
+++ b/src/MeterWidget.cpp
@@ -92,6 +92,11 @@ void MeterWidget::paintEvent(QPaintEvent* paintevent)
     Q_UNUSED(paintevent);
 }
 
+void  MeterWidget::setColor(QColor  mainColor)
+{
+    m_MainColor = mainColor;
+}
+
 TextMeterWidget::TextMeterWidget(QString Name, QWidget *parent, QString Source) : MeterWidget(Name, parent, Source)
 {
 }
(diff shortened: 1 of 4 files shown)


liversedge:
merging in, but would be useful to have some very basic instructions on how to use the overlays?

Cheers,
Mark

vlcvboyer:
in GC wiki?

liversedge:
If you can that would be ace.

THANK YOU!

vlcvboyer:
I visited again GC website and it seems that you prefer video to detail the features?
Do you prefer a capture than wiki? this seems to be more friendly.
Otherwise do you use a specific software for wiki or just text editor?

Note: at present there is a #ifdef win32 in the code but normally it should be possible to make it work under linux (not Mac unfortunately due to libvlc for synchronisation but maybe overlay is ok...). Can you test it?

liversedge:
A video is 100x more effective than a wiki, so yes that would be excellent.

If you feel up to it, a quick expliner of the train view, adding devices, changing the screen layout and working with media would be absolutely outstanding !

I can post on my vimeo account since it support hd and videos of any length (youtube has limitations)

vlcvboyer:
I'll try to find the appropriate software to capture the screen with video and overlays...
And it has to work on my light computer used for training!

I'll let you add the sound track based on a description that I'll produce.
...English is not my mother language !

liversedge:
Thats ok, you should try. I can always recreate the video if you are 
unhappy with it :)

Cheers!

Mark

PR #1669, opened 2015-11-15
ANT: Cope with event counter wraparound

nickb937:
The event count wraps at 0xFF so we need a function eventCount()
to properly calculate the number of events when the overflow occurs
i.e.
0x00 (curr) - 0xFF (prev) = 0x01
0x01 (curr) - 0xFF (prev) = 0x02


nickb937:
hmm, actually probably not.

PR #1671, opened 2015-11-15
FixElevation: support API key, fix file descriptor leak

nickb937:
Support a user defined MapQuest API key via src/Settings.h to
workaround the overloading of the existing key.

Use exceptions to trap Bad Request and API errors and stop processing
early.

Delete the networkMgr object to prevent a file descriptor leak
which can cause GC to crash when fix-elevation is configured
for auto-processing on import of files and there are many files
being imported.

### DIFF
--- a/src/FixElevation.cpp
+++ b/src/FixElevation.cpp
@@ -126,24 +126,29 @@ FixElevation::postProcess(RideFile *ride, DataProcessorConfig *)
     QStringList elevationPoints;
     QString latLngCollection = "";
     int pointCount = 0;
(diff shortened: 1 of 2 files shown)


liversedge:
I have reimplemented this patch in commit 6e2d9593e5ccca43fbb5ffa155fd432f36375405
mostly the same but moved MAPQUEST key out of Settings and into gcconfig.pri and FixElevaton.cpp
It is not used anywhere else in the code so shouldn't be in settings.h.

PR #1672, opened 2015-11-16
Strava: move configuration into athlete-private.ini

nickb937:
When applying for API, the user must supply both the Client ID and
Client Secret in their API calls.  It was confusing to configure
Strava support one would add the CLIENT_SECRET in src/gcconfig.pri
and need to update the CLIENT_ID in src/Settings.h.

With the change, the Strava checkbox in the Share dialog is enabled
if the strava_client_secret is defined in athlete-private.ini.


liversedge:
These are compile time settings not user supplied/runtime.

nickb937:
There doesn't seem to be anything that blocks these being supplied at runtime if added to the Passwords dialog and they should be athlete specific

liversedge:
They are baked into public builds, so why would you need to provide at runtime ?

nickb937:
I'm not sure how I would know the Client Secret for Client ID 83 (the default) if I do not have the availability of a public build, or if the public build  I'm therefore compelled to apply for a Strava API key in order to submit requests and then that requires me to recompile to bake in Client ID and Secret so that the Strava upload option is available.

My point of reference of setting up Strava support is this thread: https://groups.google.com/d/topic/golden-cheetah-users/7-HePlapP4w/discussion

liversedge:
if you're not using a public build you will need to build your own. when you do this you compile in the api secrets.

nickb937:
It takes me back to me original point though: the Strava settings are in two different locations.  Either that should both be in src/Settings.h, src/gcconfig.pri.in or in the athlete-private.ini.  The latter would allow for a user to provide their own tokens via the Athlete Password, along with the other values stored there.

The token that you bake-in for the public builds can still be baked in via the default argument to appvalues->csettings()

Joern-R:
They should never be in src/settings.h since this violates Stravas rules:
http://strava.github.io/api/v3/oauth/ "All developers need to register their application before getting started. A registered application will be assigned a Client ID and Client SECRET. The SECRET should never be shared."

@Mark - to avoid someone build with it's own client/secrets has to modify settings.h what about a logic like this in settings.h

//Strava / see also Athlete parameter
#ifndef GC_STRAVA_CLIENT_ID
#define GC_STRAVA_CLIENT_ID    "83" // client id
#endif

Then both client/secret can be defined in gcconfig.pri to overwrite the public build settings. (Hope that would work - but have not tested.

liversedge:
sure that that would be ok, is probably a good idea actually.

Issue #1633, opened 2015-11-07
Detect trainer brakes limits


vlcvboyer:
Depending on technology maximum brake power of trainers varies (related to heat, speed...)
How to take it into account?
How to indicate it to user?
We can also indicate a lower speed than current one in order to get back in efficient range

PR #1674, opened 2015-11-16
WIP FE-C calibration and configuration

vlcvboyer:
this add FE-C spindown time calibration and cyclist configuration messages (weight...)
- configuration is automatically done using cyclist profile in GC
- calibration is ready but not activated at present as GUI is needed ("&& false" placed in the code and pointed out with FIXME comment)

### DIFF
--- a/src/ANT.cpp
+++ b/src/ANT.cpp
@@ -309,9 +309,60 @@ void ANT::refreshFecGradient()
         sendMessage(ANTMessage::fecSetTrackResistance(fecChannel, gradient, currentRollingResistance));
 }
 
+bool ANT::getFecPowerCalibInProgress() const
+{
+    if (fecChannel == -1)
+        return false;
+
+    return antChannel[fecChannel]->getFecPowerCalibInProgress();
+}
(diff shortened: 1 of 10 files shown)


vlcvboyer:
There is an issue when not compiling on win32...
I'll investigate.

vlcvboyer:
Please wait before merging: I'll put my code on top on dresco code

liversedge:
No problem -- let me know when you're both happy :)

liversedge:
I would really like to get this into V3.3 if we can.

vlcvboyer:
Yes we will do our best to have it in V3.3!

vlcvboyer:
FYI only this is also related to #1633

vlcvboyer:
Just to keep you informed:
it goes forward but still some issues to fix properly (concern when multiple devices are used)

liversedge:
Any updates?

I think Jon E is working also on calibration .. wonder if it fits in with this?

liversedge:
Any news?

PR #1681, opened 2015-11-19
TrainDB -  fix //1 remove workout error

Joern-R:
... caused error in list-model since entries are added and "end-of" dbtable
... now only update of texts in upgrade to handle pre 3.3 translation problems


liversedge:
Excellent !

PR #1683, opened 2015-11-20
Monark: Add support for controlling power on LCx models

erikboto:
Since there's no manual control lever for resistance on LC-models (there is on my LT2) the previous "read only" support wasn't really useful on those models since the load was stuck on ~15W. I've added load control for LC-models and have gotten help from LC owners to confirm it's actually working. Would be awesome to get this in a 3.3 RC and eventually 3.3 if possible.

### DIFF
--- a/src/Monark.cpp
+++ b/src/Monark.cpp
@@ -26,10 +26,10 @@ Monark::Monark(QObject *parent,  QString devname) : QObject(parent),
     m_power(0),
     m_cadence(0)
 {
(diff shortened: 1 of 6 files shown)


liversedge:
Oh absolutely, since it is isolated to Monark support its no problem, there will be no regression impact.
I'll wait for the CI build check to complete and then merge.

Thanks again.

erikboto:
Awesome, thanks!

PR #1685, opened 2015-11-21
fix ANT device reset requence

vlcvboyer:
Take into account acknowledge from device during ANT reset sequence:
if ack not received after 500ms (kept from current code) then try again (3 times max)
if still not received we do not insist and go forward (ensure compatibility with current code)

I have one adapter which sometimes do not reset on the first attempt.
This patch solved my issue

### DIFF
--- a/src/ANT.cpp
+++ b/src/ANT.cpp
@@ -427,10 +427,20 @@ ANT::setup()
     // fixme: better synchronisation?
     msleep(500);
 
-    sendMessage(ANTMessage::resetSystem());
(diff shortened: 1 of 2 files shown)


liversedge:
I would like this to be tested !!!!
I wonder if Jon would like to try it out ?
Will send an email.

dresco:
I've not been able to force a failure to reset, but I see no issues with this patch. Thanks!

liversedge:
Perfect, will merge as it may solve issues others have reported.

PR #1692, opened 2015-11-23
Fix W' zone percentage calculation.

### LINKED ISSUE #1691: If WPrime duration is 0, NaN is displayed in the summary field.
Description says it all.

gille:
If no time was spent in the zone output it as 0% not NaN.

Fixes #1691


liversedge:
Thanks. Damn you DIVZERO.

PR #1694, opened 2015-11-24
Update .gitignore to ignore the .prl file that is dynamically generated.

gille:


liversedge:
The .prl file is also changed in this commit ?
Any chance you can submit with just the .gitignore changes?

gille:
I could I suppose, the change as far as I see is because I deleted it from the repo.

gille:
Ok, I've reverted libkqoauth.prl so now it just includes the .gitignore.

liversedge:
Thanks!

PR #1701, opened 2015-11-28
fix : 'break;' was commented at same time as function

vlcvboyer:
line comment '//' was disabling 'break;' on the same line
thereby decodeSession function was called for any devideInfo message


liversedge:
oops! good spot

Issue #1548, opened 2015-09-04
Add Garmin segments as intervals


peteonbike:
If I could work out the coding for this I would......
Garmin have introduced segments which are basically an auto start/stop interval. When you open a garmin fit file they are recorded in the same way an interval is except they are labelled as "segment" and have a segment name.
How hard would it be to show these in golden cheetah?
thanks
ps version 3.2 is brilliant!

gille:
Would you be willing to share a .FIT file (I assume that's format) containing this? either to the bug or just email me (mgille at the google mail service).

PR #1702, opened 2015-11-28
Add more FIT files data fields description

vlcvboyer:
This is mostly to write down / describe all other known data that are available in FIT files (useful as most are not documented in ANT specifications).
- Will be used for future work related to #1548 (but this will not be so easy because it seems that when segment is indicated in FIT files this is at begining only, Then we have to open the segment FIT file on device to get track points and validate is user has followed the segment from start to end. Otherwise we will not ensure to have determined expected interval)
- contain par of #1617 (power estimation is not included as the objective of current patch is mostly to add description of FIT files to source code not to estimate power)
- maybe will have to be used to fix activity time (timezone, timezone difference and UTC are now in source code). Not sure I have not investigated yet.

### DIFF
--- a/src/FitRideFile.cpp
+++ b/src/FitRideFile.cpp
@@ -1268,6 +1268,30 @@ struct FitFileReaderState
         }
     }
 
+    void decodeDeviceSettings(const FitDefinition &def, int time_offset, const std::vector<FitValue> values) {
+        Q_UNUSED(time_offset);
+        int i = 0;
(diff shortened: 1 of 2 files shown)


liversedge:
Dang. Want to merge but it has conflicts with the segment commits.
Any chance you can fix the conflicts?

vlcvboyer:
I'll check just now

vlcvboyer:
Conflicts are solved.

PR #1704, opened 2015-11-28
Add segment support to Garmin files

### LINKED ISSUE #1548: Add Garmin segments as intervals

gille:
Fixes #1548


liversedge:
any chance you can rebase ?

vlcvboyer:
Hi,

I do not understand your code as is seems to assume that each time we cross a segment start point we start a new lap. But in fact segment end point is not at same location than another segment end point. Segment is not a lap but track portion. Thus when we cross a start point we have to memorize the end point location (latitude/longitude) and when we cross this end point (during records parsing) then we can assume that we reach end of segment. It not so easy because we have to define a distance as we will never cross the path at exact end segment location and we should also take care of segment track points to ensure that athlete have followed the path (segment track points are located in another FIT file).
Maybe my analysis is based on differents records than yours. (I use edge 1000 during road cycling).

FYI: I've just added comments in the code related to FIT files (see #1702).
It's mostly comments in order to avoid other developpers to spend hours to analyze FIT messages.
Segment details were included (not so much because some fields are still to be analyzed).

Best regards

Vianney

vlcvboyer:
Additional comments:
1/ We can also connect to strava API in order to get segments when ride is completed. This API is quite clean and easy to access. Segments done during an activity are indicated. This requires user to upload to Strava prior to import in GC. On recent device this is done automatically when ride is completed and wifi available.
2/ Maybe this will create too much intervals...
3/ This might solve another issue: Garmin segments have to be uploaded to garmin device prior to ride. Thus if we have not planned our ride we will not have uploaded the segments thus only previous uploaded ones will be part of FIT file. (I don't know how it works when using strava live segment)
4/ On my side I planned to import strava segments (and gamin through segment FIT file stored on device) in GC as an additional database (at present I've done it in Excel...) but not sure this will be useful for users. Typically, on my side, I use it to compare one specific segment depending on bike which was used. For accurate compare we have also to take care of wind (which is available on garmin device when connected to smartphone or by using a dedicated weather history API which is not free of charge)...
5/ We can maybe also download segments from Garmin API but it is not free of charge!

liversedge:
Just an FYI Marcus,

We're not usually such a tough crowd!!!!

This patch hits a sensitive aspect (1) fit file format is a PITA (2) segments and Strava / GC are still a WIP from a design and UX perspective.

Its not about your code per se, as much as working out what we should code.

vlcvboyer:
The first commit is not related to FIT files segments but code optimization
Mark, what do you think about splitting the patch into 2 differents patches andapply the 1st one at present?

To answer your 1st question: parsing segment messages from FIT files should not create any regression issue as those message id were not used by earlier devices (maybe, but unlikely, another manufacturer will use the same id for something different than segment but this is another story...)
However this does not answer the main question: how to use segment data... After few hours thinking about it my opinion is: create intervals when last point of segment is reached by athlete. This means keep in memory all segments and their respective end track point, and compare to curent athlete position when parsing track point records, etc.

liversedge:
I think the changes are good, gonna merge as is.

PR #1715, opened 2015-12-04
dp: FixMoxy add config options (fix BSXInsight support)

dmonakhov:
Number of data fields depends on sensor model
MoxySensor transmit SMO2 as cadance, tHB as speed,
BSX Insight 2 transmit only SMO2 as cadance, tHB is not supported.

So in case of BSXInsight we should not convert speed data. Let's allow
user to chose which datafields to transfer.

This patch add two new check buttons FixMoxy dialog
 - Cadence to SMO2
 - Speed to tHb

### DIFF
--- a/src/FixMoxy.cpp
+++ b/src/FixMoxy.cpp
@@ -23,6 +23,7 @@
 #include "HelpWhatsThis.h"
 #include <algorithm>
 #include <QVector>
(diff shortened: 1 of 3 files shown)


liversedge:
Thanks, and thanks for making it follow the right approach for config.
Will merge once the travis build completes

PR #1730, opened 2015-12-18
Fix FLEX version checks for 2.6.0

flacjacket:


vlcvboyer:
I've done it for DataFilter.l (see #1728). I've not used exactly the same code.
Maybe you'll be able to use the same than mine in order to improve overall code consistency?
I should have check other files at that time... sorry about that!

flacjacket:
Ah, I didn't see your changes to DataFilter.l @vlcvboyer, I could change my over in a new PR if that's wanted.

liversedge:
would be nice to be consistent :)

maybe add it to a new include ?

vlcvboyer:
When I saw this patch I check across all code. There is no more flex source to fix.
Include is maybe too much for few lines!

@flacjacket: are you able to use flex 2.6?
Thanks to those fix I don't have any more building error but flex itself freeze/crashes when compiling. I don't have any issue with Flex 2.5.9 which works well with the same source code.

vlcvboyer:
I found the root cause:
The original file on github contains CR/LF instead of LF only. This is not appreciated by flex which freeze on my platform. When I change all CR/LF to LF only it compiles without any issue.
I assume that there is no reason to use CR/LF.
I cannot change it by submitting a patch as git do not consider CR/LF as a modification (maybe there is a commandline option for that?).
Thus at present it's only changed for lines modified by my patch which is already merged. I had to modify manually all other.
Mark, are you able to change whole file on github in order to remove those "CR"?
DataFilter.l only is concerned. Other flex files are correct and LF only.

liversedge:
So flex can't read files written on Windows ?!?!?!?!?
Sounds like flex needs fixing to me........

vlcvboyer:
This is just what I saw !
flex 2.5.9 read both files (LF only and mixed CR/LF)
flex 2.6.0 freeze on my computer with CR/LF file and succeed with LF only
I don't know why and maybe this unexpected freeze is platform specific.

However in GC source code there is only one CR/LF file and others are LF only.
...you know that I like consistency!

liversedge:
Sure :)

Its because git thinks it is an msdos file -- its easy to fix though, leave it with me and I'll fix it.

liversedge:
^M (CR) removed.

PR #1735, opened 2015-12-18
cosmetic tab & hex values

vlcvboyer:
align values and use only hex constants to improve consistency


liversedge:
We align to the protocol docs in most cases for greater clarity.

vlcvboyer:
I agree !
...BUT in this #define section there is a mix between decimal and hex values: 3 decimal vs 9 hexadecimal ones.
In specification documents both are shown each time but only hex is used in associated examples.

PR #1739, opened 2015-12-20
Add support for running power zones (Stryd) part 3

amtriathlon:
Final part: use the zones according to sport in metrics and charts
Also fixes a recent bug in PaceZoneTime


liversedge:
Whoa you've been busy :+1: 

The model plot needs a quick fixup.

amtriathlon:
Sorry, my build did not included 3d plots, I added a commit to fix it.

PR #1746, opened 2015-12-25
Fixed a german translation in AboutDialog to have a fixed length

maxammann:
Before:
2015-12-25-12 14 56-screenshot

After:
2015-12-25-12 04 28-screenshot

Problem: Looky ugly and could crash the programm if protected memory is accessed.


liversedge:
Happy Christmas !

maxammann:
Thanks, happy christmas, too :D What's better than fixing a bug on christmas morning? :D

liversedge:
:+1:

PR #1748, opened 2015-12-29
Add support for syncing to GoogleDrive.

gille:
This enables user to store GoldenCheetah files in their Drive accounts. The files are stored in the GoldenCheetah application data folder and are not visible to users. The directory name can be anything but must not conflict between athletes.

This uses GoogleDrive v2 API. The pull request also includes some fixes to OAuth handling for Google accounts in general.

### DIFF
--- a/src/Dropbox.cpp
+++ b/src/Dropbox.cpp
@@ -184,7 +184,7 @@ bool
 Dropbox::readFile(QByteArray *data, QString remotename)
 {
     // this must be performed asyncronously and call made
-    // to notifyReadCompleted(QByteArray &data, QString remotename, QString message) when done
+    // to notifyReadComplete(QByteArray &data, QString remotename, QString message) when done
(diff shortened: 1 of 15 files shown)


liversedge:
Awesome!

The build fails because there are no defaults for the SECRETS and KEYS.

These should be defined in Secrets.h (without the actual values) .. see the ones already in there.
We replace them (via sed) in Travis with gem secrets and users can define them in gcconfig.pri for local builds.

Really looking forward to playing with this :)

:+1:

gille:
Ok. Just need to revert some of the JSON handling in the non QT5 code.

gille:
And now it builds! Let me know if you want me to squash the commits.

I added directory support, even though directories in GoogleDrive don't behave like they do in your average file system.

liversedge:
Excellent!

PR #1753, opened 2015-12-30
MSVC - QWT Debug vs. Release

Joern-R:
... MSVC does not accept mixed libraries (debug/release) when linking
... patch has been tested with MingGW/GCC an MSVC


Joern-R:
Fixed throug PR https://github.com/GoldenCheetah/GoldenCheetah/pull/1758

PR #1759, opened 2015-12-31
Improved gitignore, ignore patch leftovers and emacs temporaries.

gille:


liversedge:
got any interesting emacs macros etc for qt developers?

gille:
Not really. I use the very basics.

PR #1769, opened 2016-01-03
ANT: log sent messages

vlcvboyer:


dresco:
That's spooky, I created an almost identical patch today (while debugging some oddities around the master channel for remote controls). Just went to create a pull request for it, and was there already :)

liversedge:
great minds think alike. :+1:

vlcvboyer:
I also created timstamp patch but you submitted yours just before me...

PR #1773, opened 2016-01-04
fix_tacx_vortex

vlcvboyer:
Should fix the Tacx Vortex "fake" speed event timestamp (when simulating real speed sensor few messages have different timestamp related to same wheel revolution event).
Tested on linux with speed and cadence sensor actuated by hand without any issue (regression test).
Note: I'm unable to test it with Tacx Vortex during more than few seconds at present (broken toe & laptop hardware failure).
Thus any intensive test with vortex will be welcome!

### DIFF
--- a/src/ANTChannel.cpp
+++ b/src/ANTChannel.cpp
@@ -622,7 +622,8 @@ void ANTChannel::broadcastEvent(unsigned char *ant_message)
                static float last_measured_rpm;
                uint16_t time = antMessage.crankMeasurementTime - lastMessage.crankMeasurementTime;
                uint16_t revs = antMessage.crankRevolutions - lastMessage.crankRevolutions;
-               if (time) {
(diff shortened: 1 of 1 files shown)


dresco:
Hi, that didn't help unfortunately. I have a better idea of what's going on though, will get an email out tonight.. Cheers, Jon.

dresco:
Hi, finally noticed this in the logs from my trainer (and confirmed against Todd's).

I don't think a fix is required. Agree the data is odd, however it's only incrementing the cadence timestamps (and therefore zeroing cadence) when the speed is zero - so has no actual affect on training data..

liversedge:
Should this PR be closed?

dresco:
Personally I'm not convinced it's needed, as the cadence timestamps only appear to be incremented while speed is zero.. Cheers, Jon.

Issue #1774, opened 2016-01-05
Add FIT export support (for e.g L/R power balance)


bdb:
It appears that GC will consume L/R power balance from .fit files, but then there is no way to get that information back out with any of the currently supported export file formats. Could .fit export be added or perhaps the TCX or PWX formats updated to support L/R balance?

It's certainly possible to just use the original fit file, but that does not work if there's garbage (spikes, etc) in it that needs to be corrected using GC.

liversedge:
Neither the TCX not the PWX formats support LR balance.

PR #1777, opened 2016-01-06
Fix bugs and warnings

bdb:
Fixes a bunch of warnings that Xcode 7 produced (and some bugs found) while working on C++11 support. Note C++11 is NOT enabled in this branch. Built and tested on 10.11.2 only. I did not test Linux build or Win build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,14 @@ Makefile
 .DS_Store
 profile
 
+# qmake Xcode generator side-effects
+*.xcodeproj
+xcshareddata/
+*.xcworkspace
+*.pbxproj
+*.mak
+*.build/
+
 moc_*
 moc/
 obj/
--- a/qwt/src/qwt_date.cpp
+++ b/qwt/src/qwt_date.cpp
@@ -160,36 +160,7 @@ static inline QDateTime qwtToTimeSpec(
     }
 
     return dt.toTimeSpec( spec );
-}
-
-static inline double qwtToJulianDay( int year, int month, int day )
-{
-    // code from QDate but using doubles to avoid overflows
-    // for large values
-
-    const int m1 = ( month - 14 ) / 12;
-    const int m2 = ( 367 * ( month - 2 - 12 * m1 ) ) / 12;
-    const double y1 = ::floor( ( 4900.0 + year + m1 ) / 100 );
-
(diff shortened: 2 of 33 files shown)


liversedge:
Will need to read through this one, so will look on saturday when have a bit more time.

Thanks for taking the time to do this, are there any changes in there that you think are worth highlighting (regression impact / refactor would be better etc) ?

liversedge:
**You will need to remove any changes to qwt**, since we take the base qwt libs and apply a >8  axis patch -- which we have to do when a new qwt release comes out.

So we need to minimise any changes we make. There may be an argument now to stop including qwt since it supports >8 axis now (and the code was changed to use that).

bdb:
There is an actual Qwt bug fix: https://github.com/bdb/GoldenCheetah/commit/3143c5ee4d4df080a9d3252f4ab36526ccd1cdd7

I don't know what impact it has on GC. But you can see the very same bug that I highlighted in #1774. Basically those addDates() are NOPs when they ignore the return value.

The other stuff was just warnings. It would however be best to compile warning free. Otherwise you can have bugs hiding in the noise of mostly benign warnings.

bdb:
The actual bugs:

https://github.com/bdb/GoldenCheetah/commit/3143c5ee4d4df080a9d3252f4ab36526ccd1cdd7
https://github.com/bdb/GoldenCheetah/commit/27a56332267fea147d693b0873a2c316388f3a70
https://github.com/bdb/GoldenCheetah/commit/f457a9cbb8b5275dc6d857df50d23d9c1ee27128

I would consider this a bug, since accessing it does not truly overload its parent method and thus accessing through a base type will not call the overloaded method (well its undefined what would really happen).
https://github.com/bdb/GoldenCheetah/commit/50db4196bf54e5f52b8bc5a50ea79904d24c37a0

And this one is a compile error when C++11 is enabled:
https://github.com/bdb/GoldenCheetah/commit/d8155c1ad100ce72da858dce75c664d9ccce3d09

liversedge:
OK, lets merge your fixes for qwt and recognise that we should no longer keep a local copy -- rather we should link to a standard qwt lib (so before v4 is released we should remove it from the GC repo and build against the latest from Uwe).

liversedge:
Thanks for all of these, some of them are embarrassing (!)

PR #1780, opened 2016-01-06
MSVC - Memory leak

Joern-R:
... missed delete[] of allocated arrays

### DIFF
--- a/src/Aerolab.cpp
+++ b/src/Aerolab.cpp
@@ -874,6 +874,11 @@ QString Aerolab::estimateCdACrr(RideItem *rideItem)
             } else {
                 errMsg = tr("At least two segments must be defined");
             }
+#ifdef Q_CC_MSVC
+            delete[] X1;
+            delete[] X2;
+            delete[] Egain;
+#endif
         } else {
             errMsg = tr("Altitude and Power data must be present");
         }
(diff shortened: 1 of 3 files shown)


bdb:
If you are wrapping these strictly for MSVC, there's no need to use naked new for buffers, use std::unique_ptr instead. I have branch to enable C++11, so that the whole project can start moving to modern C++.

Joern-R:
Improvements/learnings are very welcome - focus for me now was to fix the
missing delete first. If we get a common approach for both GCC/... and MSVC
using C++11 for sure better.


2016-01-06 17:31 GMT+01:00 Brian Bergstrand <notifications@github.com>:

liversedge:
QSharedPointer does something similar. We must be careful using these with QT objects, but for heap allocation within a single method it could be used. There /should/ be very few situations where this is needed.

bdb:
I would say Q ptrs should not be managed at all. Just from the name, I'm thinking QSharedPointer has overhead, unlike std::unique_ptr (as long as the Deleter is stateless).

I don't know the code base well enough to say how useful it would be, but Fit parsing could benefit from it as well as all the changes Joern made -- though I'm not sure why new's are required there instead of just using the stack. Joern, could you comment on that?

liversedge:
I think the point here is the original code used variable sized C++ arrays when it really would have made more sense to just use QT containers.

The original changes were to get past MSVC issues, refactoring to remove the arrays is the best answer. Just not the focus at the time.

gille:
QScopedPointer should have low overhead, at least small enough to not matter (we copy entire vectors back and forth). Otherwise I'm a huge supporter of c++11.

PR #1781, opened 2016-01-06
ANT fixes

dresco:
Don't think there is anything too controversial here, except perhaps 7b09d3a. Not sure whether it warrants further discussion on the dev list or whether you're happy as it stands?

The reason for the change is that I was seeing significant interference to remote control input from any searching channels. There's some background on the two scanning types at the following links;

https://www.thisisant.com/resources/an11-ant-channel-search-and-background-scanning-channel/
https://www.thisisant.com/resources/an15-multi-channel-design-considerations/

Haven't seen any downside to using low priority only. It resolves the channel interference issues, and I've tested searching for up to 8 active sensors/channels in the AddDevice dialog.

### DIFF
--- a/src/ANT.cpp
+++ b/src/ANT.cpp
@@ -842,6 +842,10 @@ ANT::slotStopBroadcastTimer(int channel) // timer
 
     //qDebug()<<"ANT::slotStopTimer req from channel "<<channel;
 
+    // disconnect the slot, else we duplicate signals on subsequent sessions
+    disconnect(antChannel[channel]->channelTimer, SIGNAL(timeout()), this, SLOT(slotControlTimerEvent()));
+
+
     // stop the broadcast timer..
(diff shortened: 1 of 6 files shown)


liversedge:
I agree with the low priority search, it shouldn't have a negative regression impact either.

stschake:
Probably related: I just tried a device profile where I left one channel dangling (HR sensor not transmitting, so GC will keep searching for it) and the FEC had massive packet loss, taking 30 seconds for a load setpoint packet to eventually get through, with the ANT USB transmitter a mere 20cm away from the trainer.

dresco:
Yes I think it will be the same thing (I first noticed it with an inactive HR sensor also). The original behaviour was (default) 5 seconds of low priority search, followed by 30 seconds high. Seems like acknowledged messages are particularly badly affected.

PR #1782, opened 2016-01-07
C++11 conversion

bdb:
I'm opening this for comment and discussion. As it stands right now, I've enabled C++11 and fixed the few resulting errors. I also did a basic conversion for FitRideFile as an example. This was split into 2 commits (unique_ptr RAII and general C++11 semantics).

Note that this first pull request includes the commits from my other 'fix-bugs-and-warnings' request since some of those are required for C++11. So the actual C++11 changes begin with the 'Require C++11' commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,12 +32,12 @@ before_install:
 - curl -O http://www.ftdichip.com/Drivers/D2XX/MacOSX/D2XX1.2.2.dmg
 - git clone --depth 1 https://github.com/sintegrial/qwtplot3d.git qwtplot3d
 - cd qwtplot3d
-- CC=clang CXX=clang++ /usr/local/opt/$QT_PATH/bin/qmake -makefile -recursive QMAKE_CXXFLAGS_WARN_ON+=-Wno-unused-private-field
+- CC=clang CXX=clang++ /usr/local/opt/$QT_PATH/bin/qmake -makefile -recursive QMAKE_CXXFLAGS+=-stdlib=libc++ QMAKE_CXXFLAGS_WARN_ON+=-Wno-unused-private-field QMAKE_LFLAGS+=-stdlib=libc++
 - CC=clang CXX=clang++ make -j2
 - cd ..
 - git clone --branch 0.98 https://github.com/kypeli/kQOAuth.git kQOAuth-0.98
 - cd kQOAuth-0.98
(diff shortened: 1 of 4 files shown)


bdb:
Hmm, QWT3D is failing to link for the Qt5 build (though it links for the Qt4 build). Is that an installed library or built from source? (I don't have it installed/enabled locally.)

liversedge:
I suspect ABI issues with c++11.

bdb:
OK, odd that Qt4 job built and passed but not 5. I'll try to install it locally.

liversedge:
I think QT4 builds it and QT5 doesn't. but not sure.

bdb:
Ah, that would explain it. So the travis setup needs to change to build the library instead of installing a binary (from hombrew or somewhere?).

liversedge:
Just checked and it is build from source for both.

bdb:
Thanks. It probably needs to built as C++11 too for some reason due to Qt5 changes. I'll look into it tonight after work and workout.

bdb:
OK, sorted it out, C++11 did not have to be enabled for the external libs, but libc++ did have to be used. Xcode 7+ defaults to that, I guess earlier versions do not.

I was thinking too; if you are interested in this, would it be OK to just require C++14. Not a big deal for Xcode and MSVC, but it would limit GCC to 4.9.2+. Generic lambdas, and move captures would be the big things and it would be nice to have auto return type too.

liversedge:
Trolltech have announced that from QT 5.7 c++11 will be used by QT modules and will be mandatory -- so we should be ready for this in the next release (4.1 2017)

liversedge:
Hi,

QT 5.6 now requires c++11 so we can start conversion, but we still need to support non c++11 compilers too.

Are there any changes in here that you feel warrant the effort ?

Cheers!

Mark

bdb:
Conversion can be done piecemeal. This initial patch was just enough to get things going. However if you still have to support  non-C++11 compilers than I'm not sure what you are asking. Conversion is all-or-nothing, so GCC 4.8+, Clang 3.3+ and MSVC 2015 would be the minimum supported versions.

liversedge:
Yeah I was on that page. It will need to wait for QT5.6 support across the board.

Thanks for the advice.

PR #1783, opened 2016-01-07
Strava

gille:
Fix private file uploads - I made this a choice.
Let Strava upload .tcx.gz instead to save bandwidth

+ support for commute and trainer workouts on Strava.
+ move some files around for the keys.

### DIFF
--- a/src/Pages.cpp
+++ b/src/Pages.cpp
@@ -42,6 +42,7 @@
 #include "GoogleDrive.h"
 #endif
 #include "LocalFileStore.h"
+#include "Secrets.h"
 
 //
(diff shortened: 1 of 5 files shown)


liversedge:
Is that commute/trainer metadata specific to strava or something specific to you?
It feels wrong! Should be metadata that the user maintains against the ride...

liversedge:
Otherwise the compress/private stuff is great!

gille:
It's specific to Strava AFAIK. Maybe others support this distinction as well but I agree, it should be moved to ride metadata.

liversedge:
ah ok. lets go with it as it is.

thanks.

PR #1784, opened 2016-01-09
CloudDB - LTM Charts - multiple changes

Joern-R:
... move Chart Import to Library/Sidebar (and remove from LTMTool)
... enhance Import Dialog / First Filter function - Curated
... introduce local Cache to not re-read all time
... read/display in chunks of 10 charts from GAE
... simplify data structure (in sync with CloudDB)
    ... no statusId, no explicite versioning

... Refactoring of File and Class Names to provide a common
    structure for future CloudDB artifacts


liversedge:
I get: Object::connect: No such slot CloudDBChartImportDialog::addAndCloseClicked(QTableWidgetItem*) in CloudDBChart.cpp:417
 
:)

Joern-R:
I don't even though I should have the same problem. But I wanted to remove the "double-click" thing anyway. Will just do it now and send the fix.

PR #1790, opened 2016-01-11
Added virtual power support for the Blackburn Tech Fluid trainer

max-f:
Hi, 

got the equation as described [here](https://github.com/max-f/virtual_power_curve/blob/master/virtual_power_blackburn_tech_fluid.ipynb). Thanks for all the work btw!


liversedge:
thanks!

PR #1791, opened 2016-01-13
Xcode7 compiler warning changes

cleeland:
clang from xcode7 on yosemite made lots of noise about various things.  I fixed most of them with these commits.  Hopefully they don't break other platforms.

### DIFF
--- a/src/ANT.cpp
+++ b/src/ANT.cpp
@@ -1132,11 +1132,11 @@ int ANT::openPort()
     return 0;
 }
 
-int ANT::rawWrite(uint8_t *bytes, int size) // unix!!
+int ANT::rawWrite(uint8_t */*bytes*/, int /*size*/) // unix!!
 {
+#ifdef WIN32
     int rc=0;
 
-#ifdef WIN32
 #ifdef GC_HAVE_LIBUSB
     switch (usbMode) {
 #ifdef GC_HAVE_USBXPRESS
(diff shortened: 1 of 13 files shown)


liversedge:
These changes are rather problematic as they are to fix warnings based upon your build config and O/S:

a) they break other platform support, notably Windows
b) they break OSX support if optional libs are available e.g. LibUsb
c) the CI fails

I won't merge this, but will look at each of the changes to see how we might want to proceed.

liversedge:
Brian beat you to it on a lot of these :)
I've just merged his fixes: 183ba72658abf4807ee6b183c74651edd5683aae

cleeland:
Is there a way that I can see the results of the builds on other platforms?  I presume they're from travis CI?

I'm confused about the libusb comment, because I actually have libusb configured in my build.

liversedge:
The travis results are shown alongside the pull request
https://github.com/GoldenCheetah/GoldenCheetah/pull/1791

liversedge:
Although it looks like they are removed when the pull request is closed.
See here: https://travis-ci.org/GoldenCheetah/GoldenCheetah/builds/102028182

cleeland:
They are still there--just not very obvious.  There is a red "X" that appears to the left ofthe short hash for the most recent commit in my chain.  If I click on that I'm taken to the build results.

I need to figure out what is different btw. my environment and the travis CI because at first glance it appears that they should have been pretty equivalent.

PR #1794, opened 2016-01-16
Cloud DB - various enhancements

Joern-R:
... T&C acceptance Popup for the CloudDB service (test will need review !)
... Enhancement of Publish Dialog (checking field content, mandatory field, mandatory length if input)
... Add Caching for Charts and Chart Header to minimize traffic to GAE
... Enhancement of Retrieval Dialog (adding multiple filter and search option(
... Handle "out of quota" responses

### DIFF
--- a/src/CloudDBChart.cpp
+++ b/src/CloudDBChart.cpp
@@ -72,28 +72,28 @@ CloudDBChartClient::~CloudDBChartClient() {
 }
 
 
-bool
+CloudDBResponse
 CloudDBChartClient::postChart(ChartAPIv1 chart) {
 
     // check if Athlete ID is filled
 
-    if (chart.CreatorId.isEmpty()) return false;
+    if (chart.header.CreatorId.isEmpty()) return CloudDBResponse::Others;
 
     // default where it may be necessary
-    if (chart.Language.isEmpty()) chart.Language = "en";
+    if (chart.header.Language.isEmpty()) chart.header.Language = "en";
 
     // first create the JSON object
     // only a subset of fields is required for POST
     QJsonObject json;
-    json.insert("name", QJsonValue(chart.Name));
-    json.insert("description", chart.Description);
-    json.insert("language", chart.Language);
-    json.insert("gcversion", chart.GcVersion);
+    json.insert("name", QJsonValue(chart.header.Name));
+    json.insert("description", chart.header.Description);
+    json.insert("language", chart.header.Language);
+    json.insert("gcversion", chart.header.GcVersion);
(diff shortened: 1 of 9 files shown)


liversedge:
There are a couple of issues here.

a) it uses c++11 semantics for enums and static member variables that break c++98 compilers (I really don't want to introduce that dependency yet, it will cause problems to devs on Linux).

b) it assumes Qt5.6, is that intentional? (we need to be clear if we only support this in qt5.6 (e.g. not 5.5 or lower)

liversedge:
most of the issues with enums (cloudDBResponse) and initialisers can be resolved via typedef and use of a static iniitialiser, e.g. static bool cloudDBinit = initCloudDB(); static bool initCloudDB() { ... } in CloudDBCommon.cpp

Make sense?

Joern-R:
C++11 specifics were not intended.

How this shall be handled related to which "official" builds we are
producing / on which QT version / with which compiler.
Due to the specific "Secrets" associated with the use of CloudDB - no
private build will be able leverate the shared CloudDB
functionality anyway - and if someone runs an own CloudDB on GAE he/she has
use the proper QT and C++ version.

As C++11, QT 5.6 - was not intended. But the same question. If we assume
our offical V4 build being done with QT 5.6
this is not limitation - if there will be offical builds below QT 5.6 or
with compilers where C++11 is not available that need
to be changed.

What are your thoughts?

2016-01-16 11:57 GMT+01:00 Mark Liversedge <notifications@github.com>:

liversedge:
I'm happy to mandate QT5.6 for QT5 based builds (and our official builds will ONLY be delivered this way) when it is released. I still want to be able to support QT4 builds but new features may not be available. Is the 5_6 datastream /really/ needed ?

If CloudDB needs QT5 then we should make sure the src.pro/gcconfig.pri stuff enforce that. If only just for Travis ! (At present Travis doesn't have cloudDB enabled, so it didn't pick up these issues. We can add it, and will need to make it only available on QT5.6 if neccessary)

I think we should avoid C++11 semantics unless we absolutely need them, or there is a significant benefit to using them. The C++11 stuff is clearly just by accident so probably good to fix that up :)

Joern-R:
Will remove the C++11 semantics - and also the QT5_6 in QDataStream - it's not needed.
Not sure how to adjust config.pri/src.pro to enforce QT 5.

liversedge:
I'll fix that (its really easy!) and will add cloudDB to QT5 Travis builds :)

I'm reffing today so will do later. (It is freezing here!!!)

PR #1795, opened 2016-01-16
CloudDB - C++98 compatibility

Joern-R:
... don't use C++11 specific features


liversedge:
Joern,

You totally rock. I just went through the T&C for clouddb and its perfectly done. You're several steps ahead of me in thinking and execution.

Thank you !

Mark

PR #1802, opened 2016-01-17
Fixes for Fortius USB timeout issue and new Fortius VR driver package…

WaterworthD:
Fixes usb read timeout on some machines and adds driver inf for Fortius VR which is signed and installs properly on windows 10 x64

- Fortius.cpp was reading from device before writing to it. This can cause device to error
- LibUsb.cpp contains hard coded usb timeout

Added read / write override to LibUsb so can specify different timeout to the default ANT+ 125ms

Restructured run loop in Fortius.cpp

Also added Fortius VR inf package (libusb-0.1) to \contrib built with Zadig

### DIFF
--- a/src/Fortius.cpp
+++ b/src/Fortius.cpp
@@ -372,7 +372,19 @@ void Fortius::run()
 
         if (isDeviceOpen == true) {
 
+			int rc = sendRunCommand(pedalSensor) ;
+			if (rc < 0) {
+				qDebug() << "usb write error " << rc;
+				// send failed - ouch!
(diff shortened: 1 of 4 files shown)


liversedge:
Thanks Dave,
When you say on some machines, what machines have problems ?
Mark

WaterworthD:
I didn't realize C++ had default parameters, more familiar with C#, thanks

Current build doesn't work on either of my machines - windows 10 x64

It only has usb 3,0 ports, not sure if that's the issue.

Issue #1798, opened 2016-01-17
Crash when "Copy to Library" selected in Tacx Fortius device wizard


WaterworthD:
In AddFirmware::AddFirmware() when Copy to Library checkbox is enabled GC exiis with attached crash log
[tacx.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/93195/tacx.txt)
 

Also descriptive text "On windows it is typically installed in C:\\Windows\\system32" is misleading - it's more commonly found in the the TTS installation folder

And Fortius device wizard should split into separate paths for the Fortius (1942) and Fortius VR (1932) PID's and only ask for the firmware for the pre-VR model

Dave

(feel free to assign to me, just logging)

WaterworthD:
Issue the following line of AddFirmware::validatePage()

        QString targetFileName = QFileInfo(context->athlete->home->root().canonicalPath() + "/../").canonicalPath() + "/" + fileName;

throws a segmentation fault

PR #1804, opened 2016-01-18
Fix for #1798 "Crash when "Copy to Library" selected in Tacx Fortius …

WaterworthD:
Fixes errors in AddFirmware::validatePage()

- context pointer was accessed but not intialised. Removed and accessed through parent->context
- Check to see if source and target file are the same file always returns true. Replaced fileName with filePath


liversedge:
Awesome, can't believe this has been there for so long!
Thanks.

PR #1810, opened 2016-01-24
Misc Fixes

Joern-R:
... DerivePower Prefences Dialog (2 Rows do not work in the Table View in Preferences)
... Add FR220 device - .fit


liversedge:
Well spotted.. been meaning to address that !

PR #1831, opened 2016-02-05
Add column to show keywords on LTM Popup

jonbev:
I hope you agree this is a useful addition. I find it hard to identify a ride without seeing the keywords but I guess everyone uses GC in different ways....


liversedge:
I think we need to find a more generic solution, so closing.

Sorry.

Issue #1835, opened 2016-02-08
Bugs in the WorkoutWizard


thomastaudt:
First of all, a huge thanks to all devs and supporters of GoldenCheetah. Having started power based indoor training just recently (using an Elite Qubo Digital Smart B+ trainer), I am delighted that I can use this great open source project. In fact, the existence of Golden Cheetah was a major reason why I decided on an ant+ and fec enabled trainer.

So I compiled a recent dev version of GoldenCheetah on my archlinux laptop, and besides some quirks with Qt (which are probably not GC's fault), I noted three bugs in the Workout Wizard:

1. The Gradient Creator shows inaccurate values of the distance in the metrics summary, since the rounding is done too early. So e.g. adding 4 trackpoints with 0.5 km each will still result in 0 km in the summary.

2. The saved .crs file from the Gradient Creator uses **accumulated distance** in the first column, but the .crs files I found on the web use the **distance to the next coursepoint** instead, which is also the import behavior of GoldenCheetah. So saving a gradient file with two points of 0.5 km will result in a course with 1.5 km instead of 1.0 km in the imported workout.
3. When manually saving the workout in the default workouts-directory, a warning ("... already exists ...") appears, because GoldenCheetah wants to copy it to itself when the file is imported in the workouts-database after saving. This at least feels buggy from a user perspective.

Points 1. and 2. are easily fixed, and my suggestion is to change `int totalDistance = 0;` to `double totalDistance = 0;` [here](https://github.com/GoldenCheetah/GoldenCheetah/blob/872f5d7f9fdda3ad7beadff657a1886e2dec90a2/src/WorkoutWizard.cpp#L575), and to simply remove the temporary `currentX` [here](https://github.com/GoldenCheetah/GoldenCheetah/blob/872f5d7f9fdda3ad7beadff657a1886e2dec90a2/src/WorkoutWizard.cpp#L617-L623) and replace it with `p.first` in the stream.

One way of resolving Point 3. is to check if `workoutDir` coincides with the directory of `source` before copying [here](https://github.com/GoldenCheetah/GoldenCheetah/blob/872f5d7f9fdda3ad7beadff657a1886e2dec90a2/src/Library.cpp#L204-L210). This change should probably be made for importing VideoSyncs, too ([here](https://github.com/GoldenCheetah/GoldenCheetah/blob/872f5d7f9fdda3ad7beadff657a1886e2dec90a2/src/Library.cpp#L175-L181)).

[...]

PR #1836, opened 2016-02-08
Resolve bugs in the WorkoutWizard (issue #1835)

tscode:
```
.. Corrected total distance for the metric summary in the Gradient
   Creator

.. Saving in .crs format now uses the right distance values (not
   cumulative distance)
```


thomastaudt:
Browsing e.g. [http://www.racermate.net/coursesviewer.php](http://www.racermate.net/coursesviewer.php) for `.crs` files gave me this impression, yes. And if it really was

1 100
2 200
3 100

then the ErgFile importing had to be changed -- right now, it would 
create a 6km workout instead of a 3 km one.

liversedge:
Sorry If I'm being awkward. Will check against the tests we already have for crs files tomorrow and merge. Have a good evening :+1:

thomastaudt:
Absolutely no need to apologize. Glad if I could help :grinning:

liversedge:
Apologies, this was missed in all other activity and is now stale so closing.

PR #1843, opened 2016-02-15
Give better errors if a file import fails due to basic file errors.

gille:
Before it would just say "not a valid file" even if the issue was a read bit missing, this is confusing.

Also a bit of whitespace cleanup.


liversedge:
Have applied the main fix manually in commit f3fbad5848a644b3ed738e3b01aa2254cd3468a4

PR #1847, opened 2016-02-16
Trainview intensity squash

erikboto:
I hope that this can be merged so it gets tested by a larger audience well before v4, so any quirks can be found and removed.


liversedge:
TrainIntensityAdjusment.h is missing (according to the Travis log) (check case of the header)

Also, I get the following on OSX, where it is not case sensitive....

iews.cpp:197:20: error: assigning to 'TrainBottom *' from incompatible type 'TrainIntensityAdjustment *'
    trainIntensity = new TrainIntensityAdjustment(trainTool, this);
                   ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Views.cpp:198:15: error: cannot initialize a parameter of type 'QWidget *' with an lvalue of type 'TrainBottom *'
    setBottom(trainIntensity);
              ^~~~~~~~~~~~~~
./TabView.h:61:33: note: passing argument to parameter 'bottom' here
        void setBottom(QWidget *bottom);
                                ^
Views.cpp:202:5: error: no matching member function for call to 'connect'
    connect(trainIntensity, SIGNAL(autoHideChanged(bool)), this, SLOT(onAutoHideChanged(bool)));
    ^~~~~~~

erikboto:
Oh, I had a few uncommited fixes. Sorry about that.

liversedge:
Thanks, just applying patch to play and test...

liversedge:
Jon likes it, so merging.
I might work on the cosmetics a little !

THANK YOU FOR THIS :+1:

PR #1853, opened 2016-02-19
WIP Qubo support

ihewitt:
This adds support for the "Elite Qubo Digital Wireless" ANT+ trainer, the trainer supports setting a target power level or a fixed resistance level. The trainer reports back a speed and cadence value. The cadence is either real from an additional optical sensor, or is estimated by the trainer (I was surprised how accurately this matched a real ANT cadence sensor).

This is very usable for me at the moment, but I had a question or two so thought a pull request would be a good place to track them.

There are a handful of minor behavioural changes in this branch.

1. I've added a new "manual trainer mode" this corresponds with a trainer that has a remotely settable fixed resistance level, this is connected up to both the new qubo level setting and the existing but not-wired-in fecSetResistance, although I haven't tested that since I don't have a FEC trainer.
*Q: This needs a new database entry in the workouts table, I wasn't sure of the correct method/protocol for adding the version check.*

2. There is a "power range" addition to the telemetry since the qubo target power is only valid for a set speed range, this is highlighted in the telemetry dial on the speed in the same manner as the target power. I thought this might be useful for other trainers.

3. There's a minor tweak to the up/down level buttons that matches the layout I used originally positioning the buttons above each other.

I plan on implementing a virtualpower calculation but haven't worked out how to make it degrade gracefully in "target power mode" or whether I can calculate/interpolate in that mode also.
Also I wasn't sure if this made more sense as just being provided as additional telemetry from the ANTMessage or implemented in the same manner as the other virtual power providers.

Cheers,
Ivor.


liversedge:
Is there any chance you can rebase this to make it a bit easier to read?

ihewitt:
Aha, done (albeit via trip to stackoverflow)
Still not loving git. :)

liversedge:
When you get there you realise that git is awesome.
Thanks for that. It means it will make for a cleaner merge too.

liversedge:
I'm a bit uncomfortable about this patch

a) is it really for FE-C resistance level not the QUBO
b) why are you deriving power, are there no ANT+ messages

It seems there is good things in here about fec resistance...

ihewitt:
Ignore that one! I'm working on another branch at the moment merged with some other changes. I'll resubmit and/or split it back out when I'm happier with it.

liversedge:
Hey -- did the other pull ever happen, apologies if I lost track of this .............

liversedge:
Just checking in again .. did you get any further ?

ihewitt:
Hey there, sorry for the delay! 
I had a bit of a break from my projects but have been hacking a few bits over the past few months and have been regularly dog-fooding the changes as my current virtual trainer. Using the qubo code and following maps, trying to get TT ready.
I *will* get some patches cleaned up and submitted.

xplong666:
Hi ihewitt!

I am interesting on this topic. 
Is there any news about it? 

Thanks in advance.

PR #1854, opened 2016-02-20
Performance improvements

ihewitt:
Some more simple performance improvements that make the startup a bit snappier.
1. Helps startup time a bit more, mainly by removing all of the QTextEdit controls used for html esacping and moving them into a utils namespace.
2. Stopped excessive looping in setYMax where axis height was small.

### DIFF
--- a/src/AllPlot.cpp
+++ b/src/AllPlot.cpp
@@ -33,6 +33,7 @@
 #include "Colors.h"
 #include "WPrime.h"
 #include "IndendPlotMarker.h"
+#include "Utils.h"
 
 #include <qwt_plot_curve.h>
 #include <qwt_plot_canvas.h>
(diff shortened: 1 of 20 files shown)


liversedge:
nice!

protect/unprotect are a dirty little secret. thanks for cleaning it.

jonbev:
It's nice to see that it is actually loading with the progress but I kind of feels like we're back to where we used to be with the progress bar on the loading screen.

Is there any way the loading can be multi threaded to speed it up? On my laptop it only uses one core.

liversedge:
the progress dialog is not part of this update. plus its loading the backing store from file and is only at startup and takes < 20 secs!

jonbev:
I'm afraid it take 1:10 for me. Thats running on a I5 laptop with SSD. Processing power definitely seems to be the bottleneck for me.

liversedge:
?

You are talking about little box that loads showing 1% - 100% as it goes?
How many rides do you have?

Mark

jonbev:
Yes the 1 - 100%

1058 activities.

liversedge:
If its a real concern I’d swich off interval discovery.

jonbev:
Does that run every time you startup? Surely not??

liversedge:
How much memory does your PC have available ?
You are using 64bit ?

liversedge:
Its loading the backing store RideDB.json.
Should take seconds.

Suspect there are either a gazillion ijtervals found or your PC has a memory issue.
Do you use the formal builds or your own?

Send me your RideDB.json and I’ll see why it might have an issue.

jonbev:
I have 8gig.

GC is running as 32bit. I built my own. I'll try the 64bit release you've just done.

jonbev:
Just loaded in 23secs with the 64bit version. 

I have sent my rideDB to see what you think.

PR #1856, opened 2016-02-22
Add detach and attach options to VideoWindow widget

ihewitt:
Minor hack, adds an option on the VideoWindow mini-menu allowing it to be detached and placed on a separate screen.
When detached the widget shrinks down,


liversedge:
I think we should think about a better way of doing this that works for all charts.

PR #1857, opened 2016-02-23
WIP Allow dynamic manual trainer levels

ihewitt:
Still a work in progress but wonder if the idea is worth doing. I did have this merged with the qubo support but thought it would be clearer in a separate patch.

In the process of adding the trainer level controls for the qubo, I wondered whether similar could be applied to the virtual power trainers, using the level controls to get the curve parameters from lookup tables.
This way, there's no need for multiple devices for each resistance level just the single device can be selected for all levels. The level can be altered manually on the fly and the resistance updated in the GoldenCheetah GUI.

Also, it's then possible to add the missing levels for trainers that have large numbers of them without having a huge list of devices.


ihewitt:
Closing this version, working on a merged branch, will resubmit if it's split out.

PR #1858, opened 2016-02-26
A few more low-hanging performance tweaks.

ihewitt:
Some more minor performance tweaks, mostly related to reducing intermediate string operations.

- isSwim and isRun are called a *colossal* number of times during file import, and each time are performing multiple string constructions and destructions and QMap lookups so only calculate this once since it won't change when a RideFile exists.
- fromUtf8 is now the default on Qt5.
- QTextEdit is very expensive to create, make once and re-use.

### DIFF
--- a/src/Core/RideDB.l
+++ b/src/Core/RideDB.l
@@ -40,25 +40,31 @@
 static QString unprotect(char *string)
 {
     // sending UTF-8 to FLEX demands symetric conversion back to QString
-    QString string2 = QString::fromUtf8(string);
-
+#if QT_VERSION < 0x050000
+    QString s = QString::fromUtf8(string);
+#else
+    QString s(string);
+#endif
     // this is a lexer string so it will be enclosed
     // in quotes. Lets strip those first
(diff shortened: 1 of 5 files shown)


gille:
I could be wrong, but I don't think the static bool isswim/isrun is correct. The static means that there's only one allocated so multiple files would overwrite it. Make it a local variable and cache it instead.

ihewitt:
Yup you're spot on - wasn't thinking straight. Thanks for noticing thast!

ihewitt:
I'll resubmit after I've looked at a couple of other things. Those other minor changes have a decent improvement on import speed though.

Importing a random mixture of 385 .FIT files:
```
time ./GoldenCheetah-orig
real    11m45.402s
user    8m13.752s
sys     0m5.644s
```
```
time ./GoldenCheetah-perf
real    5m42.875s
user    2m44.724s
sys     0m4.268s
```
(Optimised build)

liversedge:
I'd just add QTextEdit as a member of RideNavigator and keep using it, rather than creating one on every resetView

PR #1881, opened 2016-03-10
Add basic support for ANT-FE-C Stationary Bikes

erikboto:
This adds support for parsing the specific ANT+ page for stationary bikes, which includes power and cadence.


erikboto:
I don't see why the travis check failed, can you?

liversedge:
Looks like a travis error!

liversedge:
which bike did you test this with ?

erikboto:
It's kind of backwards. I tested this with my own implementation of FE-C for Monark bikes. They don't currently support FE-C themselves, so I did a small Monark to FE-C bridge. But then I realized that GC didn't parse the power and cadence data I sent, so I added that. The only "official" thing I've been able to try to verify the Monark-FE-C bridge with is SimulANT, since TrainerRoad and Zwift only supports FE-C trainers (quite sad considering the small amount of work needed to also support stationary bikes). 

I understand if you think this is too untested, but given that the code will only be used if that particular page is received I can't think of anything that it could break.

PR #1889, opened 2016-03-18
WorkoutWidget: increase x-axis length during workouts if needed

erikboto:
This should fix rescaling when running more than one hour in manual erg/slope mode.

I hope it's ok to use cadence.size() as a metric of how long the workout is, should always be increased even if cadence is not available, right?


liversedge:
Yes, this should work well. Apologies for the delay in applying it.

PR #1893, opened 2016-03-18
Add support for Kettler Ergo Bikes

erikboto:
A while back ago a guy asked on the mailing list about adding support for a protocol used by Kettler bikes. He found someone who had documented enough of the protocol to support it in GC, so I figured I might as well write the code for it.

It's tested on his "Kettler EX1" but should work on other Kettler ergo bikes as well.


liversedge:
Nice .. since QSerialPort is QT5 only you will need to fixup src.pro to only add the Kettler sources/headers the same way as for the Monark sources.

erikboto:
I think I got that right this time by using the same #ifdefs in headers and the conditional inclusion of source and headers in src.pro, the Qt4 build passed. The Qt5 build failed on qwt3d for some reason, have you seen that error before?

liversedge:
Brew has upgraded to qt5.6 and turned on c++11 and breaks lots of things.
Not sure what to do as, AFAIAC QT5.6 is unreasonable in deprecating WebKit when WebEngine has problems and forcing c++11 when it is not needed which breaks other libs.

Its a bit of a mess.

liversedge:
WRT src.pro that is my mistake, sorry. I misread the patch.

I will pull this in and disable qt5 builds till we support qt 5.6

gcoco:
I have set up the travis build so that it will use Qt 5.5.1 until we figure things out.

liversedge:
Damn thats clever, thanks Gareth.

PR #1896, opened 2016-03-22
LTMWindow: Fix missing include when building NOWEBKIT without GC_HAS_…

erikboto:
…CLOUD_DB

### DIFF
--- a/src/Charts/LTMWindow.cpp
+++ b/src/Charts/LTMWindow.cpp
@@ -38,6 +38,10 @@
 #include "GcUpgrade.h"
 #endif
 
+#ifdef NOWEBKIT
+#include <QWebEngineSettings>
+#endif
+
 #include <QtGui>
 #include <QString>
 #include <QDebug>


liversedge:
What platform are you trying this on?

I've found that Linux is not /too/ bad, but OSX is terrible.
Be interested to hear your feedback.

liversedge:
Also, that fix would be better placed in LTMWindow.h

erikboto:
It's on Windows, figured I might as well give it a try since you mentioned you were curious on the status on Windows. Anything in particular I should test?

liversedge:
Basically RideSummaryWindow uses QWebEngine and on OSX it flashes and renders badly and when in tiled mode the rendering goes berserk. Of particular note is moving the mouse in and out of a summary window causes it to go blank and vanish then return!

Linux seems ok though, so curious if Windows has the same problem

erikboto:
Hmm, I missed that there was a #ifdef for NOWEBKIT in the header. But isn't it generally better to keep the includes in cpp unless they are actually needed in the header, less including should keep build time down. But that might be optimizing too much?

liversedge:
Fair point.

erikboto:
I've been playing around with the ride summary in both tiled and tabbed mode for a couple of minutes, and I don't see anything weird (yet).

liversedge:
Oh that is good news.

Damien is looking at the getting Bing/Google maps and Ride view working too.

PR #1899, opened 2016-03-24
Use filename suffix to determine filetype if possible during export

erikboto:
I'd like your thoughts on this, if you think it's a good idea or just confusing since it's not valid for all cases since there's multiple csv options. I do a lot of indoor rides and always export to tcx for Garmin Connect, and often I miss to select the right format since I kind of expect that it will "do the right this".

Basically it checks if the chosen filename has a suffix that matches a "file writer", if so it will write using that format. If not it defaults to the old behavior of using the format selected in the drop-down list.


erikboto:
Any thoughts on this one? Do you think it makes sense?

liversedge:
seems sensible, especially on OSes where file extensions are often hidden from the user..

PR #1923, opened 2016-04-07
Fix corrupt timestamps on Windows builds

dresco:
The logic for building the timeval struct wasn't quite right.

@Joern-R - I don't have a Windows build environment to test this, however have verified the output matches the native getimeofday() calls on Linux.


dresco:
P.S. I only noticed this from some antlogs that had been sent through from Windows dev build users..

liversedge:
The CI Windows build on appveyor succeeded so assume this will build ok.

PR #1925, opened 2016-04-12
Fixed AvgSpeed to fallback to Distance/Duration when there are no sam…

amtriathlon:
…ples

This behavior was changed after compute() refactor.

The problem was reported on https://groups.google.com/d/msg/golden-cheetah-users/Uplh6fCGBAI/6hYNrPBSAwAJ, I think the condition can be safely removed to restore the previous behavior.

### DIFF
--- a/src/Metrics/BasicRideMetrics.cpp
+++ b/src/Metrics/BasicRideMetrics.cpp
@@ -880,13 +880,6 @@ class AvgSpeed : public RideMetric {
 
     void compute(RideItem *item, Specification spec, const QHash<QString,RideMetric*> &deps) {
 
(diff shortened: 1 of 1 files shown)


liversedge:
Thanks Ale, not sure what I did there.
Will push once the CI finishes (it takes an age)

amtriathlon:
No problem Mark, it is a minor detail for such a refactor!

BTW, a bigger problems seems to occur when you override a metric the overridden value is applied for each interval, for example if you override Duration or Distance all intervals report that Duration or Distance, I found this on manual workouts with laps but it is a general issue:

image

Issue #479, opened 2013-02-09
Add BTLE support


liversedge:
Already working on it in the code, but need to record the feature here, so we can track progress and add to the release notes when done.

Initially this will be just for Mac via the WF API.

Linux/Windows BTLE support is a whole new ball of wax.

viiru-:
I have written most of the glue code to allow Goldencheetah to use BTLE sensors through the Qt builtin BTLE support. Proper integration with the add device wizard is still missing.

Currently I have this built as a new and completely separate controller. I later noticed that the current wfapi-based bt40controller is actually disabled, and incomplete. Are there still plans to finish that work, or would it be preferrable to you if I structured the Qt BTLE based version as a replacement for that one?

liversedge:
Hi,

The Wahoo API should be removed from the code completely. We will never support it directly since it violates the GPL.

Please feel free to add support -- I can help, and also have a BT dongle and BT devices to test as well.

liversedge:
If you are willing to pursue this we can add it to the v4 queue -- do let me know.

viiru-:
Ok, I'll start working on a clean branch that replaces the bt40controller with the new stuff. I'm willing to pursue this, but I can't guarantee much about the timetable. I'll get back to you with a branch when I have one.

liversedge:
ACE! Will put it into v4 queue.

CHEERS
Mark

liversedge:
Fixed in https://github.com/GoldenCheetah/GoldenCheetah/pull/1936

PR #1936, opened 2016-04-25
Bt40 conversion

viiru-:
How about this for issue #479 ?


liversedge:
That was quick!

Will pull and it'll be in the next dev build for folks to try.

Nice one.

PR #1942, opened 2016-04-30
R-Windows - Includes

Joern-R:
.. don't use R command function to find the includes. but R_HOME


liversedge:
will pull in, but should also add include/x64 so will add separately.
not sure if it makes any real difference.

Joern-R:
My installation (from R project) does not show a include/x64 - that why I
did not add it.
It shows a subfolder include/R_ext that's all.

2016-04-30 17:24 GMT+02:00 Mark Liversedge <notifications@github.com>:

liversedge:
Ah ok. Then I won't add one !

Thanks.

PR #1959, opened 2016-05-11
Fallback to Start with Cycling zones for Running

### LINKED ISSUE #1958: [bug] After "Merge pull request #1956" TRIMP for Run activites always =0 on trend
In Activity summary (detail for 1 activity) TRIMP is OK, but on progress trend all TRIMP values always = 0 for all "Run" events.

amtriathlon:
When there is no Running zones defined to easy
transition from current config.
Fixes #1958


liversedge:
Thanks Ale !

PR #1961, opened 2016-05-11
Enum for table headings

mtbkeith:
@liversedge 

Can you review? I've changed all the indexes to use an enum. I'll have some time to embark on multi-threading this again soon.


liversedge:
Gonna close as so old -- could combine with multithreading of import?

PR #1964, opened 2016-05-12
Minor document updates for MAC

mtbkeith:
URL and version updates; Nokia has transferred ownership of Qt to Digia back in 2012.


mtbkeith:
ping @liversedge wondering about the open PRs? thanks

PR #1969, opened 2016-05-18
Align interval bars plot with main plot

antoniusriha:
This commit fixes the issue that the interval bars (in AllPlotInterval) are not correctly aligned
with the interval sections in the main plot (AllPlot). This is done by calculating the margins from
the main plotting area to the border and setting left and right margin of the interval plot
accordingly.


liversedge:
thanks will merge and test !!

PR #1977, opened 2016-05-27
Save rim and tire size settings

antoniusriha:
This PR adds rim and tire size to the athlete's settings values (in addition to wheel size). UI elements for those settings have already been present, but the values have not been persisted so far.

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -224,6 +224,8 @@
 #define GC_LTS_DAYS                             "<athlete-preferences>LTSdays"
 #define GC_STS_DAYS                             "<athlete-preferences>STSdays"
 #define GC_CRANKLENGTH                  "<athlete-preferences>crankLength"
+#define GC_RIMSIZEIDX                   "<athlete-preferences>rimsizeidx"
+#define GC_TIRESIZEIDX                  "<athlete-preferences>tiresizeidx"
 #define GC_WHEELSIZE                    "<athlete-preferences>wheelsize"
 #define GC_USE_CP_FOR_FTP               "<athlete-preferences>cp/useforftp"                       // use CP for FTP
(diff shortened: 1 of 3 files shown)


liversedge:
Hi,

Thanks for the PR .. why are you adding these?
Usually we add settings because we need to use them .. how will these settings be used?

THANKS!

Mark

antoniusriha:
There is no other use than to save user input for rim and tire size, so the
combo box values can be set when loading the settings dialog. Otherwise
users might think the program is not working properly, if the inputted
values disappear between program runs.

liversedge:
Sorry, I don't understand. What is the purpose of adding these?

antoniusriha:
They are just there to set the resp combo box values in the athlete's tab
in the options dialog. No other use, just a usability improvement.

liversedge:
The problem here is they are redundant, so no reason to add them and the maintenance overhead associated with them. There are several places where config is queried; REST api, datafilters, R api and they will all need to be updated.

There must be a reason you wanted to add them there?

antoniusriha:
From a business logic standpoint they're redundant, because only the wheel
size is needed for calculations. From a usability standpoint I believe them
to be valuable, because it feels odd, if you open a dialog and values
you've inputted earlier are missing. Although the use of these two fields
in the UI is solely to calculate the wheel size, they don't convey the look
& feel of a "calculator", but rather of setting fields.
Also, I didn't know about the other places that would need to be changed
(REST api, R api, ...).
All in all I think the usability gain needs to be weighed against the
increased maintenance effort necessary, to decide if this PR should go in.
Alternatively, one could rebuild that part of the UI to convey a more
calculator type look & feel, e.g. by adding a button "Calculate" and
putting the fields in a separate dialog. This dialog could be reused in
other parts, e.g. during training device setup.

PR #1978, opened 2016-05-27
Fix loading of binWidth setting in 3d plots

antoniusriha:


src/Charts/ModelWindow.h
```diff
@@ -99,7 +99,7 @@ class ModelWindow : public GcChartWindow
         bool isLegend() const { return legend->isChecked(); }
         void setLegend(bool x) { legend->setChecked(x); }
         QString bin() const { return binWidthLineEdit->text(); }
-        void setBin(QString x) { binWidthLineEdit->setText(x); }
+        void setBin(QString x) { binWidthLineEdit->setText(x); setBinWidthFromLineEdit(); }
```
liversedge: setBinWidgthFromLineEdit gets called when binWidthLineEdit changes anyway ?

antoniusriha: The bug that I'm trying to fix, is that when starting up GC and opening my
custom 3d plot, all values except the bin width are set correctly. This
seemed to be the least invasive fix.

liversedge: ah, ok. you're probably right. its a pickle, so this is the simplest way of resolving that. thanks!

PR #1980, opened 2016-05-28
CloudDB - add .gchart Support

Joern-R:
... add export support for the new .gchart format (allowing to post all chart types to CloudDB)
... add import in main menu for the new .gchart format
... allow to import multiple charts at once

... remove specific LTM Chart support (menu, db,... -since this is covered by general .gchart format)


liversedge:
SUPER COOL!

PR #1984, opened 2016-05-31
Reverse color range in 3D plot to run from blue (low) to red (high)

antoniusriha:
This is a proposal to reverse the color range in 3D plots to run from blue (low vals) to red (high vals), which seems to be the more natural way of displaying the data.
Also, I've added some checks to prevent division by 0 and out of range exceptions.


liversedge:
Wow, good to know someone is using this plot ! :)

antoniusriha:
I like the 3D plots a lot! I've a plot visualizing my ride profile in 3D with color indicating the gradient

PR #1985, opened 2016-05-31
Fix loading of 3D chart settings

antoniusriha:
There were still some bugs with loading the settings of 3D charts. Details are in the commit messages.

### DIFF
--- a/src/Charts/ModelWindow.cpp
+++ b/src/Charts/ModelWindow.cpp
@@ -156,7 +156,6 @@ ModelWindow::ModelWindow(Context *context) :
     // now connect up the widgets
     connect(this, SIGNAL(rideItemChanged(RideItem*)), this, SLOT(rideSelected()));
     connect(context, SIGNAL(intervalSelected()), this, SLOT(intervalSelected()));
-    connect(presetValues, SIGNAL(currentIndexChanged(int)), this, SLOT(applyPreset(int)));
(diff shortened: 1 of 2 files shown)


src/Charts/ModelWindow.cpp
```diff
@@ -321,18 +322,19 @@ static t_preset *presets;
...
-        zSelector->setCurrentIndex(presets[index].z);
-        colorSelector->setCurrentIndex(presets[index].color);
-        ignore->setChecked(presets[index].ignore);
-        binWidthSlider->setValue(presets[index].bin);
-        binWidthLineEdit->setText(QString("%1").arg(presets[index].bin));
-
-        setDirty();
-        if (index) setData(true);
-    }
+    if (index < 0 || !context->tab->chartsettings()->isVisible())
```
liversedge: this is not a good approach; you are intertwining gui state with configuration; they shouldn't be related this way.

antoniusriha: Guess you're right. But I don't know how else to fix it. What I'd like to do, is: I only want applyPreset to run, if the user selected a preset and not if the preset was set by the QT property loading engine. The isVisible flag seemed to tell best, whether it was the user or something else.

antoniusriha:
I've now employed another approach: The `currentIndexChanged` signal is only connected to `applyPreset` after the `preset` property has been loaded. I guess that is not the cleanest solution either, but better as the first one?

liversedge:
thats probably the best way of fixing it

PR #1986, opened 2016-05-31
Allow min value of 1 for binWidth in 3D plot

antoniusriha:
This enables creating high resolution 3D profiles:

profile3d


liversedge:
I seem to recall there was a reason I set it to 3 but can't for the life of me remember why. so lets merge this in, but it might break something. given you use the chart, I guess you will find it !

PR #1990, opened 2016-06-05
CloudDB - only one Nickname in GC

Joern-R:
... re-use existing Nickname from Athlete preferences


liversedge:
The GC nickname has never been used for anything so its kind of neat we _finally_ do something with it.
I think I introduced it in 2010 .. so we got there in the end!

:+1:

PR #2006, opened 2016-06-26
Added bestime(distance) to formulas

amtriathlon:
To have an easy way to track best times by distances
in Metrics Trends charts and User Defined Metrics
s.t. 40k TTs, 10k runs, 1500m swims, etc.

### DIFF
--- a/src/Core/DataFilter.cpp
+++ b/src/Core/DataFilter.cpp
@@ -24,6 +24,7 @@
 #include "RideFileCache.h"
 #include "PMCData.h"
 #include "VDOTCalculator.h"
+#include "AddIntervalDialog.h"
 #include <QDebug>
 
 #include "Zones.h"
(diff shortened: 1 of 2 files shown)


src/Core/DataFilter.cpp
```diff
@@ -2198,6 +2200,18 @@ Result Leaf::eval(DataFilterRuntime *df, Leaf *leaf, float x, RideItem *m, RideF
...
                 break;
 
+        case 36 :
+                {   // BESTTIME (distance[km])
+
+                    if (leaf->fparms.count() != 1) return Result(0);
+
+                    QList<AddIntervalDialog::AddedInterval> results;
+                    AddIntervalDialog::findPeaks(m->context, false, m->ride(), RideFile::kph, RideFile::original, 1000*eval(df, leaf->fparms[0], x, m, p, c).number, 1, results, "","");
+
```
liversedge: Isn't this /really/ slow ?

It will open all the rides in the date range (!)

liversedge:
we could add mean maximal by distance not time to the ride cache for speed only ?

amtriathlon:
I think that would be enough since it would allow to recover time for any
distance as distance/speed

PR #2008, opened 2016-07-10
Fix XDATA Model Compile errors on Windows

Joern-R:
... please check if True/False's are correct

### DIFF
--- a/src/FileIO/XDataTableModel.cpp
+++ b/src/FileIO/XDataTableModel.cpp
@@ -149,12 +149,14 @@ XDataTableModel::setHeaderData(int section, Qt::Orientation , const QVariant & v
         //headings_[section] = value.toString();
         //return true;
     //}
+    return false;
 }
 
 bool
 XDataTableModel::insertRow(int row, const QModelIndex &parent)
 {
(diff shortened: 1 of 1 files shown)


src/FileIO/XDataTableModel.cpp
```diff
@@ -149,12 +149,14 @@ XDataTableModel::setHeaderData(int section, Qt::Orientation , const QVariant & v
...
 }
 
 bool
 XDataTableModel::insertRow(int row, const QModelIndex &parent)
 {
     //return insertRows(row, 1, parent);
+    return false;
 }
 
 bool
```
liversedge: They're not really that important at the moment since these functions aren't used yet.
I will fix up all the warnings once I've added delete/insert rows etc.

I'm working on refining the UX at the moment (e.g. right now looking at using a close button on the tabs to remove an XDATA series).

Joern-R:
On Windows it broke the build. But not a problem to wait until all is fixed.
So just reject the PR.

liversedge:
Oh, then just apply it for now. I can fixup later.

PR #2009, opened 2016-07-11
Best time for standard distances

amtriathlon:
Added Best time metrics for standard distances in swimming, running, rowing and cycling.
FindPeaks, slightly generalized adding an Specification parameter, replaces both findBests and findBestsKPH and allows to deprecate BestIntervalDialog to have less code duplication.


liversedge:
Seems like a good compromise, rather than adding as Meanmaximal series in RideFileCache.

amtriathlon:
Hi Mark, If you want to merge this I just resolved the merge conflict created by the last changes to RideMetric.cpp

Issue #1999, opened 2016-06-15
Support for Kettler devices via bluetooth 


kaegi:
My `RUN 7` doesn't seem to be supported by Golden Cheetah, so I wrote a library that should make it very easy:

https://github.com/ChangSpivey/kdri

It should also provide support for other models, though they aren't tested. Since GC is the main purpose of `kdri`, I'm still open to breaking API changes (in case this is needed).

liversedge:
Cool stuff !

Was there a reason you chose to use rust ?
How have you integrated this lib with GC, do you have a patch for that ?

Not sure how we could make this work as it stands in a cross platform way. 

Since there is already support for Bluetooth devices in the GC codebase using QT's BT support I think it would be better to support it natively ...

ChangSpivey:
My choice for Rust: This is supposed to be a low-level library, so I limited myself to C, C++, Rust (or D). Since Rust makes memory management (among other things) so much easier I chose this language.

The library uses the `SDP` and `RFCOMM` interface of BlueZ. As far I as I know (https://stackoverflow.com/questions/6992333/does-qt-support-the-windows-bluetooth-api), QT does not support `RFCOMM` on windows. Is there any benefit of QT's BT?

liversedge:
Yes Windows BT support is missing, which is a shame. We chose it since we're a QT app and (hopefully) they will support Windows in later releases.

Am I right in thinking you have not integrated this with GC?

Thanks

Mark

ChangSpivey:
I'm currently restructuring `bluetooth-rs`, and because the bluetooth API on Windows is similar to BlueZ, I might implement support for it. macOS is more difficult, because it does not seem to use the POSIX socket API, but something more higher level.

The only cross platform BT libraries I found are written in Python (LightBlue) and Node.js (bluetooth-serial-port), which I don't like for `kdri`.

I don't have a patch right now but implementing a `KettlerBluetoothController.cpp` seems to be easy enough. Will the patch be accepted even if there is currently only Linux support?

liversedge:
Sure we can accept the patch as an optional dependency, it may not be put into official builds.
But if you have a controller we might look to adapt it to the QT BT API, so it would be gratefully received.

ChangSpivey:
[...]

PR #2025, opened 2016-07-28
Add support for serial Bluetooth Kettler devices

kaegi:
#1999

### DIFF
--- a/contrib/kdri-build.sh
+++ b/contrib/kdri-build.sh
@@ -1,0 +1,5 @@
+# rebuild
+rm -Rf kdri-c-wrapper
+git clone https://github.com/ChangSpivey/kdri-c-wrapper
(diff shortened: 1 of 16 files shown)


kirklentz:
I bit off topic, but I was wondering if anyone can comment on the following:
1) I have a Kurt Kinetic Road Machine with Inride sensor.  Using Windows 10 with bluetooth.  I also have the BLED 112 type bluetooth dongle (the one that is required to use TrainerRoad).  Is there anyway with this equipment that I can get GC ver 3.4 to pick up data from the Inride sensor either via bluetooth directly or via the BLED 112 dongle?
2) If not, are there any plans in the V4 build to provide support for Windows 10 and bluetooth devices - either directly or thru the BLED 112 dongle?

PR #2048, opened 2016-08-10
Add Pioneer SGX-CA500 FIT support

chocoby:
Support Pioneer [SGX-CA500](http://pioneer-cyclesports.com/us-en/products/cyclocomputer/sgx-ca500.html) FIT file 👍


chocoby:
Thanks! :smile:

Issue #2050, opened 2016-08-12
Moxy minimum Sm02 and Thb values


ryinc:
When the minimum and maximum Sm02 and Thb metrics are added to "Summary" and "Interval", Golden Cheetah is not working correctly.

(1) Minimum values displayed are actually maximum values rather than minimum values
(2) The maximum Smo2 and THb values are not actually added to the "metrics" summary despite being selected in the Options>Metrics>Summary

gernot-hochmuth:
with reference to (1): I noticed the incorrect calculation of the minimum values of **SmO2** and **tHb** as well. These errors can easily be corrected in file *src/Metrics/BasicRideMetrics.cpp*. Mind that **tHb** - i.e. the absolute concentration of hemoglobin in the blood in the specific muscle - should not become zero, as this would imply that the muscle were totally drained from blood. Following that logic the average of **tHb** should be calculated excluding zero values as well (as it is done with heart rate). 
I am not quite sure whether the minimum value of **SmO2** might become zero, as that would imply that the examined muscle had consumed all available oxygen. May someone with a deeper knowledge in physiology comment on that.

PR #2052, opened 2016-08-12
Fix calculation of minimum values of SmO2 and tHb and average of tHB

gernot-hochmuth:
In file *src/Metrics/BasicRideMetrics.cpp*  the minimum values of **SmO2** and **tHb** and the average of **tHb** are calculated incorrectly. In functions *compute* of class *MinSmO2* resp. class *MintHb* actually the maximum values get computed. 
For *MinSmO2* the comparison in function *compute* should read:

```c++
    if (point->smo2 >= 0.0f && (notset || point->smo2 < min)) {
        min = point->smo2;
        if (point->smo2 > 0.0f && notset)
            notset = false; 
    }
```

with *min* being a double initialized to 0.0f and  *notset* a boolean initialized to *true*.
For *MinSmO2* we've got to handle that the first recorded values might be zero, e.g. when the athlete did not attach the Moxy sensor to the muscle prior to starting recording. If the comparison given above holds true we have to verify that we get a valid value > 0 first before allowing SmO2 to fall down to zero again.

For *MintHb* the comparison in function *compute*  should read:

```c++
    if (point->thb > 0.0f && (notset || point->thb < min)) {
        min = point->thb;
        notset = false;
     }
```

with *min* being a double initialized to 0.0f and  *notset* a boolean initialized to *true*.
Please note that for physiological reasons the minimum value of **tHb** must not be zero as this would imply that the muscle under observation were totally drained from blood.

At the end of functions *compute* in classes *MinSmO2* and *MintHb* we have to set the correct value:

```c++
    setValue(min); // was max before
```
In class *AvgtHb* we have to take into account that **tHb** must not be zero as well for the same reason as stated above. In function *compute* the comparison should read:

```c++
    if (point->thb > 0.0f) { 
        total += point->thb;
        ++count;
    }
```

References #2050 
please review

### DIFF
--- a/src/FileIO/CsvRideFile.cpp
+++ b/src/FileIO/CsvRideFile.cpp
@@ -181,6 +181,7 @@ RideFile *CsvFileReader::openRideFile(QFile &file, QStringList &errors, QList<Ri
     int cadenceIndex=-1;
     int smo2Index=-1;
     int hrIndex=-1;
+    int kphIndex = -1;
     int gctIndex = -1, voIndex =-1;
 
     double precAvg=0.0;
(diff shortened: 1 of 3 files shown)


liversedge:
Sorry about the delay in merging, I completely missed the request.

Thanks for the fixes !!

PR #2055, opened 2016-08-17
Qwkcode fixes

erikboto:
Addresses the issues discussed on golden-cheetah-users: Fixes the inconsistency with lap markers, and the problem with compressing efforts below CP into Nx blocks in qwkcode.


liversedge:
Thanks!

PR #2114, opened 2016-10-01
Show only metrics relevant for all compared intervals

amtriathlon:
In Compare Pane and Ride Summary


amtriathlon:
I'm not happy with the RideSummary implementation and probably we should show the metrics which are relevant for any compared interval. I will retry this later.

Issue #2077, opened 2016-09-05
Broken numerical sort order.


jwm-art-net:
When ordering activities, or ordering efforts on a segment in the compare pane, numerical values are not sorted in the way expected. The problem is not apparent where the field contains integers, only where the field is a float value or a time value.

On slower activities where average speed is below 10, sorting activities by average speed results in a list like this:

10.0 ... 11.3 ... 18.4, 3.2 ... 7.4 ... 9.9

The bug is reproducable in the compare pane for example by sorting segment efforts by duration:
1:00 ... 1:48 ... 2:15, 34 ... 46 ... 58

GoldenCheetah V3.4 DEV1608 (3945)
Arch Linux

Joern-R:
Fixed via https://github.com/GoldenCheetah/GoldenCheetah/pull/2117

jwm-art-net:
This doesn't appear to be fixed in GoldenCheetah V3.5 RC2X (3982)
![20200529-194233](https://user-images.githubusercontent.com/308448/83294261-f1d36580-a1e4-11ea-8b5c-355adb854173.png)

PR #2117, opened 2016-10-02
Improve sorting in

Joern-R:
... compare pane for time/duration fields
... ride navigator for numeric fields e.g. average speed (which were sorted as string before)

see Github Issue: https://github.com/GoldenCheetah/GoldenCheetah/issues/2077


Joern-R:
QT 5 build did not succeed - but no build error but sort of time-out.
I don't see an option to restart Travis to check if QT5 passes (Win passes).

liversedge:
I've restarted it

PR #2131, opened 2016-10-08
Skip non relevant metrics for RideSummary in Trends view

amtriathlon:
Adds isMetricRelevantForRides(specification, metric) to RideCache
to check if a metric isRelevant for some of the activities passing
the specification
Also reduces from 3 to 1 the calls to getRideTypeCounts

A similar criteria could be applied to the columns of rides/runs/swims lists to filter columns not relevant for any of the activities in the list.


amtriathlon:
There is an obvious optimization missing for early exit from the or loop, I will add it tomorrow

amtriathlon:
I am finished with this for now, tested an athlete with +3000 swims/rides/runs and it seems to be well behaved.

PR #2137, opened 2016-10-14
Add files via upload

UgoBor:
Added updated Italian translation


amtriathlon:
The translation file was uploaded instead of changed, I will commit it directly.

PR #2152, opened 2016-10-25
Close and free log file object after duplicating it

antoniusriha:


liversedge:
Not sure this is neccessary at all.

antoniusriha:
It is (by definition) a leak. Although a minor one. I think it's a question of attitude towards leaks. There is a question on stackoverflow [1] about it. I like the accepted answer.

[1] http://stackoverflow.com/questions/273209/are-memory-leaks-ever-ok

liversedge:
dup() overwrites the file descriptor.

antoniusriha:
I've just looked up the dup() documentation [1]: It duplicates the file descriptor. This means proper handling would be to immediately close and free the QFile object. I'll try it out and update the PR accordingly.

[1] http://codewiki.wikidot.com/c:system-calls:dup

PR #2163, opened 2016-10-30
Don't display unsupported charts MapWindow and StreetViewWindow in menu

antoniusriha:


liversedge:
You have misunderstood what nowebkit means.

antoniusriha:
My understanding is that NOWEBKIT indicates that QWebView and other WebKit classes are not available. Since the classes MapWindow and StreetViewWindow depend on those classes, it doesn't make any sense to display the respective entry in the menu for them. Currently with NOWEBKIT defined empty GcChartWindow objects are displayed.
Could you please explain what my misconception of this is?

liversedge:
It means QWebKit is not to be used, QWebEngine is used instead.
This is early support for QWebKit deprecation.

antoniusriha:
Ok, thanks for the clarification!

On Oct 31, 2016 9:03 AM, "Mark Liversedge" <notifications@github.com> wrote:

It means QWebKit is not to be used, QWebEngine is used instead.
This is early support for QWebKit deprecation.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/GoldenCheetah/GoldenCheetah/pull/2163#issuecomment-257234073>,
or mute the thread

PR #2165, opened 2016-10-31
Handle "Abort" and "Finish" correctly in RideImportWizard

flyingflo:
I wonder if anybody else has this problem. The RideImportWizard 
dialog stays open forever. 

The current state was encoded in the text label of the abort button,
but the comparison operation did never match, so only the default state
was ever handled. This prevents aborts and causes the window to never
close.

Add a variable to store the current state, instead.

### DIFF
--- a/src/Gui/RideImportWizard.cpp
+++ b/src/Gui/RideImportWizard.cpp
@@ -53,7 +53,7 @@ RideImportWizard::RideImportWizard(QList<QUrl> *urls, Context *context, QWidget 
     init(filenames, context);
     filenames.clear();
     _importInProcess = false;
-
+    _importState = READY;
 }
 
 RideImportWizard::RideImportWizard(QList<QString> files, Context *context, QWidget *parent) : QDialog(parent), context(context)
@@ -64,7 +64,7 @@ RideImportWizard::RideImportWizard(QList<QString> files, Context *context, QWidg
     autoImportStealth = false;
     init(files, context);
     _importInProcess = false;
-
+    _importState = READY;
 }
 
 
@@ -73,6 +73,7 @@ RideImportWizard::RideImportWizard(RideAutoImportConfig *dirs, Context *context,
     _importInProcess = true;
     autoImportMode = true;
     autoImportStealth = true;
+    _importState = READY;
 
     if (autoImportStealth) hide();
     QList<QString> files;
@@ -896,14 +897,14 @@ RideImportWizard::abortClicked()
     QString label = abortButton->text();
 
     // Process "ABORT"
-    if (label == tr("Abort")) {
+    if (_importState == RUNNING) {
         hide();
         aborted=true; // terminated. I'll be back.
         return;
     }
 
     // Process "FINISH"
-    if (label == tr("Finish")) {
+    if (_importState == DONE) {
        // phew. our work is done. -- lets force an update stats...
        hide();
        if (autoImportStealth) {
@@ -918,6 +919,7 @@ RideImportWizard::abortClicked()
 
     // SAVE STEP 2 - set the labels and make the text un-editable
     phaseLabel->setText(tr("Step 4 of 4: Save to Library"));
+    _importState = RUNNING;
 
     abortButton->setText(tr("Abort"));
     aborted = false;
@@ -1073,6 +1075,7 @@ RideImportWizard::abortClicked()
     progressBar->setValue(progressBar->maximum());
     phaseLabel->setText(donemessage);
     abortButton->setText(tr("Finish"));
+    _importState = DONE;
     aborted = false;
     if (autoImportStealth) {
         abortClicked();  // simulate pressing the "Finish" button - even if the window got visible
--- a/src/Gui/RideImportWizard.h
+++ b/src/Gui/RideImportWizard.h
@@ -91,6 +91,7 @@ private:
 
     QStringList deleteMe; // list of temp files created during import
 
+    enum { READY, RUNNING, DONE } _importState;
 
 };
 


flyingflo:
My setup: Ubuntu 16.04.1 building with QT5

liversedge:
This doesn't make sense to me, are you running in English or other language?

flyingflo:
Running in English. QT seems to encode [shortcut keys in the text.](http://doc.qt.io/qt-5/qabstractbutton.html#text-prop)

The text() QString is "&Finish" and the QString returned by tr() is "Finish".

Weird that it happens only for me.

amtriathlon:
2016-11-01 5:26 GMT-03:00 flyingflo <notifications@github.com>:

> Running in English. QT seems to encode shortcut keys in the text.
> <http://doc.qt.io/qt-5/qabstractbutton.html#text-prop>
>
> The text() QString is "&Finish" and the QString returned by tr() is
> "Finish".
>
> Weird that it happens only for me.
>
But abortButton doesn't have accelerators at creation nor in the subsequent
setText...
What is the origin of the &?
I also have Ubuntu 16.04.1 building with QT5.5.1 and this is working
without problems.

flyingflo:
Ok, weird. Close the PR, if nobody else is affected.

We might use slightly different setups or default settings that could affect the recycled button text. Just guessing. Anyways, the enum state variable is very unlikely to read something that was not written :wink:

liversedge:
which exact version of QT are you building with?
we have seen issues like this before, but they were bugs in QT...

liversedge:
qmake -query will list all version info

flyingflo:
Found something: Looks like KDE causes this. It automatically inserts accelerators.

http://stackoverflow.com/questions/32688153/how-to-disable-automatic-mnemonics-in-a-qt-application-on-kde
https://bugs.kde.org/show_bug.cgi?id=337491
https://bugs.kde.org/show_bug.cgi?id=345023

flyingflo:
```
% qmake -query
QT_SYSROOT:
QT_INSTALL_PREFIX:/usr
QT_INSTALL_ARCHDATA:/usr/lib/x86_64-linux-gnu/qt5
QT_INSTALL_DATA:/usr/share/qt5
QT_INSTALL_DOCS:/usr/share/qt5/doc
QT_INSTALL_HEADERS:/usr/include/x86_64-linux-gnu/qt5
QT_INSTALL_LIBS:/usr/lib/x86_64-linux-gnu
QT_INSTALL_LIBEXECS:/usr/lib/x86_64-linux-gnu/qt5/libexec
QT_INSTALL_BINS:/usr/lib/x86_64-linux-gnu/qt5/bin
QT_INSTALL_TESTS:/usr/tests
QT_INSTALL_PLUGINS:/usr/lib/x86_64-linux-gnu/qt5/plugins
QT_INSTALL_IMPORTS:/usr/lib/x86_64-linux-gnu/qt5/imports
QT_INSTALL_QML:/usr/lib/x86_64-linux-gnu/qt5/qml
QT_INSTALL_TRANSLATIONS:/usr/share/qt5/translations
QT_INSTALL_CONFIGURATION:/etc/xdg
QT_INSTALL_EXAMPLES:/usr/lib/x86_64-linux-gnu/qt5/examples
QT_INSTALL_DEMOS:/usr/lib/x86_64-linux-gnu/qt5/examples
QT_HOST_PREFIX:/usr
QT_HOST_DATA:/usr/lib/x86_64-linux-gnu/qt5
QT_HOST_BINS:/usr/lib/x86_64-linux-gnu/qt5/bin
QT_HOST_LIBS:/usr/lib/x86_64-linux-gnu
QMAKE_SPEC:linux-g++-64
QMAKE_XSPEC:linux-g++-64
QMAKE_VERSION:3.0
QT_VERSION:5.5.1
```

flyingflo:
I'm not sure what exactly happens, but it suggests it's related to KDE. The linked bug reports describe issues with the qt designer. When KDEPlatformTheme.so is in the process, it injects the accelerators.

I checked GoldenCheetah's /proc/XX/maps, it also has KDEPlatformTheme.so.

Anybody with KDE seeing this?

liversedge:
OK, that makes sense.

One simple fix would be to add the accelerators to the texts we use, so instead of "Finish" we set the text to "&Finish". 

It would also make sense to abstract these texts out;

static const QString FINISH = tr("&Finish");

if (abortButton->text() == FINISH) ...

But this would mean we need to retranslate all the texts .. so it might be better to do:

static const QString FINISH;

FINISH = QString("&%1").arg(tr("Finish"))

Then we just hope the translated texts don't start with the same character !

OR

We use an enum to avoid being tied down to the texts !!!

SO

Perhaps your solution is best after all ?

:)

flyingflo:
From a style point-of-view, I would prefer not to encode state information in the button text, especially if it is not reliable, because other things are encoded in the text as well. 

I updated the commit comment to reflect our findings about this behaviour.

flyingflo:
I guess it's not a bug for most of you. I'll maintain a KDE-compatible branch for myself.

flyingflo:
btw, I found this problem in some more dialogs in GC.

Furthermore, I found a simple workaround.
Add
```
[Development]
AutoCheckAccelerators=false
```
to ~/.config/kdeglobals to stop KDE from injecting the accelerators, and thus prevent changes to button text.

liversedge:
Can you add that to the wiki?

flyingflo:
Wrote the workaround in the wiki and added it to the FAQ (very bottom).

liversedge:
THANKS!

PR #2178, opened 2016-11-10
Fix shadowbuild

antoniusriha:
This PR makes it possible to have clean shadow builds that don't have shared build output, which before were: libqwt, libkqoauth and the translation file lrelease compilation outputs. Shared build output lead to linker errors, when switching from one shadow build (e.g. with Qt 5.4) to another shadow build (e.g. with Qt 5.7).

### DIFF
--- a/kqoauth/kqoauth.pro
+++ b/kqoauth/kqoauth.pro
@@ -1,5 +1,5 @@
 TARGET = kqoauth
-DESTDIR = .
+DESTDIR = ./lib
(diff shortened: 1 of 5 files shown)


liversedge:
I'm going to close this, but fix the kqoauth issue separately -- it keeps biting me on MacOS, so I feel the pain, just not wanting to mess about with translation files.

liversedge:
I appreciate the pr tho 

Mark

PR #2187, opened 2016-11-15
Function to control trainers with seperate powermeter

ragman1976:


liversedge:
What is this pull request?

It looks like you need to rebase this.

ragman1976:
Many trainers have an inaccurate power measurement. For somebody who has a accurate Powermeter on the bike it would be great to control the trainer with the Powermeter from the bike.
I guess PerfPro contain such a functionality for example.

liversedge:
Your pull request is into a maintenance branch - it should be master
Can you resubmit for master and make sure it is rebased.

You might want to wait till next week as we are releasing 3.4 this weekend.

ragman1976:
Ok. I´m not familiar with the "maintenance barnch-master system" at GitHub. But I´ll try to make a new request in the master area next week after the release.

Issue #2148, opened 2016-10-23
Telemetry data accumulates timing error


stschake:
Hey,

I noticed that in train mode, the plotted telemetry data will over time fall behind the vertical time indicator line; see the attached picture (all telemetry at 0).

![sync](https://cloud.githubusercontent.com/assets/104132/19623395/08d74640-98c8-11e6-95f2-b295d93c8be1.png)

The problem seems to be that the widget responsible assumes telemetry data is received at a perfect 5Hz:

https://github.com/GoldenCheetah/GoldenCheetah/blob/V3.4-RC1/src/Train/WorkoutWidget.cpp#L190

In practice, however, the telemetry updates are controlled by a QTimer (gui_timer in TrainSidebar) where there is no guarantee that the associated handler is invoked with exactly 5Hz. The QTimer also doesn't seem to adjust the interval as necessary to average out at 5Hz, so over time the inaccuracies accumulate in the error seen above.

I think the best way to fix this would be to record the time elapsed in the workout along with the averaged telemetry values in WorkoutWidget, and use that to determine the X values when drawing the plot series.

stschake:
Fixed with 68461dc

PR #2189, opened 2016-11-17
Don't assume a perfect 1 second between every telemetry sample

stschake:
Instead record the actual recording time and use that to draw individual samples

This fixes issue #2148


bgrabow:
What's the reasoning for defining `paintSampleList` in the header?

stschake:
I figured that was standard practice in C++ for function templates to maintain portability and avoid linker headaches.

bgrabow:
Thanks, I wasn't aware of that practice. It makes sense now that I've read a little more about it.

Issue #2175, opened 2016-11-07
qwkcode -> erg -> qwkcode rounding errors


SparkyT:
Using the new workout editor, rounding of some times result in values changing when the workout is reloaded.
V3.4RC2, build 3947, Windows 10

To reproduce:
- Start a new workout in the designer
- Add 80s@250r80s@150 to the qwkcode editor
- Save workout
- Click away from workout/close GC
- Open workout in editor
- Now shows as 79s@250r80s@150

Is this due to only saving the time points to 2 decimal places?

Larger example attached. .erg file created using qwkcode with 90 rows of data as follows

1s@250r1s@150
2s@250r2s@150
3s@250r3s@150
etc.
etc..
90s@250r90s@150

When loaded as a workout, several rows have times rounded up or down.  Fine for a normal ride but not if following a video.

[numbersWorkout.zip](https://github.com/GoldenCheetah/GoldenCheetah/files/576574/numbersWorkout.zip)

amtriathlon:
This is also reported as https://github.com/GoldenCheetah/GoldenCheetah/issues/2726, let's keep just one.

PR #2221, opened 2016-11-29
Preserve timing values on qwkcode -> ERG -> qwkcode round trip translation.

bgrabow:
The behavior described in #2175 is fixed by f3e377b5683cf7216daf33d82508cbca7d7e71ac.

We already round to the nearest hundredth of a minute when saving qwkcode -> ERG. By rounding to the nearest second when translating ERG -> qwkcode we can preserve the original qwkcode timing.

While investigating I noticed some cosmetic inconsistencies in how fractional seconds are displayed. These cases have been changed to consistently show a value that accurately represents the fractional minute value stored in ERG format.

### DIFF
--- a/src/Train/WorkoutWidget.cpp
+++ b/src/Train/WorkoutWidget.cpp
@@ -1775,7 +1775,7 @@ WorkoutWidget::qwkcode()
         double ap=0;
 
         // how long is this section ?
-        int duration = points_[i+1]->x - points_[i]->x;
+        int duration = round(points_[i+1]->x) - round(points_[i]->x);
 
         bool hasStartLapMarker, hasEndLapMarker;
         hasStartLapMarker = hasEndLapMarker = false;
--- a/src/Train/WorkoutWidget.cpp
+++ b/src/Train/WorkoutWidget.cpp
@@ -795,8 +795,8 @@ WorkoutWidget::setBlockCursor()
 
     bool returning=false;
     QPointF last(0,0);
-    int lastx=0;
-    int lasty=0;
+    double lastx=0;
+    double lasty=0;
     int hoveri=-1;
 
     foreach(WWPoint *p, points_) {
@@ -1917,7 +1917,7 @@ WorkoutWidget::hoverQwkcode()
     for (int i=from; i<=to; i++) {
 
         if (i>from) {
-            int time= points_[i]->x - points_[i-1]->x;
+            double time = points_[i]->x - points_[i-1]->x;
             sumTime += time;
             sumJoules += time * ((points_[i]->y + points_[i-1]->y)/2);
         }


liversedge:
Thank you Ben.

I'm guessing you are using the editor in anger to find that kind of bug.
Would welcome your thoughts on how it could be improved.

bgrabow:
Credit goes to the author of #2175 for discovering the behavior. I've only started using the workout editor to investigate the issue, but I can give my first impressions.

The biggest thing I immediately noticed was how difficult it is to precisely draw a workout. There are a couple improvements that could help:

- Snap points to grid in the drawing area in configurable 1/5/10/15s intervals horizontally, and 1/5/10W intervals vertically.
- Increase the ability to zoom in further than 12m full width. (There are some bugs in horizontal scrolling and the horizontal axis labeling that also detract from the zoomed-in experience. I'll write these up later today.)
- Fix scroll wheel interaction with the vertical scaling of the workout. It's easy to accidentally trigger this when scrolling vertically in the Train pane, and scrolling in the pane takes precedence over scrolling to scale the workout so it's hard to intentionally use this feature when the Train pane has a vertical scrollbar.
- Maybe remove fractional seconds entirely from workouts. It's hard to keep track of fractional seconds through all the unit conversion that happens between ERG -> qwkcode -> WWPoint -> QPoint/QPointF and there is limited usefulness in fractional seconds for a workout file since sub-second changes in effort are unlikely to meaningfully change a workout. Translations can be done to/from ERG without loss of fidelity at 1s granularity as described in f3e377b5683cf7216daf33d82508cbca7d7e71ac. Activities can be sampled to the nearest second when they are converted to a workout. I haven't thought of a good use for supporting fractional seconds otherwise.
- Modify the workout creation wizard (Tools > Create a new workout...) to be denominated in seconds, or unify its syntax with qwkcode. Right now it supports only fractional minutes which is inconsistent with the rest of the workout editor UX.

I'm a software engineer looking for work, and while I'm looking I want to contribute much more to GC so I'm happy to work on some of the above suggestions myself. Tell me if you have any parts of GC that especially need attention!

liversedge:
Thanks!

Is there a piece of functionality you are particularly interested in or emerging tech / science that you want to explore? If so, that would be the best place to start. If not, then by all means keep feedback coming on where things need improving and fix them if you can.

The priorities for v4 are planning and UX. One part of the UX is to unify chart setup to a common codebase and adopt Qt Charts from QT 5.7. This will basically merge all the bits of the current charts into a single widget. If you lookup SAP Lumira that is the ultimate goal (but it will take at least 2 years to get there).

I agree with all of your observations BTW. The snap-to-grid was one of my original plans, but created issues when working with erg files that had already been created (which also happens in MS PowerPoint too). If you have a way around this then I'd love to see it implemented. Also working with blocks not points (but that is a big change).

Kind regards,
Mark

bgrabow:
I'm most interested in trainer control in Train mode. If all that is outsourced to 3rd party libraries I'm also interested in any of the real-time metrics calculated in that mode. I have a background in physics and mathematics and I found your blog posts about live W'bal calculation and maximal/sustained interval calculation very interesting.

I'll take a look at some of my suggestions above and see what I can do. :)

liversedge:
The trainer control is performed by our own code, the RealtimeController is the base class for all devices, see Computrainer.cpp and ComputrainerController.cpp as an example.

What device were you thinking of adding?

If your a mathematician you might be able to help with some of the predictive modelling -- as part of the planning update I'm looking to add predictive modelling via Bannister.

We are also looking to use machine learning ...

bgrabow:
I have a Cyclops Powersync (ANT+) at home that I would like to get connected to GC just for the sake of winter training. I remember it has some quirks that make it different from normal FE-C trainers. I haven't tried to connect it to GC yet but I'll give it a go today.

Any specific parts of predictive modeling that need work? Sounds like fun!

PR #2191, opened 2016-11-19
Kettler serial connection fixes

joekearney:
This provides some small fixes for using a Kettler serial device as discussed in [this mail thread](https://groups.google.com/forum/#!topic/golden-cheetah-users/Gg6TXXM2_c8). It's tested against a Kettler Racer 9.

The commands sent are now upper-cased, and are `ST` for status and `ID` to get the ID of the bike. `CD` returns `ERROR` from the bike, `cd` returns nothing.

cc @erikboto

PR #2276, opened 2017-01-01
Kettler serial connection fixes

joekearney:
This provides some small fixes for using a Kettler serial device as discussed in [this mail thread](https://groups.google.com/forum/#!topic/golden-cheetah-users/Gg6TXXM2_c8). It's tested against a Kettler Racer 9.

The commands sent are now upper-cased, and are ST for status and ID to get the ID of the bike. CD returns ERROR from the bike, cd returns nothing.

---

This replaces #2191, whose commit history I messed up in my forked repo. I've now tested this with my Kettler Racer 9, it finds the device and reads from it successfully. The search dialog actually fails to find the device, though it does give the port `/dev/cu.KETTLER...` in the dropdown. Selecting that does work -- the device is usable and gets data to record.

This is a strict improvement over the previous version, though the search functionality doesn't appear to work yet. Would you be happy merging this now?

/cc @erikboto @liversedge

### DIFF
--- a/src/FileIO/Serial.cpp
+++ b/src/FileIO/Serial.cpp
@@ -380,21 +380,22 @@ find_devices(char *result[], int capacity)
     // relevant for PT downloads. The original list was rather restrictive in this respect
     //
     // To help decode this regexp;
-    // /dev/cu.PL2303-[0-9A-F]+        - Prolific device driver for USB/serial device
-    // /dev/cu.usbserial               - typical for Sewell on Mac
-    // /dev/cu.ANTUSBStick.slabvcp     - Silicon Labs Virtual Com driver for Garmin USB1 stick on a Mac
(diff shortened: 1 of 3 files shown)


erikboto:
Can we hold of for a few more days and I'll make a build and send to the guy that first asked me to add Kettler support? Just want to make sure we don't break anything.

@liversedge Is there any planned dev-releases in the near time? If so I'd like to make sure I get this tested before that so I don't cause unnecessary delays for people waiting for this.

liversedge:
3.4.1 will be out Q2 2017, mostly to remove TrainingPeaks but also to fix a few bugs.

erikboto:
Tested and confirmed to work by the original Kettler user!

liversedge:
missed the update, sorry for the delay in merging.

PR #2285, opened 2017-01-08
Train view

dresco:
Minor train view updates;
 - Move train view settings checkboxes to a new tab for more space
 - Basic implementation of a lap alert sound (fixed sound 3s before new lap)


liversedge:
QSound requires QT 5.7 .. we still build with QT 5.5 until QWebEngine is sufficiently robust -or- we retire the SummaryWindow that depends upon WebKit.

I think you should be able to use QAudioOutput, but please make it optional as some folks still build on QT 4.8

dresco:
Hi, I reckon QSound **is** available under 4.8 (I deliberately chose it instead of anything newer), though I see the Travis build failed..

On an older Fedora box with both 4.x and 5.x, it's part of the base qt-devel for 4.8 (and qt5-multimedia-devel for 5.4), don't know how it's packaged on OSX though?

liversedge:
ah! you need to update src.pro to include multimedia always (its missing on osx)

dresco:
Aha! Sorry, will do that.. 

I'll also change the conditional check to require 5.6+ on Linux only, will at least get the official builds working on the other platforms. Out of interest I'm building under 5.7.1 which appears to have reinstated WebKit - is it making a comeback?

liversedge:
you may have a special patch, are you installed qt 5.7.1 from the distro packages?

dresco:
Yeah, am using the Fedora 25 repo packages for 5.7.1. building without NOWEBKIT and all appears to be working fine - unless I'm missing something. Cheers..

dresco:
Hi Mark, are you happier to pull this this now that it's passing the checks successfully?

liversedge:
Cool !

PR #2295, opened 2017-01-11
Set toolbar height dynamically for optimal height / screen usage

arunh:
Removes the OS specific code, replacing with a generic solution. Effect is that toolbar height is set to be tight on the widgets it contains - gives more screen estate for charts.


liversedge:
Will give it a try .. but may revert since those pixels may be precious on low res displays but might feel cramped on more common hires displays

arunh:
Thanks - if it does feel cramped on larger displays (on my 24" monitor it
looks just right imho, but subjective I know) I could look at making the
padding adjust depending on main window size.

PR #2299, opened 2017-01-14
QWebEngine - Caching

Joern-R:
... do not cache any of the generated HTML pages


Joern-R:
@liversedge  - the "NoCache" option was introduced with 5.7 - so all builds using a older version and QWebEngine before will fail. Shall we make 5.7 as pre-requisite for QWebEngine. If not just delete the PR.

Joern-R:
I will close, since on Mac 5.5.1 is still used.

PR #2307, opened 2017-01-21
Minor build fix

dresco:
@Joern-R Hi, current master is not building for me, just needs another CloudDB check. Cheers..


Joern-R:
@dresco - sorry - missed that spot during some refactoring.

PR #2318, opened 2017-01-28
CloudDB - Ressource Reduction

Joern-R:
... usage update only every 25 starts of GC
... version check only every 10 days

### DIFF
--- a/src/Cloud/CloudDBTelemetry.cpp
+++ b/src/Cloud/CloudDBTelemetry.cpp
@@ -31,17 +31,24 @@
 #include <QDateTime>
 
 
+int CloudDBTelemetryClient::CloudDBTelemetry_UpsertFrequence = 25;
+
 
 CloudDBTelemetryClient::CloudDBTelemetryClient()
 {
-  // only static members
 }
 CloudDBTelemetryClient::~CloudDBTelemetryClient() {
-  // only static members
 }
 
 void
 CloudDBTelemetryClient::upsertTelemetry() {
+
+    // do not update every time, but only every xx times GC is opened to save DB write quota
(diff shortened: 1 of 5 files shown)


liversedge:
I run GC *a lot* because I am debugging. Maybe an option to turn it off in general preferences?

Joern-R:
It' s not the few guys developing which have impact - we have 50k Reads and
20k Writes per Day in the free Quota.
By updating only every 25 times the quota help if we go beyond 10k if
users. By adding it as option to the preferences,
it would be another configuration for just a single (developer oriented)
purpose. Ok ?


2017-01-28 11:38 GMT+01:00 Mark Liversedge <notifications@github.com>:

liversedge:
Fair enough,

I would argue that given we ask users if they want to check for new versions at startup it might be nice to be able to switch if off in preferences if they change their mind e.g. PC in basement/garage and no internet.

How come there are so many calls? The code isn't released yet?

Joern-R:
The check for new version is independent on the Usage Counter. If there is no network the call gracefully fails (I changed the calls to be asynchronous - to no delays in GC startup). 

There are currently not a lot calls - I just did a rough prediction - assuming we have 10k Users 2times a day and all opted in - this would already exceed the quota. So only every 25 Starts of GC the usage is updated.

New version check is independent - is happens without an opt-in dialog, since we are not storing any data. Here the check period is every x (currently 10) days - so that users get the info on new versions, They get info about every new version - but can avoid getting the popup for the same new version every 10 days.

I will add an option of the Location/Usage collection on the preferences - so that a user can "opt-out" at a later point in time. Ok ?

liversedge:
Oh, sorry I see now. Leave it till we see if it is a problem.
I don't want to create unnecessary work

Joern-R:
Ok.

2017-01-28 12:27 GMT+01:00 Mark Liversedge <notifications@github.com>:

PR #2320, opened 2017-01-29
Read R-R data into Xdata

leifwar:
Added reading of R-R into Xdata.


liversedge:
That was quick !!

Do you have any files with HRV data in them to add to the tests/rides directory?

Would be good to be able to test this

leifwar:
Most of the code had already been implemented in what was rejected so it was just a matter of figuring out how to use Xdata. I've pushed test/rides/2017_hrv.FIT which contains the required data. What I don't know how to do is to plot the data. 

Kopi: "Leif Warland" <leif.warland@getmail.no>, "Author" <author@noreply.github.com> 
Sendt: 29. januar 2017 20:12:00 
Emne: Re: [GoldenCheetah/GoldenCheetah] Read R-R data into Xdata (#2320) 



That was quick !! 

Do you have any files with HRV data in them to add to the tests/rides directory? 

Would be good to be able to test this

liversedge:
Plotting the data is easy, you just add a USER DATA series and the formula will be:
XDATA("HRV", "nameofyourseries", interpolate)

See the special topics wiki page for formula syntax and expressions: https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Special-Topics_Formula-Syntax-and-Expressions

liversedge:
screenshot from 2017-01-29 20-16-12

amtriathlon:
Excellent! Good idea, I will do it for SML files since we already decode R-R data for them.

leifwar:
I guess we could do the same for hrm PolarRideFile?

amtriathlon:
Hi Leif, I saw you have already added decoding logic for R-R data to hrm files, it would be nice to add the XData series there too!

leifwar:
I've already implemented the addition of XData series but the problem with the HRM data is that they come in two separate files, one for the regular data and one for R-R only. GoldenCheetah refuse to import them both since they have the same time stamp and I loose the XData if I try to combine.  

I've been wondering if I can "search" for the .pdd file based on hrm filename and use that to find all the files (both hrm and gpx) for a given exercise. Then we can read them all in one go. One problem with this approach would be if the user tries to import many hrm files at the same time, then you might end up trying to import the same exercise many times.

amtriathlon:
> I've been wondering if I can "search" for the .pdd file based on hrm filename and use that to find all the files (both hrm and gpx) for a given exercise.

It seems to me a logic like this is already in place for the .gpx file with the same base name as the .hrm file, the extension to do the same with the .pdd file seems reasonable to me.

leifwar:
I tried to use the "which" command to filter data (which(XDATA("HRV", "R-R", interpolate)<2000, XDATA("HRV", "R-R", interpolate)), but it generates a discontinuous graph in the data. What I want is to use the which command before interpolate. Is that possible?

liversedge:
No interpolate works on the data as found. You could manage it yourself, and perhaps its not needed anyway as the data isn't sparse?

liversedge:
In general when working with user metrics and user data its best to think in terms of iteration rather than functional code.

leifwar:
I'm expanding the HRV Xdata to also include a column with zeros and ones depending on R-R data being valid or not (remove outliers). Then we can filter the data (as they do in https://physionet.org/tutorials/hrv-toolkit/)  when reading and possibly as a function on the Edit menu.  It seems to me that the filtering parameters are something that should be adjustable so I need a way for the user to configure those.  

I will also extend the reading of Polar data to support the files generated with https://github.com/pcolby/bipolar. Here there seem to be two hrm files as before, though the naming is a little different. The hrv data is stored in a file named <name>.rr.hrm where the regular is <name>.hrm.

Do you have any objections to my plans?

liversedge:
No problem with adding an additional column, but how is it being calculated?
When you say filtering, you ignore it when calculating metrics?
Do you have any plans to implement any of the common HRV metrics?

leifwar:
I planed to implement the same filter as http://physionet.fri.uni-lj.si/tutorials/hrv-toolkit/ with:
"filter outliers by deleting those intervals outside of `filt' range of the average within a window `hwin' distance on either side of the current interval", and yes they are ignored when calculating metrics. I also want to ignore them when plotting. Having an extra column gives the user to manually filter out outliers as well. 

My idea was to implement the metrics described on the same web page, but that can be done using the user defined metrics as shown here?: https://dl.dropboxusercontent.com/u/36497635/GC%20user%20defined%20metrics/Stroke%20rate%20breaststroke.png

liversedge:
In this case I think there is a good reason to make them builtin

AartGoossens:
Great to see that you are working on implementing HRV into GC.
Just to weigh in here: Be aware that when you filter the RR data by deleting values it not possible to calculate the frequency domain hrv measures anymore. For time domain measures it is not a problem.

leifwar:
I'm not deleting them, just flagging them with one or zero (outlier). That is what the second column is for.

leifwar:
Is there a way to regenerate the metrics for already imported files?

liversedge:
Change a metadata field e.g. add a space to notes. And press save.
There is no "force refresh" feature... maybe we should add one.

liversedge:
If you are asking how to do it programmatically then yes via RideItem->refresh() .. but you should never need to do this as fingerprints cause refresh -- unless you are creating a new builting metric -- then I just delete rideDB in the cache folder to force a refresh on startup.

There is also a version number in RideMetric.cpp that needs to be incremented when new builtin metrics are added.

leifwar:
Deleting the rideDB worked fine. Will update the version as soon as I have implemented all the hrv metrics (one in place so far).

liversedge:
AWESOME!

Do you have any document or reference for the new metrics?
(I would like to read up on them)

leifwar:
I'm using this https://physionet.org/tutorials/hrv-toolkit

leifwar:
I guess SDANN and SDNNIDX should be ignored as they require 24 hour measurements.

liversedge:
Yes, although it might be interesting to implement separately as an R chart where rides are 24 hours apart.

leifwar:
How do you combine (or group) the trends by day or week?

liversedge:
chart options

leifwar:
I was thinking more on the method used. Is it a mean value?

liversedge:
It depends — you define the aggregation method in the metric definition:
https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/Metrics/RideMetric.h#L56

leifwar:
Standard deviation does not fit into to any of those.

liversedge:
I guess you will need to add one for that.
http://stackoverflow.com/questions/9222056/existing-function-to-combine-standard-deviations-in-r

The aggregation is sprinkled across the code (I haven't had a reason to refactor *yet*, but its starting to smell bad).

The most important place to change will be LTMPlot where the metric curves are created, if you get that woking I'll fix up the rest.

liversedge:
Sorry :)

liversedge:
I suspect you will need to add an additional function/value for metric variances.
This will need to be stored in the rideDB and thats a big ballache for you to do.

If you can confirm this is what you will need, I can add it to the RideMetric code?

liversedge:
LOL. There is a massive regression in the Metric code - count was lost when we moved to the RideDB in v3.2 !!

All averages are likely wrong, and not sure anyone has noticed. Will need to add count to the rideDB too, so adding a variances value will be no issue.

To save space, I will only write variances/count if they are non-zero.

Do let me know if this will work for you?

I'm assuming you will aggregate std dev using that method? (http://stats.stackexchange.com/questions/25848/how-to-sum-a-standard-deviation)

leifwar:
I think the standard deviations are a little bit more complicated as they all have different number of samples (n differ). I think it can be done as shown here (http://stats.stackexchange.com/questions/55999/is-it-possible-to-find-the-combined-standard-deviation), but then you need to store the mean values in addition to the variances.

liversedge:
I'll add stdmean() and stdvariances() functions to the RideMetric and store them in RideDB. This I will have to do over the weekend. So for now, don't worry about aggregating.

I will also add count() to the RideDB since its being lost at present.

Mark

leifwar:
Can I ask (probably a stupid question) regarding the coding of the metrics? In order to make the hrv metrics I've looked at the existing code and made something similar. In all the "examples" that I've seen there is a clone definition which returns a new instance of the class, where *this is passed as input. I cannot see anywhere a  instantiation definition of the class with that argument type. Is there something I'm missing, or will any call to clone fail?

liversedge:
The clone function just returns a copy via e.g:
```
RideMetric *clone() const { return new XPower(*this); }
```
In this example, new ClassName(*this) is the copy constructor that is defined by the compiler, but you can define it yourself (maybe keeping a reference count to some shared resource). It is described here: https://www.tutorialspoint.com/cplusplus/cpp_copy_constructor.htm

You should be able to just define clone in the same way for your metrics.

leifwar:
Status on the implementation of HRV metrics:

1. Implemented a filter functionality similar to filt.c in https://physionet.org/tutorials/hrv-toolkit. This is currently called at the end of RideFileFactory::openRideFile using hardcoded values. I will start on an implementation similar to FixedDerivePower.cpp so that the user can call the filtering HRV with custom parameters. 
2. HRV metrics: RR/NN, AVNN, SDNN, rMSSD (for this one I had to add an extra metrictype which I've called MeanSquareRoot for the aggregation) and pNNi (where i in [5,10,..., 50]).  Currently the aggregation of SDNN is wrong. These give identical values as hrv-toolkit given the same filtering.
3. Re-implemented the plotting of R-R introduced in b47cb73 and removed in c0f58df, this time using Xdata. Should not interfere with the json format. This plotting makes it easier to see the effect of filtering.

liversedge:
Good stuff Leif !

Do you want me to add variances() and mean() to the metric and rideDB?

Mark

leifwar:
Yes, that would be nice.

liversedge:
OK, will look at adding this tomorrow:

1. add count() to rideDB
2. new metric method stdmean() and stdvariance() for aggregating std deviations
3. add stdmean() and stdvariance() to rideDB

liversedge:
Hi I've added count() to rideDB, which I assume doesn't really affect your code.

So -- now I want to look at the HRV stuff can you confirm what you need to store with each metric and how it should be aggregated in the LTM plot?

I'm happy to add all the code (including a new type of metric 'StdDeviation')

leifwar:
Can you take a look at the branch hrv_filter_and_metric? I've not made a pull request as I think it still need some work. 

1. The menu for tuning the filter parameters is there, but I've not figured out how to set default values. 
2. I'm also unsure on some of the hrv settings that I've made in RideFile.cpp as well as AllPlot.{h|cpp}. 
3. The aggregation metric values for standard deviation should be fixed. 
4. Reading of rr.hrm is not tested
5. I put copyright statements in two of the files, though I'm not sure if I have the right to do that (the code is based both your and others code).

liversedge:
Do you have softtabs enabled (i.e. 4 spaces) ?

I'm struggling to read the diffs ! :)

leifwar:
Sorry!

PR #2323, opened 2017-02-01
Read HRV into XDATA

leifwar:
I've implemented reading of HRV into XDATA for polar hrm files. The code scans the pdd file for the exercise if it exist, otherwise it does the same as before. If it does find a pdd file there are three options:

1: only regular hrm data is available, in which case only the main hrmfile is listed
2: both hrm and hrv data is available, with data stored in separate files
3: only hrv data is present, in which case both hrmfile and hrvfile is listed (but identical).

To avoid reading double I "abort" when the user tries to read a hrvfile when hrm data is available.

### DIFF
--- a/src/FileIO/PolarRideFile.cpp
+++ b/src/FileIO/PolarRideFile.cpp
@@ -25,349 +25,511 @@
 #include <assert.h>
 
 static int polarFileReaderRegistered =
-    RideFileFactory::instance().registerReader(
-        "hrm", "Polar Precision", new PolarFileReader());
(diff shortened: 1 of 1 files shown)


liversedge:
Awesome stuff, the file needs to be in the same directory as the source HRM file, right?

We should start thinking about adding builtin metrics for HRV

leifwar:
That's correct, they need to be in the same directory. That is usually the case anyway.  I think for the metrics I will need a little help.

leifwar:
We probably need to filter the R-R values first. At least the ones I have on my garmin watch seem to have a lot of odd measurements.

PR #2333, opened 2017-02-11
Today's Plan

Joern-R:
... do paginated access to the Activities list instead of using fixed limit of 100
   (not that many changes - most DIFFs are due to change in indentation
   by adding a pagination loop around the search request)


liversedge:
Are you using Today's Plan? 
(i.e. have you tested this, so ok to merge?)

Joern-R:
Yes  I am using Today's Plan for Planning since a few weeks (and had uploaded all my history data).
I tested the change with multiple pageSizes with my data. The API is a bit guess-work since documentation is not recent.

PR #2341, opened 2017-02-14
Add Windows crash handling

stschake:
I noticed the old crash report dialog was disabled, which is a good move since it didn't really provide any helpful information. On OSX the system automatically creates a crash report of sorts, but we don't have anything similar for Windows.

So this PR adds a very simple crash handler for Windows 64 Bit (the most popular platform) which generates in the application directory 1) a textual stack trace and 2) a minidump file for use in a debugger. 

I've attached an example stack trace: 
[crash_0006_150217.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/775551/crash_0006_150217.txt)

For development builds, we might consider including the .pdb file in the binary release to get more meaningful traces. This would add approximately 20MiB.

### DIFF
--- a/src/Core/WindowsCrashHandler.cpp
+++ b/src/Core/WindowsCrashHandler.cpp
@@ -1,0 +1,140 @@
+// Only for Windows 64 Bit and MSVC
+#if defined(_MSC_VER) && defined(_WIN64)
+
+#include <fstream>
+#include <cstdint>
+#include <chrono>
+#include <ctime>
+#include <iomanip>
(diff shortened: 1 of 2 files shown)


liversedge:
Nice idea ! will look at it over the weekend

Joern-R:
Works perfectly fine - thanks. 

@stschake 
I will do a small change in an additional commit so that (in most cases) the dump files will be stored in the athlete directory and not in the install directory of GC.

PR #2354, opened 2017-02-21
Fix QT < 5.4 build and QT < 5.0

leifwar:
Both SixCycle and TodaysPlan uses QByteArray::toStdString
(introduced in Qt 5.4), while in QJsonParseError (introduced in Qt
5.0) is used in TodaysPlanWorkoutDownload.

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -456,6 +456,8 @@ MainWindow::MainWindow(const QDir &home)
                          SLOT(uploadGoogleDrive()), tr(""));
     shareMenu->addAction(tr("Synchronise GoogleDrive..."), this,
                          SLOT(syncGoogleDrive()), tr("Ctrl+P"));
+#endif
+#if QT_VERSION >= 0x050400
     shareMenu->addSeparator ();
     shareMenu->addAction(tr("Upload to Today's Plan"), this,
                          SLOT(uploadTodaysPlan()), tr(""));
@@ -485,7 +487,9 @@ MainWindow::MainWindow(const QDir &home)
     optionsMenu->addSeparator();
     optionsMenu->addAction(tr("Create a new workout..."), this, SLOT(showWorkoutWizard()));
     optionsMenu->addAction(tr("Download workouts from ErgDB..."), this, SLOT(downloadErgDB()));
+#if QT_VERSION > 0x050000
     optionsMenu->addAction(tr("Download workouts from Today's Plan..."), this, SLOT(downloadTodaysPlanWorkouts()));
+#endif
     optionsMenu->addAction(tr("Import workouts, videos, videoSyncs..."), this, SLOT(importWorkout()));
     optionsMenu->addAction(tr("Scan disk for workouts, videos, videoSyncs..."), this, SLOT(manageLibrary()));
 
@@ -2046,6 +2050,7 @@ MainWindow::downloadErgDB()
  * TodaysPlan Workouts
  *--------------------------------------------------------------------*/
 
+#if QT_VERSION > 0x050000
 void
 MainWindow::downloadTodaysPlanWorkouts()
 {
@@ -2062,7 +2067,7 @@ MainWindow::downloadTodaysPlanWorkouts()
         "Please check your preference settings."));
     }
 }
-
+#endif
 
 /*----------------------------------------------------------------------
  * Workout/Media Library
@@ -2122,6 +2127,8 @@ MainWindow::syncGoogleDrive()
     upload.exec();
 }
 
+#endif
+#if QT_VERSION >= 0x050400
 void
 MainWindow::uploadSixCycle()
 {
--- a/src/Gui/MainWindow.h
+++ b/src/Gui/MainWindow.h
@@ -189,7 +189,9 @@ class MainWindow : public QMainWindow
         // Training View
         void addDevice();
         void downloadErgDB();
+#if QT_VERSION > 0x050000
         void downloadTodaysPlanWorkouts();
+#endif
         void manageLibrary();
         void showWorkoutWizard();
         void importWorkout();
@@ -226,6 +228,8 @@ class MainWindow : public QMainWindow
         void uploadGoogleDrive();
         void syncGoogleDrive();
 
+#endif
+#if QT_VERSION >= 0x050400
         void uploadSixCycle();
         void syncSixCycle();
 
--- a/src/src.pro
+++ b/src/src.pro
@@ -618,10 +618,12 @@ greaterThan(QT_MAJOR_VERSION, 4) {
     HEADERS += Cloud/Dropbox.h
     SOURCES += Cloud/GoogleDrive.cpp
     HEADERS += Cloud/GoogleDrive.h
+    greaterThan(QT_MINOR_VERSION, 3) {
     SOURCES += Cloud/SixCycle.cpp
     HEADERS += Cloud/SixCycle.h
     SOURCES += Cloud/TodaysPlan.cpp
     HEADERS += Cloud/TodaysPlan.h
+    }
     SOURCES += Train/MonarkController.cpp Train/MonarkConnection.cpp
     HEADERS += Train/MonarkController.h Train/MonarkConnection.h
     SOURCES += Train/Kettler.cpp Train/KettlerController.cpp Train/KettlerConnection.cpp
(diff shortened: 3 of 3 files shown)


leifwar:
I'm using qt 5.3.2 and qt 4.8.6 on my debian machine.

liversedge:
OK, I can see why, but you should be building with QT 5.8 from now on, as it is assumed in the codebase from here. Will merge this, but not likely going to distinguish between too many different versions of QT5 from here.

The new overview uses Qt Charts and will be needed when RideSummary and friends are deprecated (sometime in the next month or so).

At this point you will struggle to keep up with development.

leifwar:
I'll see what I can do, will probably have to download and compile qt myself (been trying to avoid this, prefer stuff that can be provided through package systems). The next stable release of debian (which is in freeze status) seem to provide QT 5.7.

leifwar:
That was easier than I expected.

liversedge:
You can download QT 5.8 directly from the QT website -- the Debian packages are always behind the curve.

leifwar:
That's what I did, and you're right about Debian.

liversedge:
I've added stdmean() and stdvariance() to ride metrics - these are then stored in the ridedb.
When you aggregate you can get at them via RideItem::getStdMeanForSymbol() and RideItem::getStdVarianceForSymbol().

See this commit: 0bb10462cc7e984221ffd64b5d0f30a8446f467d

CHEERS
MARK

liversedge:
Is there any chance you can tidy up the tabs/spacing issues and send me a patch of changes to review?

Quite keen to get this into the next release.

I'm also talking to Polar Flow about integration, and hopefully will use a V800 to test some of your metrics out. Not to mention the fact I'm promoting HRV metrics to the top of the new dashboard (based upon elite coach interviews, how they use this metric).

Your work is very timely !

leifwar:
How do you want the tabs and spaces? I have been struggling a little with the Qt5.8.0 installation but it seem to work now so I will continue on the work.

liversedge:
There is a style guide somewhere, but I can't for the life of me find it !

Basically all tabs should be expanded to 4 spaces.
In general we use K&R brace style, but I'm not fussed about that so much.
Class names should always start with Caps and be CamelCased.
Class instances should always start with lower and be camelCased.
e.g
```
RideFile *rideFile = new RideFile();
if (rideFile == NULL) {
    return;
}
```
The issue with your updates is a lot of the existing code has been adjusted, making it really hard to see what really changed. I spent a fair amount of time back in 2010 un-tabbifying all the sources, so am a little OCD on it !

leifwar:
I have an option in my editor of using K&R style + it is configurable, both the number of spaces and if tabs should be used, so I think it should be possible to me to adapt to your way (I just didn't know).

liversedge:
no worries on the style, its just the tabs that make git think everything has changed, when its just ben reformatted by an editor !

leifwar:
I've done a forced update/replace of the branch where all tabs have been remove. Have attached the patches to these comment. Have not included your stdmean() and stdvariance() nor have done any work on the default parameters for the filter.  So there is still some work to be done before it can enter "master". Maybe you can point me in the right direction for doing this. My idea is to have a "user" default set which can edited in the user options. 
[HRV_metrics.zip](https://github.com/GoldenCheetah/GoldenCheetah/files/794673/HRV_metrics.zip)

leifwar:
Forgot one commit.
[0001-Define-GC_RR_.-values.zip](https://github.com/GoldenCheetah/GoldenCheetah/files/794680/0001-Define-GC_RR_.-values.zip)

leifwar:
There is something wrong with the reading of hrm files. It refuse to read some of my old files claiming they're r-r files. I will look into it.

liversedge:
Hi Leif,

The patch looks solid - thank you !

I noticed you slogged your way through adding a data series to AllPlot - we are moving away from doing that and just asking users to setup User Data series to plot XData. But man, I appreciate you doing the hard yards. I can accept the hack to make XDATA look like a normal data series .. but we might ditch it later.

My only other observation would be the filter dataprocessor has no code comments -- its really difficult to work out what it is doing. But because its a discrete isolated function, it can't do any harm :)

Again -- THANKS. 

Let me know when you're ready to merge.

Mark

leifwar:
Hi Mark,

Is it possible to get hold of context->athlete->cyclist in FilterHrvOutliersConfig::readConfig? Would like to use athlete specific settings as defaults in this menu.

Leif

liversedge:
Assuming you are using a DataProcessor then RideFile::context is available when you process the data.

leifwar:
But is it available when you set the QDoubleSpinBox'es using setValue in readConfig? The way I've done it so far they always comes with a zero value.

liversedge:
OK, now I need to understand better .. what are you trying to achieve from a functional/process perspective ?

leifwar:
The filtering of R-R data consist of four parameters:
1,2) maximum/minimum r-r values, where values of R-R outside these are considered outliers
3) filt constat, where a r-r value outside +- filt times the average of the r-r values in window around the value being considered is considered an outlier.
4) window size, the number of value on both sides (see 3)

I believe these parameters are user specific and my idea was to store them when making a new athlete as well as make them user configurable. The stored values will be used when reading new r-r values (currently hardcoded), but I thought I could use them also as a suggestion when filtering on demand (from the &Edit menu). The picture below shows what this currently looks like (with default value 0): 

filterhrv

liversedge:
Bummer. Data Processors are global in nature and typically to fix/correct bad data.
You don't get a context until working on the ride.
Its not the end of the world in this case. I'll look to see if there is anything we can do about it in the future.

leifwar:
No big deal. We can provide some default values in the above figure, while using the user defined when reading.

leifwar:
Do I need a function to set stdmean and stdvariance (similar to setValue and setCount) that is called in RideMetric::compute()? As I can see from 0bb1046 there are only functions stdmean() and stdvariance() to get them. I'm a little unsure on how I'm going to implement this.

liversedge:
Earlier you said that to aggregate std distributions you would need the mean and variance.

So the HRV std deviation RideMetric needs to provide these via stdmean() and stdvariance(). They will be stored in the ridecache and available when you need to aggregate (which at present is only in LTMPlot I think).

Does this make sense?

leifwar:
So I just implement  sdnn::stdmean() and sdnn::stdvariance(), add another enum metrictype called StdDev call stdmean() and stdvariance() as required in aggregateWith()?

liversedge:
When aggregating in LTMPlot you work with a RideItem* - this is where you retreive the previously computed values via getCountForSymbol() and getForSymbol().. and now there is getStdMeanForSymbol()  and getStdVarianceForSymbol().

So metrics calculate things, they get stored in a cache and can be retrieved via RideItem (and IntervalItem too, but lets do that later)

liversedge:
To your post -- the new enum is good, but will need to implement in LTMPlot (the only place where metrics get aggregated)

leifwar:
I thought it was done in aggregateWith in RideMetric.h. Will implement sdnn::stdmean() and sdnn:sdvariance() and then look into LTMPlot.

liversedge:
Yeah, aggregateWith should be revived or deprecated ! (its dormant)

leifwar:
Is compute() called before stdmean() and stdvariance()? Then I can set private variables stdmean_ and stdvariance_ when I calculate them in compute() and return them in there respective functions.

liversedge:
Yes - that is how it is supposed to work.

compute does all calculations etc and the rest just retrieve the results.

leifwar:
Is there a reason for not making functions where there is identical code?  The code sequence ( `switch (type) { case RideMetric::Total:.....`)  in LTMPlot.cppwhich is repeated several times.

liversedge:
That was my point about refactoring and reviving aggregateWith .. originally it was there once, but as new types of curves arrived code got duplicated.

For example, right now it would make sense to refactor that code, but I suspect you have enough to deal with getting the HRV metric stuff working !

If you update LTMPlot.cpp at line 2715-2733, I'll pick up the rest.

leifwar:
Implemented stdmean() for sdnn and RideMetric::StdDev in LTMPlot. Compared the results with "manual" python scripts combining both two and three rides and got the same result. Now we only miss the configuration of the R-R filtering.

liversedge:
Excellent .. you ok to proceed ? Need anything else?

leifwar:
Actually I got to much as we don't need stdvariance(), it is calculated from value.

liversedge:
Isn't the variance needed to combine when the std deviations are computed from sample sets of different sizes?

leifwar:
Yes, but you already have this by pow(y[n],2).

leifwar:
Should I make a new tab under Options - Athlete or append to an existing tab?

liversedge:
What are you adding?

leifwar:
The parameters for the hrv filter (R-R max/min, filt, window size ..)

liversedge:
Dataprocessor config is already there:
screenshot from 2017-02-27 19-44-50

leifwar:
Ah. Nice!

leifwar:
Do I have to implement case RideMetric:: StdDev: in RideCache::getAggregate? 

What should be the default values for R-R max/min limits (for the filter)? Currently I've put 270 msec (equivalent 60000/270 = 222.22 beats per minute) and 2000 msec (60000/2000 = 30 beats per minute).

liversedge:
No.

RideCache are for plotting distributions and mean-maximal curves. getAggregate() gets an aggregated mean-maximal and distribution for a date range (typically for measures like power, hr)

I do not believe this is worthwhile for R-R data, but perhaps as we get more experience with it that may prove to be false. Either way, adding R-R to the RideCache is a big task -- I will pick that up as it is quite tricky and unfair to burden you with that.

CHEERS

PS: do not worry too much about default values -- so long as they are about right for competitive athletes they should be fine -- the user has to confirm/set them anyway.

Issue #2348, opened 2017-02-18
Offer of Help to Add Virtual Trainer Power


x2764tech:
There are a number of open issues related to  adding virtual power for various trainers, and #337, which would (potentially) simplify the adding of extra trainers.
 I also have 5th order polynomial regression for the (7 settings of the) Minoura B60-R which I'll like to add.

I just wanted to check that no one is already working on this before I start a new branch to refactor the existing support, and then add (some!) additional devices.

liversedge:
If you're up for it resolving #337 would be the best way to do this.. then users can just type in the coefficients.

We could probably add a database to goldencheetah.org (ok, a csv file that gets served from the website) so users could pick from a list.

x2764tech:
Yeah, was thinking at least the code for #337 would be the best way to do this, although looking at the current code, not all the virtual power is a function of speed (but most are).

With regards to the current list, they're referred to by the index of their place in the drop down. I can't see that being 100% compatible with a dynamic list - is there a preference for migrating data?

amtriathlon:
I am closing this issue due to lack of activity, we still have #337 to track this feature.

Issue #337, opened 2012-11-10
[Old 289] Allow user-defined coefficients for Virtual Power


liversedge:
As suggested by Clint Westhoff:
In the “virtual channel” box it would allow for a lot more customization if you had a “custom” option for the virtual power. Maybe when it was selected a box could pop up asking for wheel/tire circumference so that speed can be calculated and then giving the user the opportunity to enter the other variables for the third order polynomial manually. That way anyone with a trainer could have virtual power calculated, so long as they know their power curve, even if their trainer isn’t on the dropdown list.

CustomVPower.png - Custom virtual power UI (3.5 kB)   Darren Hague, 03/28/2011 05:06 pm
Add
Related issues

History
#1 Updated by Mark Liversedge over 1 year ago

Target version changed from 3.0 to 3.1

Deferred to 3.1 since it is a minor feature that will require a fair amount of coding and testing. Or more succinctly, lets put in the 'do it later' pile.

amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/pull/3410

PR #2357, opened 2017-02-23
Add notice that the WIN32-INSTALL Instructions are deprecated

x2764tech:
I've wasted a bit of time trying to follow these old instructions, and, from the look of the issues log and developers forum, I think other people have too.
I've tried to update the Wiki with information as I find it, and this change just adds an instruction to future developers to look elsewhere.

### DIFF
--- a/INSTALL-WIN32
+++ b/INSTALL-WIN32
@@ -1,3 +1,13 @@
+
+THESE INSTRUCTIONS ARE OUT OF DATE. 
+PLEASE REVIEW THE INFORMATION ON THE WIKI AND IN THE FORUMS.
+
+
+https://groups.google.com/forum/#!forum/golden-cheetah-developers
+
+
+
+
 Please note: This is an old build instruction set for Version 2.x.
 
 Most of it is still fairly valid for version 3.x.


liversedge:
Thanks for this David, its really appreciated.

I will wait for Joern to review and merge since he's much more experienced with the Windows builds. I must confess I have neglected the developer guides and instructions.

Mark

Joern-R:
@x2764tech Thanks for adding and sorry for wasted effort - I will merge and since the demand for building under Windows to contribute is increasing - I will start working on an updated guide over the next days. You find first info on how to build on the appveyor.yml - but I will work on giving a "from scratch" guide.

Joern-R:
@x2764tech I never considered it a complaint - for long time no-one was asking - and I actually went through 3 different ways to build on Windows (starting using mingw, msys2,...). With our move to MSVC2015 and a a last step to QT 5.8 things are now much easier for the mainline. So your feedback is fully appreciated to get this now finally documented

liversedge:
Everyone is so polite :)

Anyway.. if you are building, you must have something you wanted to develop?

What did you have in mind? (And how can we help)

CHEERS

x2764tech:
I'm hoping to implement #337 (see #2348 ) - the task itself doesn't seem _that_ hard, but getting a development environment setup is proving to be the stumbling point at the moment.

liversedge:
Perfect, there is also this post on the forums that I will followup, since we could include a database of coefficients to choose from at the same time (https://groups.google.com/d/msg/golden-cheetah-users/vqeqdOxkUck/VHcW_NbvAgAJ).

We could just publish it as a csv or something on goldencheetah.org/trainers.csv ?

x2764tech:
Is this how the system currently works for other reference data (screen layouts, etc)?
But yes, the ability for people to fork the website repo and edit the CSV list would be great.

The only concern I do have about a CSV is having some kind of reference once someone's chosen the co-efficients, and in case they are tweaked later on... and if people might expect this to back-propogate into their training history... but that's probably a problem for a different day.

Some things I've noticed from the code:

- Some trainers curves are based on MPH rather than KPH. My math isn't good enough to do the conversion, so I might include this as an option.
- `BT-ATS - BT Advanced Training System` is based on RPM rather than velocity (so may have to stay hard coded for the moment)
- Different trainers have different _`n`_ for  `Σ  `**_`a`_**<sub>_`n`_</sub>**_`x`_**<sup>_`n`_</sup> - I was think **_`a`_** would be a `std::vector<double>` (or Qt equivalent), and require the csv to have zeros for if it doesn't apply (like the current `Kurt Kinetic` trainers)

So, I'd imagine the csv to look like this (I've typed this by hand, so excuse any mess)

```
name,mph,n,a[n]...
Kurt Kinetic - RoadMachine,1,4,0,5.244820,0,0.019168
1UP USA,1,4,25,2.65,-0.42,0.058
MINOURA V100 on 4,0,4,-10,3.4036,0.1825,-0.00173
```

It might make it easier to have a fixed size of _`n`_, but some of the issues refer to _`n`_` >= 8`, so it would be a lot of zeros for most trainers...

Issue #2363, opened 2017-02-27
ZOOM ?


gregcarpenter:
Is there an option to ZOOM in & out on graphs? Difficult to distinguish points on higher resolution screens.

Cheers
Greg C

freekode:
May be it's not best answer for your problem, but you can change range of a axis.

At the bottom of a graph, also on a left and on a right side (just need to click near the axis).
<img width="1359" alt="screen shot 2017-02-27 at 12 04 50" src="https://cloud.githubusercontent.com/assets/940893/23359110/01d78e50-fce5-11e6-8bfb-4484583a38cd.png">

amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/69b0dd3c2b229858301f50c8f8f9b17e16f462f5

PR #2366, opened 2017-02-28
Zoom & Pan for the AllPlot

stschake:
I had these changes lying around for the AllPlot and saw a few repeated requests for zoom & panning capability, so I figured I would send a pull-request.

- Zoom using the mouse wheel for all plots simultaneously
- Panning when holding the mouse wheel button and moving left/right

I'm not perfectly happy with the feel of this functionality, and there might be a better way to implement it e.g. in the individual charts. Though AllPlot is special in that you want the selection to propagate across all individual plots, and it is also a bit of a trainwreck of code with >>1k LOC handling each metric individually.

So in short, this may not be merge-quality just yet but I figured I would submit it and get some feedback.


amtriathlon:
Can you rebase this?

liversedge:
I'm with Ale, would be nice to see this in AllPlot. Your comment " it is also a bit of a trainwreck of code with >>1k LOC handling each metric individually" made me laugh, you are spot on. It stinks so bad, but the thought of refactoring it is just too much pain for one lifetime. :)

amtriathlon:
I cherry-picked both commits and did some tests:

- I liked the panning facility, I think we can merge it and see how to add it to Metric Trends charts too
- Zoom works well when the plots are not stacked, but when stacked and the plot don't fit the window, the mouse wheel is already being used for scrolling, so the behavior is somewhat strange: if first scrolls and only when you reach the scrolling limit the zooming starts, it works, but it took me a while to get used to, and I don't know if there is a better or more intuitive solution.

liversedge:
can the zoom scroll be disabled in stacked view?

amtriathlon:
> can the zoom scroll be disabled in stacked view?

The stacked view uses a QScrollArea to handle scrolling automatically, if I understand the code correctly, perhaps it could be subclassed to use a modifier (in the line of the accepted response for this https://stackoverflow.com/questions/16105349/remove-scroll-functionality-on-mouse-wheel-qgraphics-view) or extend the panning functionality to do vertical scrolling on vertical mouse wheel button drag.

A simpler solution is to enable mouse wheel zoom only for normal view (not stacked) when there is no collision, this is a single line modification to the original PR and it works quite well, I can push a rebased version including this if you want to test.

amtriathlon:
I rebased this and added a condition to enable zoom only in normal mode, where the use of the mouse wheel doesn't collide with scrolling, I think it is ready for merging.
A similar implementation should work for Metrics Trends chats, as requested in https://github.com/GoldenCheetah/GoldenCheetah/issues/2363

liversedge:
Thanks Ale, will test it. Agree on trends chart too.

amtriathlon:
Thank you!

PR #2368, opened 2017-02-28
Hrv filter and metric

leifwar:
Implemented time domain hrv metrics:

AVNN*	Average of all NN intervals
SDNN*	Standard deviation of all NN intervals
rMSSD*	Square root of the mean of the squares of differences between adjacent NN intervals
pNNi*	Percentage of differences between adjacent NN intervals that are greater than i in (5,10,...,45) ms; a member of the larger pNNx family [6].


leifwar:
Have implemented both: 

- SDANN	Standard deviation of the averages of NN intervals in all 5-minute segments of a 24-hour recording
- SDNNIDX	Mean of the standard deviations of NN intervals in all 5-minute segments of a 24-hour recording

Have not compared them to the results of hrv-toolkit yet.

PR #2369, opened 2017-02-28
Bugfix reading of Polar data

leifwar:


leifwar:
We can get information about "sport" from pdd file if it exist. The pdd file provides an integer value for each exercise where the integer value is defined in a user pdd file. Typically the three first values are set by polar to (1:Running, 2:Cycling and 3: Swimming), while the others are user defined. I've added code to account for these three, but we could implement a configuration (menu) where the user can provide their own definitions of sport, sub-sport and workout code. For my own case I will implement temporary code to account for the values (as I have defined them), port the polar data to GC, and be happy (I'm using GC from now on anyway).  What do you think?

Another issue is the coding. In my case I have hrm and pdd with ISO-8859-1 characters which are "lost" when importing (get a nice question mark instead). I guess this could be handled by GC (though I solved it my case by converting to UTF-8  using iconv prior to the import).

liversedge:
Do not add a configuration menu, the metadata fields already allow for definition of allowed values and that is enough.

I am happy for the pdd file reader to set the sport to run/bik/swim on the basis of the contents of the pdd file. That is good. The user defined ones with have to be linked to "other" or similar.

The ISO-8859-1 characters should be supported by using the right codec when reading the file. See: JsonRideFile.y line 393 for how to open using UTF-8, check for the replacement character and the re-read using ISO-8859-1. Its a kludge, but it works, and only need to worry about it on import.

Hope this helps.

CHEERS

Mark

leifwar:
Just to be clear on the configuration. In my user configuration pdd I have as previously mentioned 1 through 3 (run/bike/swim) predefined by polar, though they can be altered. Number 8 defined by me to be "Ritt" (Norwegian for Bike race), that obviously is (sport) bike though I would like to add workout code "race", and 11  is "Ergometersykkel" (Stationary bike) which also is a (sport) bike though this time I would like to add a sub-sport. Other people will have other definitions, thus they will need a different "conversion". 

I will take a look at the JsonRideFile.y and see if I can adopt to the way it's done there.

liversedge:
Oh, I understand, you want to map a user defined value in one file format to a user defined value in another format. 

For me, Its an edge case that has never been requested that will create a maintenance headache that just doesn't justify the effort.

leifwar:
I agree. I'll add the code for extracting 1, 2 and 3 with default to "run" and leave it at that.

liversedge:
Default should be "Bike" .. majority of users are cyclists.

PR #2372, opened 2017-03-02
Fix Compile error if Secrets not declared for Withings / Twitter

Joern-R:
... happens e.g. in appveyor CI since it has no secrets declared (yet)


Joern-R:
Fix provided by Damien in parallel. So PR can be closed.

PR #2374, opened 2017-03-04
Add missing translatebility

Joern-R:
... for new Overview Windows
... for one special HRV Metric

Overview - PMC values
... Align Headers like in all other Cards (left justified, not centered)

Update Translation DE
.. to be complete and to test the fixes above

### DIFF
--- a/src/Charts/OverviewWindow.cpp
+++ b/src/Charts/OverviewWindow.cpp
@@ -168,37 +168,37 @@ OverviewWindow::setConfiguration(QString config)
         // XXX lets hack in some tiles to start (will load from config later) XXX
 
         // column 0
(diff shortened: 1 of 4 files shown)


sisao:
Maybe variable name "near" leads to compile error with msvc2015_64 and Windows 10, at least on my local machine; renaming solves it for me.

https://github.com/GoldenCheetah/GoldenCheetah/blame/master/src/Charts/OverviewWindow.cpp#L1592

http://stackoverflow.com/questions/22848825/compiler-error-when-using-near-as-a-variable-name

Joern-R:
Thanks - found the same. Fix is done: https://github.com/GoldenCheetah/GoldenCheetah/commit/c2839b9b241cce4d01a31b6812e3c74ad8541b10

PR #2376, opened 2017-03-05
Win10SDK error

Joern-R:
... does not like "near" or "far" as variable names (which seem to be a DOS legacy of near and far pointers)
... surprisingly the SDK version used in CI is fine with it

... renaming fixes the problem - so no further investigation if there are other ways to overcome this.


liversedge:
I just learned something new !

PR #2382, opened 2017-03-08
Fix zones set mismatch

stschake:
This was referencing hrZones when all checks around it used and later accessed zones.

This change should fix a crash first reported by Manuel Oberti on the mailing list:
https://groups.google.com/d/msg/golden-cheetah-users/zRamtWe9-Is/KWpfbcogBAAJ

[Exception call stack](https://github.com/GoldenCheetah/GoldenCheetah/files/828932/call_stack.txt)


liversedge:
Thanks .. classic cut and paste error !

PR #2383, opened 2017-03-08
Overview - Multisport awareness for Sparkline & Up/Down

amtriathlon:
Two small features to try on Overview:
1) Add metric description as ToolTip to METRIC Card
2) Consider only activities with matching sport tags to compute Sparline and Up/Down indicators

### DIFF
--- a/src/Charts/OverviewWindow.cpp
+++ b/src/Charts/OverviewWindow.cpp
@@ -614,13 +614,15 @@ Card::setData(RideItem *item)
         double sum=0, count=0, avg = 0;
         while(index-offset >=0) { // ultimately go no further back than first ever ride
 
-                // get value from items before me
-                RideItem *prior = parent->context->athlete->rideCache->rides().at(index-offset);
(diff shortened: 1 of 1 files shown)


liversedge:
Must confess, I'm not a fan of tooltips... on a related topic I'm thinking about classifying all the metrics, so they can be browsed rather than selected from a list... any thoughts on a taxonomy?????

amtriathlon:
TBF they don't look pretty on Ubuntu, but I think description is something useful to show on METRIC Cards, may be using other way...
WRT to taxonomy some categories may be:
Volume
Intensity
Distribution (aka Time in Zones)
Stress
Measures

amtriathlon:
Removed the tooltips commit, now it only restrict the activities considered to compute Sparkline and Up/Down markers to those matching sport flags since comparisson across sports is usually meaningless.

liversedge:
I like your taxonomy -- its sufficiently broad and consistent -- THANKS!

PR #2384, opened 2017-03-08
Added SDANN and SDNNIDX hrv metric

leifwar:


src/Metrics/HrvMetrics.cpp
```diff
@@ -108,7 +109,8 @@ class avnn : public RideMetric {
 
     void compute(RideItem *item, Specification, const QHash<QString,RideMetric*> &) {
 
-        double total, count;
+        double total;
+        int count;
```
liversedge: Why are you doing this?
You can store integers in a double...

liversedge:
RideMetric::count needs to be a double.

leifwar:
I was not thinking. Just saw the counter and thought that it will always be an integer. Have updated.

leifwar:
Do you have a place/channel for make comments or asking questions to some specific code?

liversedge:
We have a slack I will open a new channel for developer chat and invite you.
What is your email address?

leifwar:
leif.warland@getmail.no

Issue #2387, opened 2017-03-12
Activity, Import from file, Polar *.hrm - Fail in unit


BWonder:
Updated from GC3.3 to GC3.5 dev 1703.
Everything worked on GC3.3 (haven't tested GC3.4)
1)
Reading data  from a Polar HRM file (RS800SD) ignores? the km/mile bit. (in the file) and assumes miles.
Results are workouts that are 1.6X larger when in metric.
(reproducible 3 of 3 GC installs)
2)
Reading data  from a Polar HRM file (15010401.hrm)  results in: 
Error - Could not find in pdd file: "D:/Data/xxxxxxxxxxxx/2015/20150104.pdd"
Both files exits, however don't know why the 2nd (*.pdd) file gives an error while present.
(reproducible on 1 GC install and discovered by trying to combine activities)

As said both errors where not in GC3.3 and examples are imported successfully with that version.

liversedge:
Thanks for the feedback -- Leif added support for HRV data and has possible broken support.
Will look at it.

Any chance you can send me the hrm and pdd files to test with?

liversedge @ gmail.com ...

BWonder:
Mailed 2 of the hrm/pdd files tot test with (much more available if needed).
Will test the 2nd error on a 2nd install tomorrow.

leifwar:
Check if 8a80b3ea fixed the second issue.

BWonder:
@leifwar i quess checking could only be done by installing the complete development environment? Planned in doing that but doesn't look as a very easy process.

Update 2nd error: It looked like this error didn't happen on my work PC, however after testing with the files uploaded to liversedge the errors where the same.
I took a look at the *.hrm and they where exact the same.
The *.pdd where different, in the (newer) pdd there was one extra line with "0 0 0 0 0". 
according to the Polar PDD file format after a [ExerciseInfo1] there should follow 25 lines.
My file had 26 lines, i looks like an update last year in "Polar Protrainer 5" may cause this. (will check tonight)
After manual removing the extra line in the pdd file, importing the hrm no longer generates a pdd error.

1st "bug" still remains
First lines from my hrm file:
[Params]
Version=106
Monitor=37
SMode=110000100
First thing i did notice: according to the Polar HRM2 file format Version=106 has 8 bits (a..h) and versions 1.07 9 bits (a..i). My file has 9 bits however is labeled 106 (Polar Protrainer did this since the beginning in 2006).
[...]

PR #2391, opened 2017-03-13
Fixed use of metric reported in #2387

leifwar:
... The setting of metric was inverted in b3f8f4c1


liversedge:
Thanks Leif !

I should be getting a V800 pretty soon and will be using this a lot for testing :)

Really appreciate you sticking to task after the PR !

PR #2410, opened 2017-03-23
Fix Static Code Check results

Joern-R:
... incomplete class initialization
... duplicate if statements
... typos in objects names (copy&paste) (AllPlot)
... missing else (CPPlot)
... missing DELETEs for objects (only if obvious that object is not used somewhere else)
... copy&paste errors (e.g. ModelPlot)
... small floats instead of integer evaluating to 0
... potential SEGV since pointer is not checked for NULL
... misleading code formatting (hard to read) in *zones.cpp

### DIFF
--- a/src/ANT/ANTMessage.cpp
+++ b/src/ANT/ANTMessage.cpp
@@ -815,7 +815,6 @@ void ANTMessage::init()
     wheelRevolutions = crankRevolutions = 0;
     slope = period = torque = 0;
     sync = length = type = 0;
-    pedalPower = 0;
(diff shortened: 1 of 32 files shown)


liversedge:
Hi Joern,

How did you identify the problems? What tool did you use?

Joern-R:
I used https://www.viva64.com/en/pvs-studio/
and Clang Static analyser. There were a lot more things reported.
But those where I did a fix seemed obvious to be fixed. 
Hope I did not "overdo" it.

I will look into the CI for Mac tonight.

PR #2415, opened 2017-03-25
Body Measures UI at Options/Preferences

### LINKED ISSUE #1954: Weight tab
It would be great if there was a tab under the Athlete menu (in Options) in which the user could save his weight values for different time periods. Just like in Power and Heartrate Zones tabs. Then, these weight values could be used to calculate w/kg for each activity, instead of having the user enter a weight value each time he imports an activity (or let goldencheetah use the default weight value). 

I think this would be also helpful in case a user imports separate weight values from a file. These values could be saved in this tab, and all his activities would have the correct weight value for the corresponding time period. 

Also, as soon as the above does not currently exist, it would be useful if gc didn't read the default weight value when an activity is imported and the user has not entered a new weight value. Instead gc could read the last nonzero weight value from the most recent activities.

amtriathlon:
Moved Height, W'bal Tau, and PMC settings to About tab
Measures tab is a Zones like UI to Body Measures model
Added fingerprint to BodyMeasure to detect changes
Fixes #1954

### DIFF
--- a/src/Cloud/BodyMeasures.cpp
+++ b/src/Cloud/BodyMeasures.cpp
@@ -26,6 +26,25 @@
 #include <QJsonArray>
 #include <QJsonObject>
 
+quint16
+BodyMeasure::getFingerprint() const
+{
+    quint64 x = 0;
+
+    x += when.date().toJulianDay();
+    x += weightkg;
+    x += fatkg;
+    x += musclekg;
+    x += boneskg;
+    x += leankg;
+    x += fatpercent;
+
+    QByteArray ba = QByteArray::number(x);
+    ba.append(comment);
+
+    return qChecksum(ba, ba.length());
+}
+
 bool
 BodyMeasureParser::serialize(QString filename, QList<BodyMeasure> &data) {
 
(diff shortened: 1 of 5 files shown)


liversedge:
The About tab is way too busy and confusing.
I think we should move all the model parameter type stuff and wheelsize etc to a 'Model' or 'Settings' tab.

Joern-R:
Nice enhancement - Question - shall we store the 'bodymeasures,json" in the /activities folder (I did it, but now are not sure any more) - or shouldn't it be in /config (with the other data like Zones,...). Any opinion ?

amtriathlon:
Yes, I agree /config would be a better place

2017-03-25 11:51 GMT-03:00 Joern <notifications@github.com>:

> Question - shall we store the 'bodymeasures,json" in the /activities
> folder (I did it, but now are not sure any more) - or shouldn't it be in
> /config (with the other data like Zones,...). Any opinion ?
>

liversedge:
Apologies, I didn't notice this was a mechanism to let users maintain measures data. This is awesome!
It has been requested a lot on the forums by folks that don't use withings etc.

KUDOS!

PR #2424, opened 2017-03-31
Fix warnings, leaks and null pointer access

mtbkeith:
There are several common issues fixed such as a possible null pointer to be used, division by zero, and memory leaks to happen when a new'd object is not deleted. A smaller nit was that parameters are not marked as unused and therefore many compiler messages get output due to this condition.

### DIFF
--- a/src/Charts/AllPlot.cpp
+++ b/src/Charts/AllPlot.cpp
@@ -4218,8 +4218,12 @@ AllPlot::setDataFromPlot(AllPlot *plot)
         } else if (scope == RideFile::lpco || scope == RideFile::rpco) {
 
             // get the min/max values for both curves
-            double min = qMin(float(thereCurve->minYValue()), float(thereCurve2->minYValue()));
-            double max = qMax(float(thereCurve->maxYValue()), float(thereCurve2->maxYValue()));
+            double min = thereCurve->minYValue();
+            double max = thereCurve->maxYValue();
+            if (thereCurve2) {
+                min = qMin(min, thereCurve2->minYValue());
+                max = qMax(max, thereCurve2->maxYValue());
+            }
             setAxisScale(QwtPlot::yLeft, min, max); // not more than 15 degrees
 
         } else if (scope == RideFile::lppb || scope == RideFile::rppb) {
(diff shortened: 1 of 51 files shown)


liversedge:
NICE!

Its dull work, but really appreciate this, thank you.

liversedge:
Did you perform an analysis with Coverity?
If so, any hints on how we can correct the config here in github?

mtbkeith:
Hi @liversedge The analysis was performed with XCode using 'Analyze'. I was surprised at how it rivalled some of the other commercial tools I've used, however not a true test as I still haven't been able to d/l Coverity. It still shows "Please contact project owner/administrator to submit a build for the project."

src/Charts/AllPlot.cpp
```diff
@@ -4968,6 +4972,9 @@ AllPlot::setDataFromPlots(QList<AllPlot *> plots)
                 }
         }
 
+        if (ourCurve) delete ourCurve;
+        if (ourICurve) delete ourICurve;
+
```
mtbkeith: deleting these curves is a bug; see https://github.com/GoldenCheetah/GoldenCheetah/pull/2438 for backing this change out.

src/Cloud/CloudService.cpp
```diff
@@ -1344,6 +1344,7 @@ CloudServiceSyncDialog::downloadNext()
             QByteArray *data = new QByteArray; // gets deleted when read completes
             store->readFile(data, curr->text(1), curr->text(6));
             QApplication::processEvents();
+            delete data;
```
mtbkeith: can't delete this here, it's required after the read and should be cleaned up elsewhere.

src/Cloud/CloudService.cpp
```diff
@@ -1262,6 +1261,7 @@ CloudServiceSyncDialog::syncNext()
                 QByteArray *data = new QByteArray;
                 store->readFile(data, curr->text(1), curr->text(8)); // filename
                 QApplication::processEvents();
+                delete data;
```
mtbkeith: can't delete this here, it's required after the read and should be cleaned up elsewhere.

PR #2425, opened 2017-03-31
Fix warnings2

mtbkeith:
Another round of compiler warnings and use common constants/defines.

Wko code appeared to have a bug calculating the hours by using 360000 rather than 3600000.

### DIFF
--- a/src/FileIO/WkoRideFile.cpp
+++ b/src/FileIO/WkoRideFile.cpp
@@ -638,8 +638,10 @@ WkoParser::parseHeaderData(WKO_UCHAR *fb)
     } else {
 
         // date not available from filename use WKO metadata
+        // Note: WKO format sure is special, tested with file 2008_06_25_20_59_49.wko renamed to 'my.wko'. 
         QDateTime datetime(QDate::fromJulianDay(julian),
-            QTime(sincemidnight/MS_IN_ONE_HOUR, (sincemidnight%MS_IN_ONE_HOUR)/6000, (sincemidnight%6000)/100));
+            QTime(sincemidnight/MS_IN_WKO_HOURS, (sincemidnight%MS_IN_WKO_HOURS)/6000, (sincemidnight%6000)/100));
+                
         results->setStartTime(datetime);
     }
 
--- a/src/FileIO/WkoRideFile.cpp
+++ b/src/FileIO/WkoRideFile.cpp
@@ -638,7 +638,6 @@ WkoParser::parseHeaderData(WKO_UCHAR *fb)
     } else {
 
         // date not available from filename use WKO metadata
-//< I think this was a bug, incorrectly used 6 minutes of ms as calucations.
         QDateTime datetime(QDate::fromJulianDay(julian),
             QTime(sincemidnight/MS_IN_ONE_HOUR, (sincemidnight%MS_IN_ONE_HOUR)/6000, (sincemidnight%6000)/100));
         results->setStartTime(datetime);
--- a/src/FileIO/WkoRideFile.cpp
+++ b/src/FileIO/WkoRideFile.cpp
@@ -17,6 +17,8 @@
  */
 
 #include "WkoRideFile.h"
+#include "Units.h"
+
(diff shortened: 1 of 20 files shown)


src/FileIO/WkoRideFile.cpp
```diff
@@ -637,7 +639,7 @@ WkoParser::parseHeaderData(WKO_UCHAR *fb)
 
         // date not available from filename use WKO metadata
         QDateTime datetime(QDate::fromJulianDay(julian),
-            QTime(sincemidnight/360000, (sincemidnight%360000)/6000, (sincemidnight%6000)/100));
+            QTime(sincemidnight/MS_IN_ONE_HOUR, (sincemidnight%MS_IN_ONE_HOUR)/6000, (sincemidnight%6000)/100));
```
mtbkeith: using 360000 appears incorrect; dividing by 6 minutes rather than 60 minutes.

src/Metrics/RideMetric.h
```diff
@@ -298,9 +298,6 @@ class UserMetric: public RideMetric {
     // Get the value and apply conversion if needed
     double value(bool metric) const;
 
-    // The internal value of this ride metric, useful to cache and then setValue.
-    double value() const;
```
mtbkeith: this method was just a dup of the base class.

src/FileIO/WkoRideFile.cpp
liversedge: Have you tested this?
It may be an artefact of the WKO file format, rather than a bug.

src/Metrics/RideMetric.h
liversedge: That's a cracker, would never have noticed, a remnant from old code.

src/FileIO/WkoRideFile.cpp
liversedge: Or more bluntly -- don't fix it if you can't prove its broken, the WKO file format is a POS.

mtbkeith: No, I don't have a WKO file to test with and that's why I called it out here. Gotcha, I'm going to update to a new constant.

liversedge: cool -- its never been reported as a bug and would have been spotted by now (you'd hope!) .. this code is 8 years old !

liversedge: BTW there are test files in the test/rides directory, I think there are a couple of WKO files in there.

mtbkeith: I'll test it; this branch is unlikely due to the WKO naming format taking precedence over using the metadata.

src/FileIO/WkoRideFile.cpp
```diff
@@ -636,8 +638,10 @@ WkoParser::parseHeaderData(WKO_UCHAR *fb)
     } else {
 
         // date not available from filename use WKO metadata
+        // Note: WKO format sure is special, tested with file 2008_06_25_20_59_49.wko renamed to 'my.wko'. 
         QDateTime datetime(QDate::fromJulianDay(julian),
-            QTime(sincemidnight/360000, (sincemidnight%360000)/6000, (sincemidnight%6000)/100));
+            QTime(sincemidnight/MS_IN_WKO_HOURS, (sincemidnight%MS_IN_WKO_HOURS)/6000, (sincemidnight%6000)/100));
```
mtbkeith: @liversedge As noted, I did test the file and in fact the numbers are screwy!

liversedge: So there *is* a bug?
Crikey. Thanks.

mtbkeith: No, no sorry WKO is just funky and requires it to be divided by 6 minutes at least in this test file.

liversedge:
THANKS FOR THESE. 

If you have MS VS2015 there are *hundreds* of warnings waiting for you ;)
Albeit mostly with their C and C++ std library headers (!)

Mark

Issue #2421, opened 2017-03-27
Computrainer handlebar controller calibrate button switches calibrate mode on/off too quickly


bwalding:
## Steps

1. Setup train mode
1. Start session (erg / slope etc)
1. Press F3 on handlebar controller
1. GC flicks to calibrate, and then back off calibrate too quickly

## Expected

1. GC accepts the calibration and does not flick back to calibrate mode again

## Comments

I think this is caused by the HBC leaving the F3 button depressed too long - and so GC is detecting the calibrate button as being "pressed" because I was still holding it down after clicking.

I think the solution is to debounce this button - e.g. it won't notice it being pressed again until after it is released.

I don't think this logic should be applied to all buttons (e.g. the up/down buttons) as it is better if they can be held down.

## Patch

I have a patch that I tested successfully today, but then I moved a few things around so need to retest.  You can see changes in commit from my repo.

liversedge:
Let us know when its tested and I'll push.

bwalding:
PR #2429 merged

PR #2429, opened 2017-04-01
Issue #2421 - debounce the calibrate button on the Computrainer HBC

bwalding:
Fix for #2421 - works by preventing a held F3 button from generating more `Calibrate()` invocations until the F3 button is released.

Debouncing the other buttons didn't make sense, so I didn't over-engineer the code.


liversedge:
Yeah that's a clean solution, no timers needed and can only be triggered once.

Issue #2379, opened 2017-03-07
Gear Ratio Precision


GCuser99:
Please add one more decimal (hundredths place) to the calculated gear ratio activity data.

mtbkeith:
@GCuser99 where in the application are you talking about? I don't see a problem but I just created my own chart see here:

<img width="1192" alt="grab_gearratio" src="https://cloud.githubusercontent.com/assets/9326275/24575387/8c002ad4-1661-11e7-8e09-9c089fc9441f.png">

GCuser99:
Under Activities->Ride, with gear ratio curve turned on. See chart below... Also, I wrote an R script to write the calculated gear ratio to file and it only output one decimal place... I had some ideas of how I could use gr to auto-edit out cadence spikes but need one more decimal place for it to be of any use. 

![image](https://cloud.githubusercontent.com/assets/26237126/24579890/2ee612ee-16b3-11e7-8666-936b37b3e2a8.png)

mtbkeith:
Hi @GCuser99 I looked at this briefly; The window code simply shows a tooltip of the calculated gear ratio is that all you are looking to change? It might be helpful to hear a use case.

There are other areas of the code in which the ratio is calculated and preserved into the JSon file and subsequently would then could potentially make it into output files like CSV. 

<img width="1196" alt="grab-gear-ratio-testing" src="https://cloud.githubusercontent.com/assets/9326275/24581116/535d5212-16d2-11e7-9eb7-428766a24131.png">

PR #2431, opened 2017-04-01
Change gear ratio precision to hundredths.

mtbkeith:
** Fix comparison of headers which simply used ‘contains’ rather than equals and would cause incorrect precision to be selected if a translation/title is not found.

Tooltip in ‘Ride’ view will now show hundredths due to precision selection.
Change rounding of gear ratio in derived series to hundredths.
@liversedge does this make it into the JSon or out into CSVs?

See request:
https://github.com/GoldenCheetah/GoldenCheetah/issues/2379#issuecomment-290926521

### DIFF
--- a/src/Charts/AllPlot.cpp
+++ b/src/Charts/AllPlot.cpp
@@ -7018,18 +7018,19 @@ AllPlot::pointHover(QwtPlotCurve *curve, int index)
         }
 
         // need to scale for W' bal
-        if (curve->title().text().contains("W'")) yvalue /= 1000.0f;
+        if (curve->title().text().contains("W'"))
+            yvalue /= 1000.0f;
 
         // output the tooltip
-    int precision;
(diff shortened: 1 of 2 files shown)


src/Charts/AllPlot.cpp
```diff
@@ -7018,18 +7018,19 @@ AllPlot::pointHover(QwtPlotCurve *curve, int index)
         }
 
         // need to scale for W' bal
-        if (curve->title().text().contains("W'")) yvalue /= 1000.0f;
+        if (curve->title().text().contains("W'"))
+            yvalue /= 1000.0f;
 
         // output the tooltip
-    int precision;
-    if (curve->title().text().contains("Hb")){
```
mtbkeith: The precision would default to 1 if running a translated version and did not contain expected Hb/R-R..

PR #2435, opened 2017-04-02
CloudService - Fix Error Reading Files

Joern-R:
... when asynchronously reading Files, "data" object needs to survive as it it referenced in the "buffers_"   QMap
    (this is a quick fix - I suspect there are better ways to handle this)

### DIFF
--- a/src/Cloud/CloudService.cpp
+++ b/src/Cloud/CloudService.cpp
@@ -1261,7 +1261,7 @@ CloudServiceSyncDialog::syncNext()
                 QByteArray *data = new QByteArray;
                 store->readFile(data, curr->text(1), curr->text(8)); // filename
                 QApplication::processEvents();
-                delete data;
+                //delete data;
 
             } else {
                 curr->setText(7, tr("Uploading"));


liversedge:
Is there a crash if you don't comment out that delete?

Joern-R:
Yes - the crash happens here:

void

TodaysPlan::readyRead()

{

    QNetworkReply *reply = static_cast<QNetworkReply*>(QObject::sender());

    buffers.value(reply)*->append(reply->readAll());*

}


since the buffer value references *data, but data does not exist any more.


2017-04-02 19:04 GMT+02:00 Mark Liversedge <notifications@github.com>:

liversedge:
OK, will look into deleting when buffers goes out of scope

mtbkeith:
@Joern-R @liversedge Sorry for introducing this issue. XCode reports it as a leak and I was unaware of the dependency on it, despite it not having a clear reference. In the end it probably is a leak. 

The same happened with another delete of the curves causing an issue with the comparison; at least I was the one who found that and ultimately figured it out.

liversedge:
Its ok, I have a love/hate relationship with QT's object hierarchy and owenership semantics. When you have dependency across the object hierarchy, threads and time it all gets painful.

PR #2432, opened 2017-04-02
Make the +/- buttons render consistently

bwalding:
There is an issue with native lnf rendering a QPushButton on OSX with small text labels (a very bizarre issue IMO).  E.g. as you can see in the screenshot, the QPushButton with a "-" renders with an ugly lnf. Whereas the "+" renders properly. The cause seems to be the width of the text label - ` - ` renders with a native lnf, but `-` renders with a default toolkit lnf.

QT: 5.8.0
QWT: 6.1.1
OS: Mac OSX 10.12

I have copied the setup pattern from src/Gui/Pages.cpp - I can't comment on whether it is a sane pattern or not.

### Before

screen shot 2017-04-02 at 11 38 55 am

### After

screen shot 2017-04-02 at 11 50 51 am

### Future work

It seems like this is a common metaphor through the app - "addition and removal" of list items.

Doing a search for `QPushButton("` (for untranslated buttons yields 23 hits - mostly `<`, `>`, `+`, `-` - and it seems some authors have used ` - ` to work around the rendering bug in the workout scanner.

```
$ grep -ir 'QPushButton("' src
src/Charts/AllPlotWindow.cpp:    addCustomButton = new QPushButton("+");
src/Charts/AllPlotWindow.cpp:    deleteCustomButton = new QPushButton("-");
src/Charts/AllPlotWindow.cpp:    scrollLeft = new QPushButton("<", this);
src/Charts/AllPlotWindow.cpp:    scrollRight = new QPushButton(">", this);
src/Charts/ExhaustionDialog.cpp:    deleteRefButton = new QPushButton(" - ");
src/Charts/LTMTool.cpp:    addCustomButton = new QPushButton("+");
src/Charts/LTMTool.cpp:    deleteCustomButton = new QPushButton("-");
src/Charts/LTMWindow.cpp:    scrollLeft = new QPushButton("<", this);
src/Charts/LTMWindow.cpp:    scrollRight = new QPushButton(">", this);
src/Charts/ReferenceLineDialog.cpp:    deleteRefButton = new QPushButton(" - ");
src/FileIO/XDataDialog.cpp:    addXData = new QPushButton("+", this);
src/FileIO/XDataDialog.cpp:    removeXData = new QPushButton("-", this);
src/FileIO/XDataDialog.cpp:    addXDataSeries = new QPushButton("+", this);
src/FileIO/XDataDialog.cpp:    removeXDataSeries = new QPushButton("-", this);
[...]

PR #2436, opened 2017-04-02
Make the +/- buttons render consistently

bwalding:
See #2432 for an earlier concept that was closed.

The root cause is that the size of the "-" button is less that 50x30 and so Qt changes from a rounded button to a square button.

https://doc.qt.io/qt-5/qpushbutton.html

"Note that when a button's width becomes smaller than 50 or its height becomes smaller than 30, the button's corners are changed from round to square. Use the setMinimumSize() function to prevent this behavior."

### DIFF
--- a/src/Train/Library.cpp
+++ b/src/Train/Library.cpp
@@ -263,13 +263,17 @@ LibrarySearchDialog::LibrarySearchDialog(Context *context) : context(context)
     findMedia->setChecked(true);
     findVideoSyncs = new QCheckBox(tr("VideoSync files (.rlv)"), this);
     findVideoSyncs->setChecked(true);
-    addPath = new QPushButton("+", this);
-    removePath = new QPushButton("-", this);
-    removeRef = new QPushButton("-", this);
+    addPath = new QPushButton(tr("+"), this);
+    removePath = new QPushButton(tr("-"), this);
+    removeRef = new QPushButton(tr("-"), this);
 #ifndef Q_OS_MAC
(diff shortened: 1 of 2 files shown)


liversedge:
Any clues why the '+' button renders ok?

bwalding:
@liversedge Because it is just over the size limit - even changing the text on the "-" to " - " fixes the LNF.

RemovePath - 49x32 (only 1 pixel short of rendering with rounded corners)
AddPath - 51x32

liversedge:
Then I think the fix will be to change "-" to " - ".

liversedge:
I've been reflecting on this and am starting to think your solution is better.
a) add, delete have a much clearer meaning
b) +/- have more techy connotations that we should avoid
c) when we have 'edit' the +/- metaphor is jarring

You fancy doing a global search and replace for +/- buttons.. remembering that they will also have a fixed size of 20x20 or similar to unpick ?

bwalding:
Ok - will have a look in next few days.

"Add" / "Delete" is used a fair bit already - which is what I based original work off too.

To add more weight to that approach - I was using my non-retina laptop (which is what I use with the Computrainer) and it displayed both buttons with square layout - fiddling around with the fonts caused the button to change layouts.

gcoco:
Are you still planning on submitting a patch for replacing +/- with Add/Delete?
If so can you close this Pull request and submit a new one with the changes.
Thanks

bwalding:
I'll create a new one when I get time.  Weather has been good, so I haven't been using that area of GC (as I only use that for indoor).

Issue #2433, opened 2017-04-02
Crash when comparing two rides


mtbkeith:
When I attempt to compare two rides and then view the 'Ride' tab. 

With a quick look the code demonstrating the issue is this line of code and while accessing the size member. 
seriesPlots[i]->compares[0]->data()->size()

If I comment out that paragraph of code which deletes/removes the seriesPlots the program continues as one would expect and then the graphs show correctly. Either the memory has already been released, something overwrote the memory or incorrectly allocated.

Happens with the latest codebase; SHA 81170a947d5be35704711821681645e75b3bbb1c
Ride files: 3_11_2017 (base) & 3_18_2017
OS: MacOSX, MacOSX10.12.sdk, Qt5.8.0 

The change which breaks the comparison, something Xcode suggested as a memory leak?
SHA da1cd0fcda67690853388222d9e72458b499f4d7

[gc_crash_comparingTwoRides_sha.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/888205/gc_crash_comparingTwoRides_sha.txt)

liversedge:
WHOA!

It looks like i is being incremented beyond the end of the array !?
In the debugger I see a huge value, but cannot for the life of me see how that can happen.

Are you looking into this?

mtbkeith:
Yes for now you can mark it as assigned to me. I bet it is either stomped
or released.

Thanks!

liversedge:
Been trying to track down the change that causes this, so far I've only manually bisected to d2e8b5f87b39f1cf3c1e4954ebbca28eca951afa where it doesn't crash.

mtbkeith:
See the fix, I was the culprit in the first place.

PR #2438, opened 2017-04-03
Fix crash of comparison rides

mtbkeith:
Xcode reported this as a leak however, ownership of the curve is owned by this class and will be deleted by the Qt framework as the parent is deleted.

Fixes issue I reported;
https://github.com/GoldenCheetah/GoldenCheetah/issues/2433

Sorry about that!


liversedge:
Thanks. Thanks for the swift fix !

PR #2451, opened 2017-04-15
Provide zoneslow and zonescolor in GC.athlete.zones().

sisao:
Hi,
this commit provides the lower ends of the powerzones and the matching zone colors for R charts.

for Example:
activity <- GC.activity()
metrics <- GC.metrics()
date <- metrics$date

zone <- GC.athlete.zones(date=date, sport="bike")
lows <- zone$zoneslow
colors <- zone$zonescolor

p <- activity$power
p2 <- p[seq(1, length(p), 4)]
breaks <- c(unlist(lows), Inf)
cols <- c(unlist(colors))[findInterval(p2, vec=breaks)]

barplot(p2, space=0, border=NA, col=cols)

Is it possible to add this to GC or should this be done in a different way?

### DIFF
--- a/src/R/RTool.cpp
+++ b/src/R/RTool.cpp
@@ -340,6 +340,7 @@ class gcZoneConfig {
     bool operator<(gcZoneConfig rhs) const { return date < rhs.date; }
     QString sport;
     QDate date;
+    QList<int> zoneslow;
     int cp, wprime, pmax,ftp,lthr,rhr,hrmax,cv;
 };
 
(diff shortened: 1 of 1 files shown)


liversedge:
nice.

liversedge:
Can you update the wiki page to reflect the new changes?
https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Special-Topics_Working-with-R

liversedge:
Thanks for updating the wiki, nice job.

sisao:
thx,
I will provide a better "power in zones" graph later today with screenshot, 
but I have to read a little more in the R documentation...

PR #2459, opened 2017-04-19
Use correct units for CV hover dialog.

mtbkeith:
Previously shows both units, despite config in preferences. 

Before change:
grab-cv-before-unit-fix

After fix:
grab-cv-fix-units


mtbkeith:
CP dialog not showing watt value with this change. will investigate.

PR #2462, opened 2017-04-20
Use units specified in zone settings for run/swim/default.

mtbkeith:
Using Preferences->Athlete->Zones->Pace Zones for both Run/Swim to override Preferences->General->unit in tooltips.
Simplify some code with if/else blocks.
Reorganize Tooltip variable argument order to match placement in text.
Still several problems around kph being used in chart CV y-value despite settings.
Use const when appropriate.

Using metric in swim pace zone:
grab-swim-with-metric-units-in-zone
Using imperial in swim pace zone:
grab-swim-with-imperial-units-in-zone

### DIFF
--- a/src/Charts/AllPlot.cpp
+++ b/src/Charts/AllPlot.cpp
@@ -6989,8 +6989,10 @@ AllPlot::pointHover(QwtPlotCurve *curve, int index)
     if (index >= 0 && curve != standard->intervalHighlighterCurve && 
                       curve != standard->intervalHoverCurve && curve->isVisible()) {
 
+        int precision = 1; //< default to tens precision
+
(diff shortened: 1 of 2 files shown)


liversedge:
We generally adopt K&R style formatting across the code, would be good if you could adopt that for changes since its rather jarring to see Allman style for small sections of code.

Sorry to be picky, but its like a wonky picture, catches the eye and detracts from what is important.

mtbkeith:
Will do; there has been a mix in some of the files.

liversedge:
Yes, little patches here and there do that !

TBF, some of the past contributors mixed up all sorts of styles, I did a blitz in v3.0 to remove some of the "worst". Being an old hack I can't see beyond K&R ;)

PR #2468, opened 2017-04-21
Re-introduced vector expressions

### LINKED ISSUE #2278: Vector formulas don't work anymore in V3.4
Reported at the forum: https://groups.google.com/forum/#!searchin/golden-cheetah-users/vector%7Csort:date/golden-cheetah-users/WmUGbNHviJE/Gb-ZdXn-DwAJ

amtriathlon:
Using |Date_From:Date_To| instead of [Date_From:Date_To] to
avoid conflicts with array indexing
Fixes #2278

### DIFF
--- a/doc/user/formula-syntax.txt
+++ b/doc/user/formula-syntax.txt
@@ -23,6 +23,10 @@ isinf(p1)
 isnan(p1)
 
 LIST / VECTOR REDUCE FUNCTIONS
+Vectors are about collecting data from across a date range, specified as
+    expression[[Date_From:Date_To]]
+where dates can be absolute s.t. 2017/09/19 or relative s.t. Date-7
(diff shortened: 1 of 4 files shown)


liversedge:
We should avoid using | .. why not use [[ and ]] ?

liversedge:
But also, thanks for re-introducing it !

liversedge:
In terms of how we would do this -- the lexer should return something for '[[' and ']]' patterns.
e.g. LSB and RSB for left square bracket and right square bracket.

amtriathlon:
I used | because it was simpler and it doesn't generate conflicts, but I understand | could be useful in the future, I will change for [[]] and push an update.

liversedge:
thanks!

PR #2471, opened 2017-04-23
Default Weight vs. BodyMeasurements

Joern-R:
... separate (single) default weight stored in GC_WEIGHT from time-dependent body measurements on UI
... To Do - analyse and if necessary adjust use of GC_WEIGHT vs. use of time-dependent body measurement


Joern-R:
@amtriathlon - Ale can you please have a look. The change is to separate the data on the UI - I faced the fact that changing the weight on the UI updated "GC_WEIGHT", but did not create a new body measure. Since the logic of the "default" and the "time-dependent" measures may not be obvious for many users I tried this approach.

I still need to check the places where GC_WEIGHT is used directly in the code - vs. reading the weight via the functions which do the cascaded access - BodyMeasure, RideItem. GC_WEIGHT, Default.

Joern-R:
Noticed some missing changes - so closed - new PR on the way

PR #2472, opened 2017-04-23
Default Weight vs. BodyMeasurements

Joern-R:
... separate (single) default weight stored in GC_WEIGHT from time-dependent body measurements on UI
... To Do - analyse and if necessary adjust use of GC_WEIGHT vs. use of time-dependent body measurement

### DIFF
--- a/src/Gui/Pages.cpp
+++ b/src/Gui/Pages.cpp
@@ -865,12 +865,40 @@ RiderPhysPage::RiderPhysPage(QWidget *parent, Context *context) : QWidget(parent
     QString weightUnits = (metricUnits ? tr(" kg") : tr(" lb"));
 
     QVBoxLayout *all = new QVBoxLayout(this);
-    QGridLayout *grid = new QGridLayout;
+    QGridLayout *measuresGrid = new QGridLayout;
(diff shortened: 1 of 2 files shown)


Joern-R:
@amtriathlon - Ale can you please have a look. The change is to separate the data on the UI - I faced the fact that changing the weight on the UI updated "GC_WEIGHT", but did not create a new body measure. Since the logic of the "default" and the "time-dependent" measures may not be obvious for many users I tried this approach.

I still need to check the places where GC_WEIGHT is used directly in the code - vs. reading the weight via the functions which do the cascaded access - BodyMeasure, RideItem. GC_WEIGHT, Default.

amtriathlon:
Hi Jörn, I agree it would be better to keep the original default weight separate from time dependent measures, my approach would be to put default weight together with height on about page but I think your implementation is clear enough.

Joern-R:
Hi Ale,

thanks for the feedback. I was thinking about that as well - it's the trade-off of keeping "same things together".

All default values on one Tab vs. All Weight data on one Tab. I will merge and then work on the ToDo.

Cheers.
Joern

PR #2478, opened 2017-04-25
Only listen on localhost for API calls

stschake:
This changes the enabled by default HTTP API to only listen on localhost, not all interfaces.

The issue with listening on all interfaces is that it enables access potentially to the whole internet depending on your local network configuration. Since the API is unauthenticated and accesses are not indicated in the UI, this could result in someone silently downloading your entire training data without any notice. One added benefit of this change is that the Windows firewall popup should no longer appear.

For advanced users, listening on all network interfaces remains possible by editing the *httpserver.ini*


liversedge:
Great stuff. Thanks.

PR #2490, opened 2017-04-28
Support for SRM manual zero offset calibration

dresco:
As there is no notification of calibration completion in the SRM profile,
we progress the calibration state on to completion five seconds after we
start receiving calibration messages.

Temporarily removes support for auto-zero, as was not used, and current
code did not follow the standards set in the ANT+ documentation.


liversedge:
COOL !

Thanks for this. Will try and give it a go. But will be in this weekends development build.

PR #2494, opened 2017-04-29
Fix the Zwift activity extension since it is not added by Zwift durin…

bwalding:
…g automatic upload.

Zwift registers activities with Today's Plan using the filename  "zwift-activity-[id]" - when it should have a `.fit` file extension.

I hate having special cases like this - but the alternative is that GC downloads files from TodaysPlan that have no real data in them (because it doesn't recognise the zwift-activity file - and so just registers an empty shell of a ride file).

I'm going to continue badgering Zwift to fix at the source - but my first attempt at getting them to fix their upload system was unsuccessful.

### DIFF
--- a/src/Cloud/TodaysPlan.cpp
+++ b/src/Cloud/TodaysPlan.cpp
@@ -289,8 +289,15 @@ TodaysPlan::readdir(QString path, QStringList &errors, QDateTime from, QDateTime
                 // file details
                 QJsonObject fileindex = each["fileindex"].toObject();
                 QString suffix = QFileInfo(fileindex["filename"].toString()).suffix();
-                if (suffix == "") suffix = "json";
-
+                if (suffix == "") {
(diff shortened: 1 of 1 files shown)


liversedge:
Damien and Joern need to review this since TodaysPlan::readFile converts the response to JSON.
Having a workaround for the bug in the TodaysPlan implementation is ok though.

Joern-R:
@grauser - I will ask Today's Plan if they can fix it. I suspect they have integration source specific code anyway / or can better align with Zwift if it's their issues. Will let you know the result and decide the PR on this.

Joern-R:
I checked with the Today'sPlan team. They are aware on the fact, but always keep in the DB exactly what they get from a linked service. Recommendation is to do what the PR is doing. So let's merge.

Thanks for bringing it up.

PR #2500, opened 2017-05-01
Disabled Linux OpenGL acceleration if OpenGL lower than 2.0

amtriathlon:
I don't know if 2.0 is the minimum version required or it needs to be higher, mine is 1.4 and it doesn't work.


liversedge:
brilliant, thank you !

PR #2502, opened 2017-05-02
Add QT+=opengl to fix build

erikboto:
I'm not sure if the plan is to make OpenGL part of other systems than linux, if not maybe the GLWidget include should be conditional which would also allow the QT+=opengl to be conditional. Otherwise this PR should be enough.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -52,7 +52,7 @@ lessThan(QT_MAJOR_VERSION, 5) {
 } else {
 
     ## QT5 modules we use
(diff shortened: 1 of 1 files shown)


maxammann:
Haha, a few minutes too late :D https://github.com/maxammann/GoldenCheetah/commit/7d46c31175b9b511fa20672b1c4dc7a598f3cc53

maxammann:
This fix is testet then though :D

liversedge:
I'll add it to the section that declares OverviewWindow so we only add it if we need it.

Thanks.

PR #2503, opened 2017-05-03
TCX export: don't write bogus HR values

stschake:
I noticed a trainer ride showed a bogus HR chart with all "1" values after I shared it from GC to Strava. I traced that back to the TCX export which always wrote HR data if *withHr* was true (which it is for every invocation in GC!).

There is a mysterious comment on a HeartRate hack, but always writing HR of 1 is clearly not correct when the ride has no heartrate info. Also, the XSD allows for this data to be omitted.

With the commits here applied, the HR chart on Strava also promptly vanished. I also added notes and device type to the exported file if the ride has them, but Strava doesn't care for them (or certainly didn't for a creator of "GoldenCheetah"). Notes are apparently only taken from their side-channel form upload..

All changes here are cross checked with the TCX XSD and should be correct.

### DIFF
--- a/src/FileIO/TcxRideFile.cpp
+++ b/src/FileIO/TcxRideFile.cpp
@@ -88,6 +88,14 @@ TcxFileReader::toByteArray(Context *context, const RideFile *ride, bool withAlt,
     id.appendChild(text);
     activity.appendChild(id);
 
+    // notes if present
+    if (ride->getTag("Notes","") != "") {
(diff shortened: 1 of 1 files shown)


liversedge:
Its possibly a cut and paste error. The PWX exporter used to write values of 1 when HR or Power was 0 to get around the way TrainingPeaks processed files (and rejected ones with >3 zeroes in a row for no apparent reason).

PR #2504, opened 2017-05-03
Strava upload: don't use filename as fallback for activity name

stschake:
The Strava cloud service currently uses the filename if no field is configured as activity name source (but there is currently no option in the GUI to do that). This results in an activity name such as "2017_05_02_22_36_29".

The old share code didn't send an activity name at all, which results in a Strava automatically generated name. Since the filename for the majority of devices out there is autogenerated, I think this behavior is preferable. Once there is an option in the interface to set the metadata field, "no setting" should equal not sending an activity name, so I think this is the correct behavior in that instance also.

(I also noticed the form field was called "activity_name", but the Strava documentation says the correct field is called just "name": https://strava.github.io/api/v3/uploads/)

### DIFF
--- a/src/Cloud/Strava.cpp
+++ b/src/Cloud/Strava.cpp
@@ -275,14 +275,12 @@ Strava::writeFile(QByteArray &data, QString remotename, RideFile *ride)
     QString filename = QFileInfo(remotename).baseName();
 
     QHttpPart activityNamePart;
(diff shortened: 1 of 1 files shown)


liversedge:
Didn't realise it generated a name. your solution is better, thank you.

stschake:
You're absolutely right, sorry, I didn't test the positive case (activityName set). I'll PR a fix.

PR #2506, opened 2017-05-03
Strava upload: always set activityNamePart body

stschake:
This fixes an issue with #2504 spotted by @amtriathlon 

My apologies for the oversight. I've confirmed that the configured tag is now used as activity name, and, when none is set, the Strava generated name.


amtriathlon:
Thanks!

PR #2511, opened 2017-05-07
Share Menu

Joern-R:
... move "Body Measurement Download" from "Tools" to "Share"
... enable "Check for New Data" menu action


liversedge:
cool. good idea.

PR #2517, opened 2017-05-17
Athletes weight setting.

vk7rb:
Just changed the athletes weight and it then went backwards and changed it for all previous rides which is invalid as my weight had been correct for all previous rides and should only have been changed for the current ride and any future rides until such time as another adjustment was made.

This definitely needs working on as the calculations for all my previous rides are now invalid due to incorrect weight.


liversedge:
Cannot review in this form. Please rebase and submit with a description of the change being made.
You should also review the changes for athlete weight in v3.5 which has introduced time based weight settings. Prior to this the weight is universal and not time based so applies to all rides.

PR #2519, opened 2017-05-18
DI2 ANT+ messages used to populate dedicated XData serie

vlcvboyer:
This patch read DI2 gears which are part of ANT+ messages in FIT file.
Additional XData serie is created and populated with those values.

...special thanks to Joern who worked on building GC with VC thereby allow me to participate again to this project !

### DIFF
--- a/src/FileIO/FitRideFile.cpp
+++ b/src/FileIO/FitRideFile.cpp
@@ -136,6 +136,7 @@ struct FitFileReaderState
     QVariant isGarminSmartRecording;
     QVariant GarminHWM;
     XDataSeries *weatherXdata;
+    XDataSeries *gearsXdata;
     XDataSeries *swimXdata;
     XDataSeries *deveXdata;
     XDataSeries *extraXdata;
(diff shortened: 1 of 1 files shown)


Joern-R:
Thanks - I had this on my list (since I just got a DI2 this year :-) 
I will change the "gear" derivation so that is uses the real so it uses
the real gears to calculuate the ratio (if XDATA for "GEARS" is present.)

vlcvboyer:
Yes this is also in my TODO list... but is really not my priority.
In case you do it before me : please consider that you may not have received any message during the first minutes of ride. Thus I propose to derive gear-ratio by using existing formula until timestamp of first DI2 message is detected.
I'm not sure to be clear...

Joern-R:
I have already added it - see latest commits - but I did not consider your point. In my data I had any event exactly at "0" so did not notice. Question is whether to calculated in a mixed way - or just take the "first real event" also as the "starting event" and then have a starting ratio - which is wrong by one shift - but probably better than the bad quality of the speed/cadence based calculation - what do you think ?

PR #2520, opened 2017-05-18
Fix gear ratio calculation when no power meter is present

vlcvboyer:
Previous code do not allow to calculate gear-ratio when no power meter is used (as power > 0 is used to filter and reduce inaccurate data)
However when no power meter was used during the ride, as per my point of view, it is better to have few inaccurate data than no data at all.
In this patch I kept the existing filtering when power meter is there. Thus accuracy, when using power meter, is not affected by this patch.


Joern-R:
Thanks - makes sense - was a bit restrictive and not taken care on non-powermeter users.

PR #2524, opened 2017-05-20
Calculate Gear Ratio serios from XData

Joern-R:
... if "GEARS" are available as a XData series

### DIFF
--- a/src/FileIO/RideFile.cpp
+++ b/src/FileIO/RideFile.cpp
@@ -2479,35 +2479,49 @@ RideFile::recalculateDerivedSeries(bool force)
             }
         }
 
-        // can we derive gear ratio ?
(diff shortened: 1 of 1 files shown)


Joern-R:
@liversedge - can you please have a look - I am not sure whether the calcuation should take place in the "dependent" data series code - in the main code (like the HRV XData calcs.) Thanks. Joern

liversedge:
Looks fine. We can reorg later if we need to. I'm not 100% convinced the way HRV data is being processed is correct.

Joern-R:
Ok - thanks - at least all Gears code is in one place. I tested it - with GEARS from DI2 we now get accurate values - very nice. I will work on some DI2 stats charts in R - my first encounter with that technique.

Issue #356, opened 2012-11-20
[Old 724] Error importing Sigma 2.1 files


gcoco:
GC is unable to read Sigma 2.1 SLF/SMF files.
I suspect the file structure has changed with Sigma version 2.1.

PR #2554, opened 2017-06-29
Crash at *.slf import

fliicek:
When I try to upload *.slf file exported from Sigma Data Center 5.2.8 it asks me to confirm the date and time and after the confirmation the whole software "stops working" and crashes. I was able to import *.fit file, however, I would rather use *.slf file which is Sigma primary.


amtriathlon:
I'm closing this because Pull Requests are intended to contribute code, to report problems please use "Issues", there is one related to what you are reporting (https://github.com/GoldenCheetah/GoldenCheetah/issues/356) but I don't know if it is the same, also a sample file provoking the error is useful in theses cases.

PR #2562, opened 2017-07-17
TCX and FIT parser improvements

dhewg:
These are all improvements for files I had from multiple devices:
* TCX files exported from Polar Flow using a V800
* FIT files from a Wahoo ELEMNT BOLT


liversedge:
Thanks !

I need to add you to the contributors list, what is your name ?

dhewg:
Neat, thanks!

And it's "Andre Heider", as the "Author" on the patches

PR #2565, opened 2017-07-18
Fix MSVC Compile Error

Joern-R:
... default settings in MSVC2015 do not support "not" operator - so change to "!"
... since this is only a single occurance in overall code


liversedge:
Apologies for letting that thru!

Joern-R:
No problem at all - I am pretty busy currently - so no recent content contributions.

PR #2567, opened 2017-07-19
Changed completion in formulas to replace the whole word

### LINKED ISSUE #2566: Completion in formulas is confusing due to case insensitive matching when syntax is case sensitive
For example starting to write average and selecting Average_Power from the completion popup gives average_Power which is marked as invalid, this may confuse users as reported at the forum: https://groups.google.com/d/msg/golden-cheetah-users/D6GOzAIjVjw/49ppj1a0BgAJ
I think would better to replace the whole word with the selected completion.

amtriathlon:
To avoid problems due to case insensitive prefix match
Fixes #2566


rgeorge1985:
Thank you.

Issue #2568, opened 2017-07-20
HRV Measures


amtriathlon:
I would like to discuss how to add HRV Measures for 3.5, what follows is what I plan to do if there are no objections, better ideas or others working on this:
1) Introduce HRV Measures is a similar way to Body Measures, day/time record with: HR,AVNN,SDNN,rMSSD,pNN50,LF,HF and LFHF. Cached on Athlete, API to get/set.
2) Framework for Download copied from BodyMeasures, no online services yet.
3) Import from CSV file in HRV4Training format with unsupported columns and empty rows removed, supported columns are: timestamp_measurement, HR, AVNN, SDNN, rMSSD, pNN50, LF, HF, HRV4T_Recovery_Points. timestamp_measurement and rMSSD are mandatory.
4) Add pseudo metrics to access HRV measures
5) HRV Metrics integration: add an option to FilterHRV dataprocessor to set HRV Measures from computed HRV Metrics
6) UI to edit HRV Measures in Options/Preferences
7) Start a wiki page with how to use information:
 (https://github.com/GoldenCheetah/GoldenCheetah/wiki/HRV)
8) Upload a chart with daily and rolling averages values for the most commonly used measures. Uploaded to the cloud.

All items above are implemented in master and available on  [v3.5-DEV1708](https://github.com/GoldenCheetah/GoldenCheetah/releases/tag/v3.5-DEV1708).
For other features which also apply to BodyMeasures s.t. plot by days in LTM charts (ala PMC) and access from formulas, R and REST interfaces probably would be better to open other FRs.

liversedge:
Do we need to treat measures like we do pmc/stress calculations, when you request them a series is created for every day (not ride) ?

Additionally, RideItem::getMeasure(..,) like RideItem::getWeight() ?

The rest sounds perfect.

amtriathlon:
>Additionally, RideItem::getMeasure(..,) like RideItem::getWeight() ?
Yes, I will include this in part 1.

> Do we need to treat measures like we do pmc/stress calculations, when you request them a series is created for every day (not ride) ?
I was thinking more in the line of Athlete Weight, the pseudo metrics of part 4 use getMeasure to obtain the HRV measurement closer to the date of the ride, mostly for implementation simplicity.

liversedge:
Sure, I was thinking about plots with daily points, not just on the day of a ride like with PMC.

amtriathlon:
[...]

PR #2575, opened 2017-07-26
HRV Measures

amtriathlon:
Implements #2568 parts 1, 2, 3, 4 and 6.
CSV file format is the one generated by HRV4Training once removed unsupported columns and empty rows.


liversedge:
I am struggling for time at the moment, lots going on, so you should merge and we can work out fixups later. 

I scanned through and it looks terrific. Really quite excited about having a more rounded solution for tracking and managing performance and recovery. Will be adding a ROF widget myself shortly (if I get the time).

liversedge:
One small suggestion; might be worth adding an HRV4Training file to the repo so we can test.

amtriathlon:
Thanks Mark, I added a format document and a sample file for testing purposes.

I will merge but something I would like to review for Hrv and Body Measures is the cache refresh involving pseudo metrics s.t. Ahtlete Weight / Rest rMSSD

PR #2587, opened 2017-08-02
HRV Measures - Part 3

amtriathlon:
Added HRV fingerprint to RideItem to trigger refresh when
HRV Measures change.
Part of #2568

### DIFF
--- a/src/Cloud/HrvMeasures.h
+++ b/src/Cloud/HrvMeasures.h
@@ -44,8 +44,8 @@ public:
 
     double  hr,             // Average heart rate
             avnn,           // Average of all NN intervals
-			sdnn,           // Standard Deviation of all NN intervals
-			rmssd,          // Square root of the mean of the squares of differences between adjacent NN intervals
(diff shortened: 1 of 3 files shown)


rgeorge1985:
Ale,

I saw that the new Dev build is out, but is there an exe to install? I've never built from source nor have experience.

amtriathlon:
Binaries for Windows, Mac and Linux will be attached there when available, in the next days, just like all the previous dev builds.

rgeorge1985:
Thank you. Downloaded it. Should I uninstall dev v1705 first? With the
v1708, I'm not finding anything new in the interface. Can I assume it's
conflicting with the older version?

Also, importing polar r-r native files with the Get Measurements HRV isn't
working as it's not locating the file saved in the directory. The file is a
text file, coverting to .csv returns a header error.

amtriathlon:
Please see the wiki and use the forum for usage questions, thanks.

rgeorge1985:
Ok. Please see forum.

-Ronnie

PR #2602, opened 2017-08-20
More flexible HRV CSV file parsing

amtriathlon:
-Allows lines ending in CR as generated by HRV4Training iPhone app.
-Allows ";" additionally to "," as separator.
-Skip unsupported columns.
-Skip empty lines and values.
-Recognize the alternative names generated by HRV4Training when
orthostatic test is selected.


liversedge:
NICE!

PR #2607, opened 2017-08-24
Fix W' col resize bug in Power Zones settings

riccioclista:
The line
`ranges->setColumnWidth(3, (ranges->columnWidth(3)/2) *dpiXFactor);`
causes the 4th column to shrink a bit more each time the FTP vs CP combo box value changes.

Hiding the FTP column (instead of not adding it to the widget) relieves us of the duty to resize the col manually.


riccioclista:
Closed, because it's a mess.

PR #2619, opened 2017-09-10
Compile with gcc5.4 (Ubuntu 16.04) needs c++14, --> Sorry, Fork broken, please delete this pull request, will make new one...

Walibu:
Setting compilerflag c++14 for gcc-5.4.0 (Ubuntu16.04), according: https://gcc.gnu.org/projects/cxx-status.html
"Nested namespace definitions" are supported with gcc-6.x

Without this flag I get errors like this:
Cloud/CloudService.h:587:65: error: ‘CloudService::CloudServiceSetting’ is not a class or namespace
                 value = returning->settings.value(CloudService::CloudServiceSetting::DefaultURL, "");

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -139,8 +139,8 @@ win32-msvc* {
         # Linux Flex compiler grumbles about unsigned comparisons
         QMAKE_CXXFLAGS += -Wno-sign-compare
 
-        # Ubuntu 16.04 with gcc-5.4.0, it does not compile, unless c++ 2014 is set
-        CONFIG += c++14
+        # Ubuntu 16.04 with gcc-5.4.0, it does not compile, unless c++ 2011 is set
+        CONFIG += c++11
     }
 }
 


Walibu:
I was not able to compile GC on Ubuntu 16.04 LTS
I found this solution with the compilerflag setting to c++14 and it works.
Am I alone in having this issue, last time I could compile it was in March 2017

stschake:
Notice that setting this flag will cause builds to likely no longer work on 14.04 since you then need a newer version of the runtime.

Walibu:
Ok, I just did a full compile with c++11 instead of c++14, which shoud be compatible with Ubuntu 14.04 (Gcc 4.8) and it runs through.
As far as I know: with the default Gcc of Ubuntu 14.04 I am sure it does not compile without the c++11 switch, because since April 2017 there are enum Namespaces in the code.

I will change the patch to c++11 instead of c++14

Walibu:
Sorry!
My fork is broken, please delete this pull request, will make a new one

amtriathlon:
Ok, I will close it. Have you seen this: https://groups.google.com/forum/#!topic/golden-cheetah-developers/IG5whIx53rc?

liversedge:
Its also worth noting that we don't really support building with QT < 5.6 anymore, so with a later version of QT c++11 will be automatically added.

We can't add it to the default setup for QT < 5.6 since its not necessarily supported by the toolchain or QT.

PR #2631, opened 2017-09-26
Detect Garmin Express blocking ANT stick

stschake:
Trainer season is coming up and I've learned that Garmin Express (certainly on Windows) has a new trick up its sleeve: it registers an always on-in-the-background service that will block the ANT stick. This is separate from the Garmin Express app that has the tray icon and no longer starts automatically.

So now on Windows to use ANT+ in GC, you have to go into the services manager, find the Garmin Device Interaction Service and stop it manually. That is if you know to do that :)

This PR checks if the Garmin Express service is running before connecting to the ANT+ stick and offers to stop the service temporarily, for both Windows and MacOS (where at least you still have an icon in the menu bar). It also removes the fixed 500 ms delay and fails faster when we couldn't open the ANT+ stick.

### DIFF
--- a/src/ANT/ANT.cpp
+++ b/src/ANT/ANT.cpp
@@ -487,6 +487,12 @@ ANT::start()
 int
 ANT::setup()
 {
+    if (channels == 0)
+    {
+        // we obviously have not managed to open an ANT stick, fail fast
+        return 0;
+    }
+
     uint8_t attempts = 0;
     do
     {
--- a/src/ANT/ANT.cpp
+++ b/src/ANT/ANT.cpp
@@ -85,7 +85,7 @@ const ant_sensor_type_t ANT::ant_sensor_types[] = {
 // thread and is part of the GC architecture NOT related to the
 // hardware controller.
 //
-ANT::ANT(QObject *parent, DeviceConfiguration *devConf, QString athlete) : QThread(parent), devConf(devConf)
+ANT::ANT(QObject *parent, DeviceConfiguration *devConf, QString athlete) : QThread(parent), devConf(devConf), portInitDone(1)
 {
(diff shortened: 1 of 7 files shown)


stschake:
Rebased after NPE sillyness.

Joern-R:
@stschake - is this one still up-to-date ? if yes, is there anything specific to test it (besides installing Garmin Express?) - Thanks Joern

stschake:
GitHub says it still merges cleanly so yes, it should be fine. The AppVeyor failure was a timeout IIRC, not related to the code.

For testing, installing Garmin Express & rebooting (so it has a chance to start its service) should be sufficient. On MacOS, the tray icon should be there.

Joern-R:
Just did a first test - locally - I get the message that Garmin Express is running - and confirm to stop it locally - then after > 20 seconds get the popup 
image

System is Windows 10 - 64Bit - Pro - latest official version 1709. Any suggestions ?

Maybe one speciality - I have an Garmin Index Scale - which also has an ANT+ connectin. If I start Express it shows the scale as "connected".

stschake:
20 seconds sounds suspiciously like it timed out stopping the service. That usually happens when the service doesn't terminate cleanly (i.e. like kill you can catch the signal, then do nothing).

Could you test starting the Garmin service through services.msc, stop it in GC, then if it has the same error, check in services.msc (might have to press f5 to refresh) if it was stopped?

Joern-R:
I tested again (with your additional info in mind.) Windows worked perfectly fine. Stopping - starting when leaving GoldenCheetah. 

@liversedge - I cannot test on Mac on short notice. If you agree I suggest to merge.

Thanks Stefan - the feature will definitely help a lot people.

liversedge:
Sure. Windows is the main one anyway.

PR #2645, opened 2017-10-04
Add option to caculate new speed and distance using FixGPS

leifwar:
I sometimes have odd gps measurements that are causing extreme speed values. By clearing (Ctrl - 0) the Latitude and Longitude for any marked segment I can interpolate new gps coordinates using FixGPS.The speed values are still wrong, but by using the added option in this commit it will updated both speed and distance accordingly.


liversedge:
I never really understood this one .. why did you close it ?

leifwar:
I need to fix a bug in the distance calculation and will reopen after that. 

As you can see from the two attached figures I sometimes get wrong GPS data that can cause extreme speed values. The idea with this code is to update both speed and distance for the rows containing wrong latitude and longitude only.

Before: 
fix_gps_1

After:
fix_gps_2

PR #2646, opened 2017-10-06
Fix Strava temperature series import

stschake:
This was mangling the temp series name. With this commit, Strava requests will also include the temp series.

The API docs indicate this parameter only has an effect when also specifying resolution, and in any case it needs to be sent as POST data - not appended to the URL.

See also https://strava.github.io/api/v3/streams/#activity


liversedge:
Weird bit of code !

time was already on the list of series to retrieve, suspect this "broke" the url request that has temp at the end of the list.

Thanks for the fix.

PR #2647, opened 2017-10-06
Use logarithmic scale for speed chart, like CS

Walibu:
Set scale for speed like CS to logarithmic. Same scale as CP
It has been changed with pull request #2004, but I am not happy about it.
Running workouts with intervals 200m to 1km are easier to interpret..... and it is the same as CP

Sorry, I already comited the c++11 change to my fork and could not remove it from that pull request, so it comes with this as well.

See my two CS plots, one with logarithmic and the other with linear scale.
The new GPS tracking watches are quite reliable with speed. For e.g. I use a 4 year old Suunto Ambit 2 and the logarithmic CS plot is quite useful for comparing short running intervals up to 3 min.
cs_logarithmic

cs_linear


amtriathlon:
I stand my original position but, if you make the scale log/linear an option for the user, I will be glad to apply the change.

Walibu:
Ok, hope I had some more time left for coding and getting used to QT

PR #2656, opened 2017-10-18
Fixed malformed JSON file due to empty metric overrides

amtriathlon:
When the user check/uncheck a metric override checkbox
on RideMetadata an empty metric override is generated,
the line removed on JsonRideFile.y produces an extra ","
with errors reported at the forum

https://groups.google.com/d/msg/golden-cheetah-users/CiwAwufvQRA/ACHM0hhxAwAJ and https://groups.google.com/d/msg/golden-cheetah-users/Tpt0cUZBAbA/dQkz4R3EAwAJ

This is the minimum change to fix the issue, alternatively we could avoid the generation of an empty override on JsonRideFile.y  or at the source on RideMetadata.cpp


liversedge:
i think thats good. I should also fix RideMetadata leaving empty ones.

amtriathlon:
Mark, the added commit prevents the addition of empty overrides by RideMetadata, let me know if you’re ok with this change.

liversedge:
looks good, thank you.

Issue #2669, opened 2017-10-26
Problem with Date and Time while importing ZWIFT Files into Golden Cheetah


mtbkeith:
Importing from ZWIFT files causes dates to show up prior to 1989-12-31 00:00:00.

This was reported by Frank Wechsel on the Google group for GoldenCheetah. 

![grab-gc-user-bildschirmfoto 2017-10-24 um 14 26 10](https://user-images.githubusercontent.com/9326275/32036082-b40535e8-b9da-11e7-9397-3edd90766deb.jpg)

Thanks to TJDunnigan for attaching some files for investigation.

mtbkeith:
Opened pull request to merge fix:
https://github.com/GoldenCheetah/GoldenCheetah/pull/2670

mtbkeith:
Attached you can see the outcome of parsing a file which was duplicated and name obfuscated; they parse to nearly the same time; lacking the timezone information which was left out of the Zwift file. 
 
<img width="922" alt="grab-gc-zwift-date-issue-my-fix" src="https://user-images.githubusercontent.com/9326275/32036480-5114c540-b9dd-11e7-96c4-7b7890dc3a9c.png">

Joern-R:
PR is merged - so I consider this done. Thanks.

PR #2670, opened 2017-10-26
Fixes ZWIFT files which are not reporting local timestamp.

mtbkeith:
Secondary fix for regular expression which did not catch the ZWIFT file naming convention using dashes rather than underscores.

Initial Bug report:
https://groups.google.com/forum/#!topic/golden-cheetah-users/9JXU9Z5vMrE
Tracked as:
https://github.com/GoldenCheetah/GoldenCheetah/issues/2669


liversedge:
PERFECT!

PR #2672, opened 2017-10-28
updated INSTALL-LINUX

cdknight:
A couple of weeks ago, I tried building GoldenCheetah on Linux, and found trouble as my distribution was struggling to build it. The dependencies were missing. I furthermore noticed the "INSTALL-LINUX" file had not been updated in three years. 

I updated INSTALL-LINUX with Arch linux instructions, removed some old version numbers, like Ubuntu 11.04 on amd-64, and slightly cleaned this up. After further research, I may contribute to this file more. My commit will have to be checked for accuracy and installation success later on.


liversedge:
thanks.

Issue #2534, opened 2017-05-27
CP history


henolo:
v3.5-DEV1705

Some curves have messed up x-axis values for the newer values (these dates are shown as older than they should).
This depends on the time period.
Longer periods have "compressed" new values.
Shorter periods are ok.
"Latest 28 days" and shorter are shown ok, "Latest 2 months" and longer are not.
![latest 28 days](https://cloud.githubusercontent.com/assets/15927134/26523320/8ae6ffd4-4315-11e7-924b-ead14c9096e9.jpg)
![latest 2 months](https://cloud.githubusercontent.com/assets/15927134/26523323/93e568f0-4315-11e7-942a-30495cb30e1a.jpg)

riccioclista:
I can confirm that one for current master (167165ca15724d88b7a796e40d4494a14216996f).

henolo:
The error behavior in my case has changed since I initially reported it. Now (Oct 25) all Date Ranges but one are ok. "Last 12 Months" is still plotted with wrong dates. Could be that some data (activities or analysis of activities) are not "shaped" as this plot needs them to be.

riccioclista:
I've debugged into the code and found that the plotting code doesn't account for gaps in the underlying model data. For example, if you have model estimates for January and May, it plots the May data right next to the January data without leaving the proper gap. I'll try to make the code more rebust in that regard.

henolo:
I can confirm that there are larger gaps during the winter (2016/2017) in my case - fits well with the very messed up 12 month view. And also some shorter gaps more recently due to various "issues" like inspecting the tarmac and germs and yes they are ideed "time-compressing" more or less all the durations to the left. Didn't observe this until now.

henolo:
Closing!

PR #2673, opened 2017-10-28
Handle gaps in estimate data in LTMPlot #2534

riccioclista:
This PR makes the plotting of model estimate data more robust against gaps in the estimate data. Normally estimate data (CP, W' Prime, ...) is stored sequentially without gaps (each point in time from a defined start to a defined end date has a model estimate). However, if this is for some reason (bug?) not the case, LTMPlot.cpp would still plot the data without leaving the proper gap (See #2534). This PR inserts 0s in LTMPlot for time spans without estimate data.

### DIFF
--- a/src/Charts/LTMPlot.cpp
+++ b/src/Charts/LTMPlot.cpp
@@ -45,6 +45,7 @@
 #include <qwt_legend.h>
 #include <qwt_legend_label.h>
 #include <qwt_plot_curve.h>
+#include <qwt_plot_gapped_curve.h>
 #include <qwt_plot_canvas.h>
 #include <qwt_curve_fitter.h>
 #include <qwt_plot_grid.h>
@@ -437,7 +438,9 @@ LTMPlot::setData(LTMSettings *set)
         if (count <= 0) continue;
 
         // Create a curve
-        QwtPlotCurve *current = new QwtPlotCurve(metricDetail.uname);
+        QwtPlotCurve *current = metricDetail.type == METRIC_ESTIMATE
+                ? new QwtPlotGappedCurve(metricDetail.uname, 1)
+                : new QwtPlotCurve(metricDetail.uname);
         current->setVisible(!metricDetail.hidden);
         settings->metrics[m].curve = current;
         if (metricDetail.type == METRIC_BEST || metricDetail.type == METRIC_STRESS)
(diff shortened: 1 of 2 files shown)


liversedge:
What is the problem this is trying to fix?

Gaps in data are perfectly valid -- the plot should also contain gaps.

Is there an issue when aggregating (by week/month etc) ?

riccioclista:
The problem are gaps in the power model estimate data (`context->athlete->getPDEstimates()`). Model estimate entries in this list have a start and end date. For LTMPlot to work correctly, this list must not contain any time gaps, because LTMPlot ignores them and just plots all data in sequence. This can be observed on the "Trends/CP History" chart with by-day resolution.

riccioclista:
Here is the CP history chart with current master:
screenshot from 2017-10-28 15-30-21
and here with fix:
screenshot from 2017-10-28 15-27-47

In the fixed version, you can see that gaps exist and are accounted for.

liversedge:
Not sure I prefer the fixed version, and need to check how data is aggregated when show month / year...

There is a special plot curve called QwtPlotGappedCurve that may also be useful to show gaps (instead of dropping to zero) perhaps that is a better compromise ?

riccioclista:
QwtPlotGappedCurve might be a better solution, I'll look into it.

liversedge:
Cool, you will see it is used to plot power in AllPlot.cpp, if you want an example of it in use.
CHEERS, Mark.

riccioclista:
The CP history chart now looks like this:
screenshot from 2017-10-29 10-53-12

liversedge:
Excellent, have you checked it works correctly (and calculates correctly) when you show by week/month/year ? (i.e. aggregating)

liversedge:
Also, I noticed you changed qwtplotgappedcurve, please remove that change. It is used in other places.

riccioclista:
I've removed the QwtPlotGappedCurve change. However, there are 2 problems that come with the switch to QwtPlotGappedCurve:

* QwtPlotGappedCurve omits the last element in the data series. The upper (exclusive) bound is determind like so: `to = dataSize() - 1;`. This can be best observed when aggregating by month:
screenshot from 2017-10-31 11-34-46

* QwtPlotGappedCurve connects the last data point of a gap with the first data point after the gap resulting in a very steep line:
screenshot from 2017-10-31 10-33-36

Other than that, I've not observed any problems.

liversedge:
Thanks ! We can fix the fencepost error separately.
Can you confirm the aggregate is not affected by zero values?

If so, I'll merge, thank you for being patient with me :)

riccioclista:
No, it's not good to merge yet. You're right, aggregates might not get calculated correctly.

riccioclista:
I've fixed it now for all aggregate options. There were 2 issues in the previous version that I've fixed now:
* Before performing a gap fill-up in an aggregate chart, the last item on the aggregate stack (yTotal/xCount) is flushed, i.e. the avg is calculated and written to the values array `y`.
* The calculation of the previously processed estimate should only use the estimate end date if the chart resolution is by-day. For any other chart aggregate type the start date should be used, because the end date might already be in a subsequent aggregate group.

I'm sorry, I didn't spot the issues earlier. I've now tested all aggregate options thoroughly by comparing the calculated values with the model values in a [spreadsheet](https://github.com/GoldenCheetah/GoldenCheetah/files/1434205/testdata.xlsx).

liversedge:
Excellent. Will merge then we can test via the development builds.

PR #2677, opened 2017-11-01
QwtPlotGappedCurve: Don't draw a steep line after a gap

riccioclista:
This PR addresses 2 issues with QwtPlotGappedCurve:

1. Leading steep lines after gap:
QwtPlotGappedCurve draws a steep line after each gap that connects the last point of the gap with the first point of the next segment. This line is now excluded from the plot.
steep_lines_compare

2. QwtPlotGappedCurve constructs the plot by drawing 2 consecutive points each iteration, which results in a not-so-smooth line being drawn. This PR changes the behavior, so that the whole segment between two gaps is drawn in one go. This presumably has also performance benefits, because the underlying `drawSeries` method is not invoked each iteration, but only when a segment is complete. Here is a comparison of the line smoothness between old and new code:
smooth_line_compare

Additional notes:
QwtPlotGappedCurve is used in LTMPlot.cpp and AllPlot.cpp. The fix improves the appearance of LTMPlot, but doesn't have much of an impact on AllPlot:
qwtplotgappedcurvecompare

### DIFF
--- a/qtsolutions/qwtcurve/qwt_plot_gapped_curve.cpp
+++ b/qtsolutions/qwtcurve/qwt_plot_gapped_curve.cpp
@@ -33,6 +33,8 @@ void QwtPlotGappedCurve::drawSeries(QPainter *painter, const QwtScaleMap &xMap,
 
     int i = from;
     double last = 0;
+    bool isGap = true;
+    int startPoint = to;
     while (i < to)
     {
         // First non-missed point will be the start of curve section.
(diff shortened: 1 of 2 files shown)


amtriathlon:
I am closing this because I cannot reproduce the original issue looking at a measures curve with gaps, no steep line is draw after a gap in current master.

image

After applying this PR the point following the gap is not draw, perhaps this is related to the change in QWT version.

Issue #2588, opened 2017-08-03
Access to Body and HRV Measures by days


amtriathlon:
I should be possible to plot Measures (Body/HRV) by days in LTM charts (ala PMC) and access them from formulas, R and REST interfaces independently of activities.
The ability to select some of them to be shown in the Calendar would be a nice addition.
Part 1 - LTM Chart
Part 2 - Formulas
Part 3 - R Chart
Part 4 - REST API
Part 5 - Refactor Measures

PR #2679, opened 2017-11-02
LTMPlot - Access to Measures by days

amtriathlon:
Added Daily Measure to curve config, settings and plot
Added API for common access to Body/Hrv Measures.
Part 1 of #2588 
NB: I plan to refactor Measures code after the functional part is complete.


liversedge:
Having a look now.

amtriathlon:
Thanks Mark, I agree, there is a lot of duplication between Body and Hrv measures which could be avoided by subclassing and it would make it easier to introduce other daily measures in the future, I plan to fix it.

PR #2682, opened 2017-11-03
Access to Measure by days in Formulas

amtriathlon:
Eg. measure(Date-1, "Hrv", "RMSSD") to get the RMSSD
measure from Hrv group the day before the activity date.
Part 2 of #2588

### DIFF
--- a/doc/user/formula-syntax.txt
+++ b/doc/user/formula-syntax.txt
@@ -81,6 +81,9 @@ DATA PROCESSOR FUNCTIONS
 autoprocess(filter)
 postprocess(processor, filter)
 
+DAILY MEASURES ACCESS
+measure(date, "group", "field")
+
 RIDE SAMPLE DATA
 
 SECS
--- a/src/Core/Athlete.cpp
+++ b/src/Core/Athlete.cpp
@@ -634,21 +634,21 @@ Athlete::getHrvMeasure(QDate date, int field, bool useMetricUnits)
 
 // Common access to Measures
 QStringList
-Athlete::getMeasureGroupSymbols() const
+Athlete::getMeasureGroupSymbols()
 {
     const QStringList groups = QStringList() << "Body" << "Hrv";
     return groups;
 }
 
 QStringList
-Athlete::getMeasureGroupNames() const
+Athlete::getMeasureGroupNames()
 {
     const QStringList groups = QStringList() << tr("Body") << tr("Hrv");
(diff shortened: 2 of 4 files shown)


liversedge:
Small confession -- the autocompleter code for datafilters needs to be updated independently (blush).

To find where used I fgrep for const(pi)... and it found:
```
markl@monolith:~/Projects/Live/GoldenCheetah/src$ fgrep "const(pi)" */*.cpp
Charts/LTMTool.cpp:    list << "const(pi)";
Core/UserData.cpp:    list << "const(pi)";
Gui/EditUserMetricDialog.cpp:    list << "const(pi)";
Gui/SearchBox.cpp:    list << "const(pi)";
```

liversedge:
I'm not proud of this code.

amtriathlon:
Hi Mark, I was in doubt about to use symbols for measure groups/fields and decided for strings using XDATA as model, instead of const(ss) or config(ss), but I could change this since group/fields are known and we could use simbols, currently the autocompleter show this:

screenshot from 2017-11-04 09-38-28

which is included in all that cases via the call to builtins already present:
ale@hppavilion:~/Projects/Goldencheetah/src$ fgrep "DataFilter::builtins" */*.cpp
Charts/LTMTool.cpp:    list = DataFilter::builtins();
Core/DataFilter.cpp:DataFilter::builtins()
Core/UserData.cpp:    list = DataFilter::builtins();
Gui/EditUserMetricDialog.cpp:    list = DataFilter::builtins();
Gui/SearchBox.cpp:    list = DataFilter::builtins();

I could add all groupXfield combinations to the autocompleter(s) but I think the syntax should change to put the data parameter at the end for this to work, like measure(HRV, RMSSD, data).

What do you think?

amtriathlon:
Ok, I've decide to continue with strings to avoid the symbol space over-population but I've enhanced the autocompleter for the measure builtin so the user doesn't need to RTFM to know which the options are.
Also add a safety check for type=String before to access the value union.

liversedge:
I like your solution actually - quick Q, does it work the same in different languages? i

Apologies for delay in responding.

amtriathlon:
In formulas I'm using group/field symbols which are in english and not translatable, and the same I plan to do with R access, so the scripts should work independent of the language.
For UI elements s.t. LTMPlot and Pages the idea is to use group/field names which are the translatable version of the symbols.

PR #2713, opened 2017-12-03
Fixes parsing Zwift Workouts

Joern-R:
... support "Power" Tag - in cases where no Low/High Power is given
... support Duration provided as Float number (not only Integer)


liversedge:
Ah excellent -- I only had a single example to work from and not sure I got it right.

PR #2714, opened 2017-12-04
Fix for left and delete keypress events

sisao:
Prevent from moving into prompt :-)


liversedge:
thanks. you're watching closely ! :)

I'm hoping to get access to the activity data in next week/over the weekend.
I reckon if I create GC.window() to return a pointer to a QWidget in the view, then you might be able to use PyQt5 to create stuff ... but haven't tried it yet...

sisao:
Nice. I like the whole ml thing. I hope I can learn a bit about the ml tech stuff in connection with my own sports data... So not just doing tutorials with demo data etc...

PR #2721, opened 2017-12-09
Move cursor in PythonChart to correct start point

sisao:
Again one small layout fix. Home key press jumped to line start.

And I used http://doc.qt.io/qt-5/qtextcursor.html#positionInBlock


amtriathlon:
Thanks, I use the Home key a lot...

PR #2722, opened 2017-12-09
Use utf-8 for series names

sisao:
With german translation activated GC.activity() complains about some utf-8 decoding stuff and returns error.
First I thought came from the real activity json file, but it's from translation: Altitude -> Höhenmeter. Is it necessary to use toLatin1() in goldencheetah.sip? With toUtf8() it works and you get delta symbols instead of question marks.


liversedge:
The generated files had a merge conflict, which I kind of fixed by hand in the web editor but will close and apply by hand. Thanks for the fixup which applies for all QStrings.

The series names are likely to be translated, so this may change as that would cause a few issues down the line (charts would be language specific).

liversedge:
I just committed it by hand, to fixup UTF8, but also changed to use the same naming convention for series names that we use in R (no spaces, untranslated, follow R trackR package).

PR #2723, opened 2017-12-10
OSM - Tile Server Handling Changed

Joern-R:
... all Tile Servers are configurable by the user - there is only a default for the "standard" OSM Tile Server, but even this default can be overwritten
... there are no dedicated Tile Servers Sources named / hard-coded, but in addition to the default OSM server 3 additional options are available for the user to configure

Why:
... tile servers either change their URL or their T&C - which happened for some of the "hard-coded" once recently - since this may happen again in future the recent change de-couples the code from the Tile Server URL - and the availability of certain organisations providing free tile servers

### DIFF
--- a/src/Charts/RideMapWindow.cpp
+++ b/src/Charts/RideMapWindow.cpp
@@ -80,37 +80,31 @@ RideMapWindow::RideMapWindow(Context *context, int mapType) : GcChartWindow(cont
     commonLayout->addRow(new QLabel(tr("Show Intervals Overlay")), showInt);
     commonLayout->addRow(new QLabel(""));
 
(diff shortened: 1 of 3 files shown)


gcoco:
Would this be something we could change to have a pick list from the cloud? That way we can update quickly if something changes without changing code?

Joern-R:
Yes - that would be a cloud feature. But honestly - I do not really want to maintain such configuration over time. When searching for tile servers, even the OSM Wikis are partly outdated - so hard to find a proper list. On the other hand I assume that 99 % of the user may not even care on different maps. Want I wanted to achieve with this change is that we have nothing hard-coded in GC. Now OSM is fully configurable. 

For a commercial offering putting such configuration into either an automated upgrade process or into a cloud configuration service would be my choice. But that requires a team keeping up with changes and operate such service.

PR #2725, opened 2017-12-12
Python activityMetrics

amtriathlon:
Returns a Python dict with metrics and metadata fields.
It doesn't implement compare mode yet.


src/Python/SIP/Bindings.cpp
```diff
@@ -92,6 +90,66 @@ PythonDataSeries::~PythonDataSeries()
...
+        const RideMetric *metric = factory.rideMetric(symbol);
+        QString name = context->specialFields.internalName(factory.rideMetric(symbol)->name());
+        name = name.replace(" ","_");
+        name = name.replace("'","_");
+
+        bool useMetricUnits = context->athlete->useMetricUnits;
+        double value = item->metrics()[i] * (useMetricUnits ? 1.0f : metric->conversion()) + (useMetricUnits ? 0.0f : metric->conversionSum());
+
+        // add to the dict
+        PyDict_SetItemString(dict, name.toLatin1().constData(),
```
liversedge: toUTF8 ?

src/Python/SIP/goldencheetah.sip
```diff
@@ -122,6 +122,9 @@ public:
     int seriesLast() const;
     PythonDataSeries series(int type=10) /TransferBack/;
 
+    // working with metrics
+    PyObject* activityMetrics() /TransferBack/;
```
liversedge: how does the dict get deleted? (does the python gc do it?)

liversedge:
I added a couple of queries. But itching to merge it !

src/Python/SIP/goldencheetah.sip
amtriathlon: Yes, that's the idea of the /TransferBack/ annotation, I tested it with the series method above and the PythonDataSeries destructor is called by Python GC when there are no more active refereces, it's not so easy to test for dicts, but I assume the same behavior.

liversedge: Ah ok - so the PyObject destructor is called. Cool.

src/Python/SIP/Bindings.cpp
amtriathlon: That's was copied from RTool but you are right, even when there is no problem with metric internal name, metadata fields and contents may give problems. I will change all conversions to Utf8.
BTW, I'm not that convinced to return value according to useMetricUnits in the API, as implemented it doesn't handle pace units, and it may be better to stick to SI for the API and let the conversions to user's code.

liversedge: Your code is held to a higher standard than mine then !!! :)

PR #2733, opened 2017-12-21
Remove unnecessary linking against libvlccore

sebastinas:
libvlccore is for plugins, but GoldenCheetah does not build a plugin. It
requires only functions available from libvlc.

Signed-off-by: Sebastian Ramacher <sramacher@debian.org>


liversedge:
Thanks .. that's a long standing mistake!

PR #2738, opened 2017-12-29
2nd version with option to choose: logarithmic/linear scale for speed chart, like CS #2647

Walibu:
Added options in to configure "linear" or "logarithmic" scale of the critical speed chart.
According amtriathlon:
"I stand my original position but, if you make the scale log/linear an option for the user, I will be glad to apply the change." see #2647

It was mainly straight forward, after I got used to Qt again.

The minor change for compiling on Ubuntu 16.04 with gcc-5.4.0 or older comes as well, its the c++11 in src.pro

### DIFF
--- a/src/Charts/CPPlot.cpp
+++ b/src/Charts/CPPlot.cpp
@@ -64,6 +64,7 @@ CPPlot::CPPlot(QWidget *parent, Context *context, bool rangemode) : QwtPlot(pare
     shadeIntervals(true), rangemode(rangemode), 
     showBest(true), filterBest(false), showPercent(false), showHeat(false), showHeatByDate(false), showDelta(false), showDeltaPercent(false),
     plotType(0),
+    xAxisLinearOnSpeed(true),
 
     // curves and plot objects
     rideCurve(NULL), modelCurve(NULL), effortCurve(NULL), heatCurve(NULL), heatAgeCurve(NULL), pdModel(NULL)
@@ -235,7 +236,11 @@ CPPlot::setSeries(CriticalPowerWindow::CriticalSeriesType criticalSeries)
     case CriticalPowerWindow::kph:
         series = tr("Speed");
         units = tr("kph");
-        scale = linear;
+        if (xAxisLinearOnSpeed == true) {
+            scale = linear;
+        } else {
+            scale = log;
+        }
         break;
 
     case CriticalPowerWindow::nm:
@@ -1411,8 +1416,9 @@ CPPlot::plotBests(RideItem *rideItem)
 
     // now set the scale
     QwtScaleDiv div((double)xmin, (double)xmax);
+
     if (criticalSeries == CriticalPowerWindow::work ||
-        criticalSeries == CriticalPowerWindow::kph)
+        (criticalSeries == CriticalPowerWindow::kph && xAxisLinearOnSpeed))
         div.setTicks(QwtScaleDiv::MajorTick, LogTimeScaleDraw::ticksEnergy);
     else if (criticalSeries == CriticalPowerWindow::veloclinicplot)
         div.setTicks(QwtScaleDiv::MajorTick, LogTimeScaleDraw::ticksVeloclinic);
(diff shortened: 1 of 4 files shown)


liversedge:
why is src.pro changed?

Walibu:
I had problems building with gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.5) and QT 5.5.1
I got errors like this:
Cloud/CloudService.h:590:65: error: ‘CloudService::CloudServiceSetting’ is not a class or namespace
                 value = returning->settings.value(CloudService::CloudServiceSetting::DefaultURL, "");
After asking google, I found this:
https://gcc.gnu.org/projects/cxx-status.html

Sorry, you can remove this change src.pro, then a add it in my gcconfig.pri
Right now, I try a full build without this flag, but a week ago a could not build.

liversedge:
You need to build with 5.8 or higher really (GC needs a fair few of the bug fixes), since c++11 support only arrived in QT 5.6 you are way out of date. if you remove the src.pro updates I can merge.

Walibu:
Ok, thank you, does that work with the revert inside the pull request?

So, I have to update Qt in the next days.

amtriathlon:
Can you rebase to consolidate the changes on one commit?

Walibu:
Sorry, how i do that?
I am reading through https://help.github.com/articles/about-git-rebase/
but not sure what to do, what is the purpose, remove the commit with src.pro changes?

Walibu:
With the command:
$ git rebase -i upstream/master

I get the following lines:

pick a3304a3 Use logarithmic scale for speed chart, like CS
pick 35ef4a1 Ubuntu 16.04 or 14.04 with gcc-5.4.0 or older, does not compile, unless c++ 2011 is set
pick dd9ba9b Configurable scale for speed chart like CS, logarithmic or linear
pick c3907de Enable option for configurable time scale on Speed data series only
pick 41c6d77 Refactoring, extract method for updating options checkboxes
pick aa69aa0 Revert src.pro changes

and now?

mtbkeith:
you would do as follows from your command line;

GoldenCheetah # git checkout master
GoldenCheetah # git pull
GoldenCheetah # git checkout Your-Branch-Name
GoldenCheetah # git rebase -i master

# this now gives you a VI editor; which should show your two commits.
something like
pick 99949 38484 your first commit message
pick 94944 29394 your second commit message

change the second "pick" to an "f" for "fixup" (you need to do this in
insert mode by pressing 'i' and moving the cursor around).
exit the vi editor by exiting insert mode by hitting 'esc' and then write
the changes of the file ":wq"
then you do a push to your branch as follows;
GoldenCheetah # git push --force

This essentially does a merge/squash of the two commits. Now your two
commits will be put together as one, and the changes to the "src.pro" file
will not show up for everyone else in master.

Thanks,
Keith

mtbkeith:
Now that I see your output a slight change to my previous directions,
change all picks after the first one to 'f' for fixup. It's just cleaner to
merge/squash them all together.

Walibu:
So I change all "pick" to "f" except the first
Write it out and close the editor.
Hope this works...

Walibu:
Done, but now I can't see the changes and the refactoring as separate commits, should I change that if possible or leave it?

mtbkeith:
Walter,

You can see it in the history of your branch. It often does no one else any
good to see separate commits in "master";

ex:
change1: change log scale of drawing
change2: oops, change variable names
change3: code review changes

those should all be squashed down into one commit.

One thing I didn't mention is if there is a very active project you will
have to resolve conflicts during the rebase operation for the files which
others have changed. For your work, I assumed no one else was changing the
same files (a safe bet that shows to be true).

Keith

Walibu:
Thanks Keith

Lg Walter

amtriathlon:
I added a refresh after Linear scale option change and merged.
Thanks Walter!

Walibu:
Thanks Alejandro!

PR #2741, opened 2018-01-02
LTMTool: Set symbol for measure metrics

riccioclista:
When adding measures to an LTMPlot with other metrics, those measures
are not cleaned up properly on replot, because they don't have a symbol.
LTMPlot uses the symbol to store them in a 'curves' dictionary. All
metrics without a symbol are then identified by the default metric
symbol value, which is: "".


amtriathlon:
Thanks!

PR #2742, opened 2018-01-03
sv translation

henolo:
Various cleanup: naming in general, garmin cycling dynamics naming, spelling, and then some.


amtriathlon:
Thanks Henrik!

PR #2752, opened 2018-01-14
LTMSidebar: Fix crash when opening event popup if phase is selected

riccioclista:


amtriathlon:
How do you reproduce the issue?
I cannot reproduce it using current master, when I select a phase to create an event it is created in the corresponding season as expected.

amtriathlon:
Thanks, I included this case when fixed Edit and Delete events with phase selected since the 3 are similar.

PR #2758, opened 2018-01-20
attempt to fix travis build (changed qt 5.5.1 mirror download location)

sisao:
This will maybe fix the travis build with qt 5.5.1. The build is failing at the moment because the mirror doesn't provide qt 5.5.1 anymore. I replaced the the url with qt archive and saved the qt.rb formula in the travis folder. I can't test this, bacause I have no Mac :-)


liversedge:
thanks !

liversedge:
it looks like that didn't fix the issue though. probably better just to remove 5.5 from build tests since we don't support that anymore anyway.

Issue #2764, opened 2018-01-25
Withings API Broken - Since Jan 21 2018


arunh:
Withings API appears to be silently failing since Jan 21. Using a proxy to log responses I've established that valid data is being returned per the Withings API documentation. However, it does not seem to be being parsed and stored into the athlete record.

I don't have enough knowledge to update the Yacc parser code, but a look at the WithingsParser.l file suggests that in the response below that I logged there is a "timezone" field, whereas there is no corresponding field in the parser definition. Perhaps this is causing the issue?

```
{
	"status": 0,
	"body": {
		"updatetime": 1516908612,
		"timezone": "Europe\/London",
```

A couple of things I found out in WithingsDownload::parse

QStringList "errors" is empty - which is surprising as I expected it to contain the parser errors.
Function is seeing zero WithingsReading objects when it calls parser->readings().

Attached the full response I am currently getting from Withings showing valid data but not parsing.

[withings.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/1665751/withings.txt)

liversedge:
Oh god. I wish we'd had a JSON parser back then. Will fix urgently and get into next dev build.
Thanks for doing the legwork.

arunh:
@liversedge I've re-implemented the Withings API parse using the Qt Json Parser. It works for my local environment and the values check out against my known values on Withings and now have data in my athlete measurements right up to today. At your discretion if you choose to use it, but my hope was that it would be more robust to future un-versioned/un-announced API changes by Withings. In the case of the new timezone field it has proven to be the case (as you would expect) and it is duly ignored without error since we don't even try to read it. Pull request created.

liversedge:
NEED TO MAKE 4.8 BUILDS COMPILE.

liversedge:
I think it would be good to make it dependant on QT > 5.1 so 4.8 builds still work, they just have less functionality?

amtriathlon:
Ok the new Body/Hrv Measures functionalities and previous Withings download would not be available but it should work

drnovsek:
Hi all,
Not sure if relevant but will mention it for visibility.
Using 'dashboard.health.nokia.com' for weight data and works for me without issues.
GC ver: 3.5-dev1710

liversedge:
thanks, that's useful to know !

drnovsek:
[...]

PR #2765, opened 2018-01-26
Use Qt Json Parser for Withings API (Fixes Issue #2764)

arunh:
Fixes issue #2764


liversedge:
Brilliant -- thats great -- but did you update src.pro to reflect the source code name change?

liversedge:
And now it also needs to be dependant on Qt5 .. we informally support Qt 4.8 for older platforms.

arunh:
Hi Mark, I didn't see any references in src.pro relevant to the rename but apologies if I have missed something - it did seem to build/work fine on my workstation after a 'make clean'.

It's a good point about the Qt4 support, I didn't think of that. Your call on that one obviously - whatever you decide it was an interesting Friday afternoon activity!

liversedge:
Ok will merge and fixup for Qt4.

Issue #2767, opened 2018-01-28
Zwift "zwo" Export from Workout Editor


arunh:
It would be useful to be able to export Zwift "zwo" files from the Workout Editor so that workouts designed using the "scientific W' approach" can be easily executed in the Zwift environment.

Mark @liversedge - I'd like to have a go at implementing this, but don't want to go too far without your input. Presumably we could just integrate something like this by adding a "file type" drop down to the existing "Save As..." dialogue in the workout editor?

liversedge:
Go for it. I think it would be great to be able to write workouts for Zwift.
I still don't really understand why they invented a new format.

arunh:
Dug into this a bit and functionality already exists to write Zwift XML but currently will never execute. The intention seems to be that the logic in the commit below should trigger if the user provided a file with a ".zwo" extension. However, Qt is automatically appending a ".erg" extension in the _File Save As_ dialog so you get something like "workout.zwo.erg" which results in the file format being erg. This should be easy to fix...

https://github.com/GoldenCheetah/GoldenCheetah/commit/6f34cf2b6715258c89876af3977cac2848256676

kamilj:
> I still don't really understand why they invented a new format.
Because .MRC sucks.
Not suitable for export/import.
No tags, multiline description, erg/slope mode etc.

liversedge:
When I click 'save as' on the workout editor and type in test.zwo it writes to zwo format.
I'm running on Linux with Qt 5.9.2

arunh:
@liversedge on Mac it gets a .erg ext appended even if I enter a .zwo
extension. The QFileDialog uses the system provided one on Mac and Linux I
believe so this could explain difference in behaviour. Either way I think I
have a fix but will need someone to test hasn't regressed Linux...

PR #2768, opened 2018-01-29
Adding file type selection when saving workout (#2767)

arunh:
Ensure the correct file extension is appended during workout save. Adds a drop down to select file type as shown below - the selected extension is appended to the filename entered by user if necessary by Qt/system inbuilt functionality.

screen shot 2018-01-29 at 08 41 26


liversedge:
Thanks. No issues with this, its the same for exporting an activity.

Issue #2681, opened 2017-11-02
No Danniels points in running activities


jmrius:
I get Danniels points in cycling activities, but not in running ones.

GC version = 3.5 DEV1710

grauser:
Currently Daniels Points in GC are adapted for cycling using power (instead of pace for runner in the original Daniels Points) and standardized to 100 points for one hour at CP like BikeScore or TSS (instead of 33 in original).
Now GC is for runner too. Perhaps could we have both. 

In 2009 @srhea, after adding this metric, proposed to rename it "Daniels Hundreds" (https://groups.google.com/d/msg/golden-cheetah-users/SpL_zfkLvPc/goyC1QsNiTwJ)

Walibu:
I am interested as well in the same function. As a short distance triathlet I prefere DanielPoints for cycling and for running/swimming I set them manually according my feeling.

Today I was looking through the code of src/Metrics/DanielsPoints.cpp. And it looks quite easy the add the Danielpoints for CS (Critical Speed) like it is done for CP, if sport is set to "Run" (item->isRun)

A point to think about is: if somebody uses a powermeter for running like stryd, see:
https://www.dcrainmaker.com/2015/01/stryd-first-running.html
A solution would be, just check the running activity, if there are raisonable powervalues saved, use power for daniel points, which is more accurate than speed if running in hilly enviroment. Or let the user choose. But easier to implement is, use power if we have it.

An other not easy task would be, if we want to adjust the speed according ascending/descending, like in Strava they have a function for that.

First I would try to implement it, without speed adjust. Just take the current speed and filter it perhaps like a 10 sec rolling average, that peaks do not count and then use CS like (->getCV(item->paceZoneRange)) instead of Critical Power CP.

amtriathlon:
If you want speed adjust considering changes in potential and kinetic energy you could use the xPace metric instead of average pace

Walibu:
Thank you for that hint, I see you have written a lot about that in Govss.cpp and Coggan.cpp.

A question:
Is xPace not similar to xPower which is Normalized Power and for running xPace would be the Normalized Pace?

In Govss.cpp I see the function running_power() which looks like calculating RunningPower according the slope and runners weight/height which would be very useful.
But instead of power I would like to have an adjusted speed according the slope, higher if running upwards and lower than the current speed if running downwards.
Like GAP in Strava (Grade Adjusted Pace)

amtriathlon:
[...]

PR #2776, opened 2018-02-03
Added DanielPoints for running, with GAP grade adjusted speed

Walibu:
Issue: https://github.com/GoldenCheetah/GoldenCheetah/issues/2681

Calculated Danielpoints for running according speed (average over 10 secs) and running speed adjusted according slope (grade adjusted pace model)
Similar to research from Minetti 2002, but a little lower factor uphill and much lower downhill, see
points from strava curve https://medium.com/strava-engineering/an-improved-gap-model-8b07ae8886c3

If there are power values in the running datafile, like from stryd, .. Danielpoints will be calculated according the power and not speed. If somebody is using a powermeter for running, he likely wants to analyse according power.

### DIFF
--- a/src/Metrics/DanielsPoints.cpp
+++ b/src/Metrics/DanielsPoints.cpp
@@ -19,6 +19,7 @@
 #include "RideMetric.h"
 #include "RideItem.h"
 #include "Zones.h"
+#include "PaceZones.h"
 #include "Context.h"
 #include "Athlete.h"
 #include "Specification.h"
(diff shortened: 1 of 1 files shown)


liversedge:
You should isolate the calculation to only work if it is (a) a run and (b) have altitude and speed (c) no power present.

Walibu:
(a)
Yes, for swimming it does not make sense, because the recorded swimming speed is often not reliable and gives utopic danielpoints. I thought there is a pace zone in GC for swimming and TriScore uses them as well. Thats why I let it for calculation and it corresponds with the TriScore value. Maybe somebody will like it, but he could also be confused.

(b)
If there is no altitude but speed, maybe it makes sense to calculate danielpoints. If somebody records running on a track or almost flat road, he would be happy to have it anyway.
For calculation I use slope, which per default is set to "0.0" in RideFilePoint() and if no altitude is present it will stay "0.0" and I use the speed value without calculation.

(c)
Yes, it is checked and no computeOnSpeed if power present.

amtriathlon:
I think Mark's observation refers to the need to add an isRelevantForRide method to avoid computation when it is not necessary.
OTOH I liked the idea to use GAP for this and probably we could add it as an independent metric also, but I have some concerns with respect to the implementation:
1) there is a method to check if power data is available without the need to traverse all the datapoints (see areDataPresent).
2) running_grade_adjusted seems to complex and large to be inlined and to allocate slopefactor array in the stack on each call seem very inefficient, this can be avoided making it static but I think would be better to offload the interpolation fitting a 5th order polynomial to the data and use the resulting equation in the code, which would be simple enough for in-lining, for example:
slope_factor

Walibu:
Thank you, good input!

1. Yes I was looking for a method, could not find it at the moment, areDataPresent()->watts does it, so no need to traverse all points, my fault.

2. I was not happy about it also. Will try to implement a interpolation, like your suggestion.

3. GAP as independent metric.
Yes I had that also in my mind, like point->kphgap, but was not sure where to start and what the implications are.

Walibu:
For your information, I did a correction for 1. and 2, see:
https://github.com/Walibu/GoldenCheetah/commit/d5b521e59aa1208ea7e49d7dda287d8934e04548

Checked it with my recent 20 runs, which gives the same values except 1 sample which gave 30 instead of 31 danielpoints (this run has some km slightly downwards with 1% slope) but all in tolerance, not sure whether 30 or 31 matches my running stress score.
The polynom gives slightly lower values for -1 to -3%, but its just minor and it will depend more on running skills than those percent.

Walibu:
For your information, I did a correction for 1. and 2, see:
https://github.com/Walibu/GoldenCheetah/commit/1745903212ed19c9b92a86cf9232a1ea7ed370e3

Checked it with my recent 20 runs, which gives the same values except 1 sample which gave 30 instead of 31 danielpoints (this run has some km slightly downwards with 1% slope) but all in tolerance, not sure whether 30 or 31 matches my running stress score.
The polynom gives slightly lower values for -1 to -3%, but its just minor and it will depend more on running skills than those percent.

amtriathlon:
Great! Can you rebase to consolidate all the changes on one commit for a cleaner merge?

Walibu:
Ok, I rebased it to one commit, so far.

Maybe, need to extract the function running_grade_adjusted() to separate metric file. Would be nice to have it as data series.
Currently I am happy with the Daniels Points for running and cycling.

amtriathlon:
Ok, I will cherry-pick your consolidated commit and will add some validations and isRelevantForRide plus to update the DBSchema to force the recomputation of metrics.
I'm not sure about the GAP as derived data series, it is a lot of work if you want to do something useful with it, for display it could be added as User Data to a cloud chart.
Something simpler would be to add an (average) GAP metric similar to (average) Pace which can be used for activities or intervals.

amtriathlon:
I will close this since the relevant commit is in master: https://github.com/GoldenCheetah/GoldenCheetah/commit/0d94e0200ef7764e2233af5b2358b3700962d88b

Walibu:
Thank you!

PR #2782, opened 2018-02-09
Bugfix: GC crashes on DB update, fixed if paceZoneRange < 0

Walibu:
In my previous commit, https://github.com/GoldenCheetah/GoldenCheetah/commit/0d94e0200ef7764e2233af5b2358b3700962d88b there was an error, checked "zoneRange" instead of "paceZoneRange". This causes a segmentation fault during DB update (no pace zone defined before certain date)

### DIFF
--- a/src/Metrics/DanielsPoints.cpp
+++ b/src/Metrics/DanielsPoints.cpp
@@ -175,11 +175,11 @@ private:
         double lastSecs = 0.0;
         double weighted = 0.0;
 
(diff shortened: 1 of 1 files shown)


amtriathlon:
That's right, but there are also another couple of related copy-paste errors: in the same line it should be paceZones instead of zones and isSwim instead of isRun, I will fix them altogether.

amtriathlon:
https://github.com/GoldenCheetah/GoldenCheetah/commit/f215e2b43358bd446e827fab024c137c384f0478

PR #2784, opened 2018-02-12
Allow slope up to 40% instead of only 20% and remove rounding effect from smoothing

Walibu:
I figured out, that the slope value is limited to 20% 
But there are a couple of paved roads which are steeper :) And and the strave GAP curve goes up to 34% with the interpolation function in DanielsPoints calculation we have reasonable values up to 40%.

An other issue I saw: 
after smoothing the slope value, there are rounding effects like 1e-12 instead of 0, which causes the function running_grade_adjusted() to calculate instead of returning 0.
See for e.g. 8.881779999999 e-17
slope_closeto_zero

### DIFF
--- a/src/FileIO/RideFile.cpp
+++ b/src/FileIO/RideFile.cpp
@@ -2493,7 +2493,7 @@ RideFile::recalculateDerivedSeries(bool force)
                 } else {
                     p->slope = 0;
                 }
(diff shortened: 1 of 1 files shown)


Joern-R:
Useful for MTB and Running - Thanks.

PR #2799, opened 2018-02-25
FilterHRV: Recalculate rest HR based on filtered AVNN. Fixes #2797

### LINKED ISSUE #2797: HRV - R-R Filter-Tool: Calculate avg HR based on filtered values
The short timespans used for HRV measurements make avg HR calculation susceptible for outliers. Therefore it would be better to calculate avg HR based on the R-R filtered HR values only (the ones with "R-R flag" = 1)

riccioclista:
This PR changes the behavior of the R-R HRV filter tool. Previously the rest HR was calculated from the "average_hr" metric of the data series imported. This metric was not subjected to the filter of the R-R HRV filter tool and thus yielded skewed rest HR values because of the relatively short time spans of HRV measurements. This PR changes the behavior so that HR is derived from AVNN which is calculated based on filtered R-R values only.
Also an option has been added to correct HR values of past measurements.

### DIFF
--- a/src/FileIO/FilterHRV.cpp
+++ b/src/FileIO/FilterHRV.cpp
@@ -283,10 +283,12 @@ FilterHrvOutliers::postProcess(RideFile *ride, DataProcessorConfig *config=0, QS
         if (setRestHrv) {
             RideItem *rideItem = ride->context->rideItem();
 
+            double avnn = rideItem->getForSymbol("AVNN");
+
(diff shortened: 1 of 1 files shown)


amtriathlon:
It makes sense, I cherry-picked the first commit and merged to master, I don't think the second one is necessary since only dev builds have been published and it is possible to recompute the measures using formulas if the user really wants to.
Thanks for the contribution.

PR #2804, opened 2018-02-28
Settings: Allow using different global conf file name

riccioclista:
This enables setting a different global conf file name for testing purposes. The default (./config/goldencheetah.org/GoldenCheetah.ini) might be used for "production" instance of GC. Using a different .ini file allows testing without compromising the "production" settings and data.


Joern-R:
Useful - thanks.

PR #2805, opened 2018-02-28
LTMSidebar: Load most recently selected phase on GC start

riccioclista:
This PR saves the most recently selected phase in GC_LTM_LAST_DATE_RANGE, so it can be loaded and selected on the next GC start. (This is already implemented for seasons/date ranges but not phases so far.)
Also: The season containing the selected phase is expanded on GC start.


Joern-R:
I did not (yet) use phases - but with this may consider it in future. Nice little enhancement.
Thanks.

PR #2806, opened 2018-02-28
Imagic extended

peterbrant14:
**Experimental support for Tacx Imagic Trainer**

This is split into 2 separate commits. 
The first provides the basic imagic support and is relatively self-contained, consisting mainly of new sections of code with minimal changes to existing. This can stand alone if required.

The second adds extended support to use the Tacx steering accessory as a means of scrolling the Train view. However it does hit a few more areas of existing code.

One thing to check:
In order to support imagic on Windows, I needed to remove a check within the ctrl_msg procedure in EzUsb.c. Previously this read:
#ifdef WIN32
    return -1;

This is now:
#ifndef GC_HAVE_LIBUSB
    return -1;

Given that EzUsb.c is only included in src.pro if GC_HAVE_LIBUSB is set anyway, I think my new check is probably redundant. I can't actually see how USB devices like fortius would have worked at all under windows with the existing code as no USB I/O would ever be done. I am new to GC though, so its possible I've completely misunderstood something.

The Imagic code is very much based on the existing fortius support, with similar functions for the handlebar controller buttons etc. However, where fortius used a dedicated ".hex" file from which to load firmware, Imagic has it embedded in the USB driver. We therefore ask the user to supply a copy of the I-Magic.sys driver from which we can extract the relevant data.


liversedge:
It looks pretty standalone and shouldn't affect other device types?
I think the code looks clean, so happy to merge and see if it works for others?
Make sense?

liversedge:
Not sure about the libusb changes -- we run on Linux, MacOS and Windows.
Why did you makes those changes?

peterbrant14:
The change to the ctrl_msg procedure in EzUsb.c was required to make it work on windows. The existing code simply disabled any USB I/O from EzUsb if running on a windows platform. Like I said, I can't really understand how fortius devices that also use this code would have been working on windows. Linux and MacOS would be OK, but I'm sure it would not have been possible to load the firmware on windows. I suspect it was done to prevent unresolved references to libusb0.dll when the code had been compiled with the GC_HAVE_LIBUSB flag, but libusb was not available on the machine?

I'd be happy to revert that particular section, but it would mean it would only work on Linux or Mac.

liversedge:
will look at it and get back to you

liversedge:
@Joern-R can you explain to Peter how the libusb0.dll embedding works from commit d2a97fbce2d5df436a574dbf0a125c8e9766bf70, since I think the change to EzUsb.c in this patch will break Mac/Linux builds ?

peterbrant14:
Ah - Thanks for pointing me at that commit. It makes a lot more sense now I can see the original intention. Sorry - I should have looked into that a bit more deeply.
So I think my change as it stands would be OK for Mac and Linux, but cause problems for Windows users who do not have libusb installed. 

The quick and easy fix would be to revert the ctrl_msg function in EzUsb back to how it was, but that would rule out imagic running on Windows (and I think also the T1942 Fortius).

A better solution might be to explicitly link to the usb_control_msg function in LibUsb.cpp in the same way that the other libusb functions are referenced and then pass that as a pointer to EzUsb to use instead of its own reference.  I'll add another commit to do that so you can see what I mean....

Joern-R:
Was just trying to remember what the reason for the change was - I think we had "libusb0.dll" as part of the installer locally for GC before - which was not a good way - e.g. if a different version was already installed on Windows. Without the change GC did not link/did not start (not sure). 

If you have an variant how to patch it so it works without a local "libusb0.dll" in the GoldenCheetah Directory for Windows - happy to test it. I also think Mac and Linux should not be affected by your change. I can have a deeper look next weekend. Have to admit that my change was "not very educated" on what the function does - and since nobody on Windows complained ;-) it's still in this way.

liversedge:
Hey @Joern-R I will wait for your thumbs up and then merge -- I think the code is good and aside from the libusb0.dll piece is pretty much standalone.

peterbrant14:
I've just added a change that I think will solve the libusb problem. See commit comments for details.
I've tested on Ubuntu (16.04 LTS), Windows 10 and Windows 7.  On windows systems without libusb installed, GoldenCheetah starts up OK and just refuses to connect to the Imagic, which I think is the behaviour we want. With libusb available everything seems to work OK (at least for me...).

liversedge:
And if libusb0.dll is not found it should fall back to the version published with GC ?

I lost touch with this bit of the code so will wait for @Joern-R to say either way.

Really appreciate your efforts to stay with this. I'm pretty sure there will be a few folks out there that want to try this. It comes up on the forums every now and again.

peterbrant14:
If libusb0.dll is bundled in the GC directory it will use that one, otherwise it will traverse the normal Windows search path.

Glad to be able to contribute - I hope it proves to be useful to someone.

Joern-R:
FYI - I do not bundle libusb0.dll with GoldenCheetah since a number of versions - basically starting with the fix you mentioned. Seeing the last fix from Peter there was obviously a different way to achieve the same. Sorry - did not know that.  Will also build / have a brief test with the current delivery environment over the weekend (but do not assume any surprises based on what I saw/read). Thanks.

Joern-R:
@peterbrant14 @liversedge - if have build the PR - thanks Peter for the last change - no problem to run GC without providing an own version of "libusb0.dll" in the installer. I can't test functionality since I have not USB devices connected to GC - but that should be good for the next dev- build (and better than workaround I have added in the past.

PR #2812, opened 2018-03-10
 Deprecate Bing Map and set default tile server URL for new OSM maps

### LINKED ISSUE #2717: Bing Map not working/available any more
For none of my rides the Bing Map is working anymore.
I am still investigating - looks like Bing has changed 
their Terms of Use and require an API key now.

If someone picks this up - very welcome.

amtriathlon:
Remove Bing code
Make OSM the default map type and use it for existing Bing maps
Fix #2717

New OSM maps had a blank default tile server URL and they are the new default.
Also include minor change to use an existing method and to avoid currentData.


amtriathlon:
There is a previously existing bug I couldn't fix yet: the first time RideMap is activated on a previously selected ride it is not correctly positioned, only after selecting a ride while RideMap is visible it works as intended.

liversedge:
looks good to me.

PR #2826, opened 2018-03-18
Fixed LTM Chart to show Events on Phases and Season end date

### LINKED ISSUE #2620: Events is not appearing on Phases Metric Trends Chart's
When you create Phases from an Event, only the first Phases shows Events on a Metric Trends Chart. Any one above it does not show the events on the chart.

amtriathlon:
Events ocurring on Season end date were not showing
Seasons ending, but not starting, on chart date range were not considered when looking for events
Fixes #2620


desie:
In Trends module, when selecting a phase in a season, if phase is not starting the first week of the season then events are not showing.

PR #2829, opened 2018-03-19
Consistency checks for seasons-phases-events, event reordering, priorities and description

### LINKED ISSUE #565: Events Date Ordering & Column Ordering
At the min there is no way to change the order of events, which for me would be logically arranged chronologically therefore it would be good to be able to either manually reorder events list or apply a sort to the list.

I would much prefer the date down the left side than the right as the event name length affects where the date is shown and it can force it off the right side if the name is long. I would much rather my date take the fixed left position and the name to overrun the right edge  as I can often still get enough info from the name to know what is or amend it's length however when a date ends up off the edge it's a bit useless. I think this feature would be best implemented in the same way that the activity list is in that the column positions can be moved around as suits the user.

### LINKED ISSUE #1724: Add priorities to events
- Add priorities to events e.g. A race, B race etc.

- Be able to organise events in the side panel by date or priority making them easier to find.

### LINKED ISSUE #1943: Allow events to stay always on top
Currently events are displayed on top of the uppermost metric in the Tracker chart. Once you start scrolling down for more metrics, the event lines are still present, but event names aren't visible anymore, so with a busy schedule ahead it's hard to tell which is which. It would be helpful to see events just stick to a certain position in the viewport, sorta like freezing a row in a spreadsheet app of your choice. Am I making any sense? :)

### LINKED ISSUE #1979: Sync Events with calendar
It could be very useful, to see events in your calendar (google calendar, ical, etc..)

### LINKED ISSUE #2617: More event information
A new field for a more event detailed information

event_descreption

amtriathlon:
Names must be not empty
Phases must be fully included in the corresponding season
Events must belong to the corresponding season
Enable Event drag&drop in LTMSidebar to allow reordering - Fixes #565
Add priority to Season Events - Fixes #1724
Add Description to Season Events - Fixes #2617 
Refresh LTM charts when seasons change, so season boundaries and event markers are updated accordingly
Show Season/Event label in stacked LTM Charts for each 4 charts - Fixes #1943
Sync Events with Calendar if remote calendar is configured - Fixes #1979
Reenabled remote calendar configuration and autorization - To upload and download Events, tested with Google Calendar

### DIFF
--- a/src/Charts/LTMPlot.cpp
+++ b/src/Charts/LTMPlot.cpp
@@ -55,8 +55,8 @@
 
 //#include <QDebug>
 
-LTMPlot::LTMPlot(LTMWindow *parent, Context *context, bool first) : 
-    bg(NULL), parent(parent), context(context), highlighter(NULL), first(first), isolation(false)
+LTMPlot::LTMPlot(LTMWindow *parent, Context *context, int position) :
+    bg(NULL), parent(parent), context(context), highlighter(NULL), isolation(false), position(position)
 {
     // don't do this ..
     setAutoReplot(false);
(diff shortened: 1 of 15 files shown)


amtriathlon:
Replaced by https://github.com/GoldenCheetah/GoldenCheetah/pull/2844

PR #2837, opened 2018-03-26
Improve RLV Syncronization

### LINKED ISSUE #2817: Slope in RLV mode: Resistance change is ahead of video
I am a new user to the train mode and used RLV for 4 runs now. The machine and GC emulate slope well. I feel a delay between change in slope and the video. I feel the hill incline before I get to the base of the hill. Also it becomes easier and downhill like resistance before reaching top. I am talking of up to 1 minute offset within a 1 hour ride. I know the PGMF and RLV data are ok, because I had used the same on Tacx software in the past and they were spot on. I had read, that some users feel the resistance changes being delayed. However, this is vice versa. Video lags behind resistance change.

peterbrant14:
Fixes #2817 
I think this PR might sort out a few sync issues with riding RLVs in train mode that I've encountered over the past couple of months of using it. 
Each of the commits solves a particular issue, so could be cherry picked if required.

Any use?

### DIFF
--- a/src/Train/ErgFile.cpp
+++ b/src/Train/ErgFile.cpp
@@ -401,6 +401,13 @@ void ErgFile::parseTacx()
     // if we got here and are still happy then it
     // must have been a valid file.
     if (happy) {
+        // Set the end point for the workout
+        ErgFilePoint end;
+        end.x = rdist;
+        end.y = Points.last().y;
+        end.val = Points.last().val;
+        Points.append(end);
+
         valid = true;
 
         // set ErgFile duration
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -1670,6 +1670,11 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                 if (!(status&RT_MODE_ERGO) && (context->currentVideoSyncFile()))
                 {
                     displayWorkoutDistance = context->currentVideoSyncFile()->km;
+                    // If we reached the end of the RLV then stop
+                    if (displayWorkoutDistance >= context->currentVideoSyncFile()->Distance) {
+                        Stop(DEVICE_OK);
+                        return;
+                    }
                     // TODO : graphs to be shown at seek position
                 } else {
                     displayWorkoutDistance += distanceTick;
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -1667,9 +1667,9 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                 displayLapDistance += distanceTick;
                 displayLapDistanceRemaining -= distanceTick;
 
-
-                if (!(status&RT_MODE_ERGO) && (context->currentVideoSyncFile())) {
-                    displayWorkoutDistance = context->currentVideoSyncFile()->km + context->currentVideoSyncFile()->manualOffset;
+                if (!(status&RT_MODE_ERGO) && (context->currentVideoSyncFile()))
+                {
+                    displayWorkoutDistance = context->currentVideoSyncFile()->km;
                     // TODO : graphs to be shown at seek position
                 } else {
                     displayWorkoutDistance += distanceTick;
--- a/src/Train/VideoSyncFile.cpp
+++ b/src/Train/VideoSyncFile.cpp
@@ -72,6 +72,7 @@ void VideoSyncFile::parseRLV()
 
     // running totals
     double rdist = 0; // running total for distance
+    double rend = 0;  // Length of entire course
 
     // open the file for binary reading and open a datastream
     QFile RLVFile(filename);
@@ -228,6 +229,7 @@ void VideoSyncFile::parseRLV()
                             happy = false;
                             break;
                         }
+                        if (courseInfo.end > rend) rend = courseInfo.end;
                     }
                     // FIXME : add those data to training messages
                 }
@@ -242,6 +244,17 @@ void VideoSyncFile::parseRLV()
             block++;
 
         } else happy = false;
+    }
+
+    // Make sure we have a sync point to represent the end of the course
+    // Some earlier rlvs don't supply an end point frame reference,
+    // so we need to assume the speed from the last ref point continues to the end
+    if ((!Points.isEmpty()) && (rend > Points.last().km * 1000) && Points.last().kph > 0) {
+        VideoSyncFilePoint add;
+        add.km = rend / 1000;
+        add.kph = Points.last().kph;
+        add.secs = Points.last().secs + ((add.km - Points.last().km) * 3600) / add.kph;
+        Points.append(add);
     }
 
     // done
--- a/src/Train/VideoSyncFile.cpp
+++ b/src/Train/VideoSyncFile.cpp
@@ -164,6 +164,8 @@ void VideoSyncFile::parseRLV()
                 {
                     // read in the mapping records
                     uint32_t PreviousFrameNbr=0;
+                    double PreviousDistance=0;
+
                     for (unsigned int record=0; record < info.records; record++) {
                         // get the next record
                         if (sizeof(framemapping) != input.readRawData((char*)&framemapping, sizeof(framemapping))) {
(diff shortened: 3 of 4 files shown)


ericchristoffersen:
wrt the videosyncfile.cpp changes: I would like to understand why is it more correct to interpolate distance from averaged speed instead of using the rlv mapping directly? I would think that rlv mapping was adjusted manually by creator and could not be improved upon. I ran the new and old calcs in parallel and on all my rlvs the mapping distance was monotonically increased by < 0.1%. What is the case where this averaging helps or matters? Are there a set of choppy rlvs that need this?

peterbrant14:
Hi,
Its been a while since I posted this PR, so although I've been running it ever since in my version of GC, I had to refresh my memory on the logic.

The problem I was trying to solve was that when riding an RLV, the slope (ie resistance) and video seemed to get gradually more and more out of sync as the video went on. I came to the conclusion that this was caused by the way that distance was calculated within videoSyncFile.cpp.

In the PGMF file, slope is set according to the distance into the workout. Meanwhile, in the RLV file, the video filming speed (expressed as distance travelled per frame) is given for various points in the video, defined by frame number. Therefore, videoSyncFile.cpp needs to calculate the distance into the workout that each RLV-file sync point represents. The existing code assumed that the video speed changed abruptly at each sync point and calculated the distance accordingly. However, I think what happens is that the video speed changes gradually from one sync point to the next. This means each time there was a speed transition, the existing code introduced an error in syncronisation between distance and video frame number. The problem was worst in videos with many speed transitions.

Through trial and error, I determined that the correct way to calculate the distance was to apply a weighting factor such that the average distance travelled per video frame between two sync points was calculated as (2A+B)/3 where A is the lower speed and B is the higher. This allowed for the gradual speed change between sync points and gave me accurate syncronisation across the whole RLV workout.

I use several (mainly old) RLVs with GC. A good example was Tacx Tour of Flanders 2007, which is quite "up and down". Before I applied my changes, the resistance of the trainer would ramp up hundreds of metres away from where the hills began on the video. After I applied my changes, the syncronisation was perfect. 

I'm not entirely happy with the corresponding calculations in VideoWindow.cpp, which attempt to interpolate the speed between sync points in order to play the video as smoothly as possible. It works for me, but I can't help thinking there should be a simpler way to achieve it, that maybe I've missed.

ericchristoffersen:
I get a delay for slope change of 4-5 seconds for most of my rlvs. I don't know where the delay comes from but it doesn't increase. When I run my rlv with the frame averaging change I don't see much difference at all (which I guess is good) but then I can't figure what is different about your files that would fix your progressively worsening synchronization issue. Hundreds of meters wrong is pretty big mistake!

I'd like to understand what is different about your files. Could you share your rlv and pgmf? I don't need the avi.

Each point in frame mapping implies an instant change in frames/meter, but it also specifies what frame will be played at that point.

Hypothetical case:

Section A of length 50 has 5 frames/meter
section B of length 50 has 10 frames/meter
section C of length 50 has 5 frames/meter

I think maybe ideal: Criteria is that 250 frames should played before section B is reached, etc. If we smooth the frame rate then the early part of A must be lower than average so that the frame rate in section A can finish at more than 5 frames/meter. Section B should start below 10 frames/m, raise its frame rate above 10 frames/m, then drop it back down, and 500 frames must play.

This could be done by synthesizing a cubic function to model the framerate at the specified points, then dropping it until it integrates to the right number of frames per range.

But I still don't understand how anything can fix a worsening sync problem (which implies counts are lost.) What I'm afraid of is that if your averaging is fixing a progressively worse frame rate that its actually missing the precisely described section boundary frame mappings. Is it possible that your files are just broken?

Eric

peterbrant14:
Hmm. Maybe I was stretching things a bit when I said hundreds of metres. Maybe tens would be more accurate - sorry. Anyway it was enough to be annoying (4-5 seconds would qualify as annoying for me). 
I also found that the workout would end before the end of the video, which seemed to support my theory that there was an accumulating error in the running distance calculation in videoSyncFile.cpp.

As I understand it, the section boundaries in the RLV file are defined in terms of frames, not distance. Therefore a running distance value has to be calculated.
If we take an example of 2 sync points:
A: frame 1, 1 metre per frame.
B: frame 100, 2 metres per frame
then the original code would calculate the distance covered over the 100 frames as 100 metres.
My version would calculate weighted average speed = (2x1 + 2) / 3 =   1.33 metres / frame, therefore 133 metres covered.

I know these aren't realistic numbers but I'm just trying to illustrate the point that there is a significant difference in my way of calculating distance vs the original. I guess whether or not the sync error increases would be down to the particular RLV. Errors introduced by speed-ups might be cancelled out by those introduced by slow-downs. I didn't measure this scientifically, it was just a perception. 

I have quite a collection of RLVs and while I could believe that some of the downloaded free ones are broken, I would expect the official Tacx ones to be good. They all ride OK on my Tacx Fortius software.

ericchristoffersen:
Ah. I totally missed that it was rate. So it might be it is sometimes derived from the speed at the framedata point instead of giving the average rate for the previous segment. Bummer. Would have been so much more precise to just use distance, which wouldn't fall out of sync if you used the wrong interpolation. If there was a 'correct' specified rate to distance interpolation then everyone could agree, maybe there is? But without a spec we're left with magic thinking.

Thinking magically I hacked up a per frame rate interpolation using cubic interpolator. It sums an interpolated rate for each frame and I get pretty similar results. My longest rlv finishes 4 meters ahead. I ran this in a standalone app, haven't integrated it into gc. Might be fun to try.

0: 0.248ms orig gc approach sum: 20425.6​
1: 0.0909ms​ peter  gc approach sum: 20425.1​
2: 6.9074ms​ cubic approach sum:   20429.5

But fundamentally we're at the mercy of how steady the ride recorder's speed was, or some good choices by the file maker. A sportbike with lots of extreme stopping (around the framedata snapshots) and an instant speed calc would leave us with an un-sync-able video.

You have the tacx software, their trainer and the tacx rlv - the full meal deal. Do you think your averaging gets GC to matching (or exceeding) the tacx behavior? If so that is fantastic, hard to expect better than that.

Also: thanks so much for taking the time to respond.

Eric

peterbrant14:
Yes - its a shame Tacx don't seem to publish information about their file formats. If there is any, I'd love to see it...
To answer your question, yes - the weighted averaging now seems to perfectly match the ride I get when using my old Tacx Fortius software and my even older Imagic trainer.

Regards,
Pete.

ericchristoffersen:
> Yes - its a shame Tacx don't seem to publish information about their file formats. If there is any, I'd love to see it...
> To answer your question, yes - the weighted averaging now seems to perfectly match the ride I get when using my old Tacx Fortius software and my even older Imagic trainer.
> Regards,
> Pete.

I just spent more time than I can believe becoming intimate with the rlv format. I am now certain that your averaging is correct since its self consistant with other data in the rlv file. I am super thankful that you figured it out.

What really took me time was figuring out how to round trip that averaging system back to disk so I could resynchronize my rlvs.

I think this was an unnecessarily complex format for Tacx to choose. They could have easily stored the distance steps instead of velocity.

Anyway, I'm now using this with my own training rides and its working well.

amtriathlon:
So both agree this is ready to merge? Any possible disadvantage?

ericchristoffersen:
I think wait for 3.6.

ericchristoffersen:
Needs to be rebased, the changes don’t merge automatically. I have them working against current but not happy because of (unrelated?) VLC/widget issues.

peterbrant14:
Thanks for picking this up and putting in so much effort to verify it.
I'll try to get it rebased against current master...

peterbrant14:
Rebased onto current master as discussed.
The only possible disadvantage I can think of is increased CPU utilization because of the slightly more complex calculations used to sync the video. Having said that, I did run some monitoring and couldn't find any discernible difference.

amtriathlon:
Thanks Peter!

>I have them working against current but not happy because of (unrelated?) VLC/widget issues.

Eric, if you want to isolate them, truncate config/video-config.html to size zero and video overlays will be disabled on next session.

ericchristoffersen:
I just tried this latest against current master and it works really well.

amtriathlon:
Since it fixes pending issues with existing features and it doesn't change the UX I think it is a good idea to include it in v3.5, I will merge it.

PR #2841, opened 2018-04-05
Added ScriptContext for R/Python APIs called from DataFilter scripts

amtriathlon:
Carry RideItem, Specification and Metrics to consider the case when
the script is part of a UserMetric.
Python series, activityWbal and activityXdata honor RideItem and Specification,
allows them to be used in Scripts at UserMetrics and LTMPlot formulas
to obtaing the data points correspoding to the activity/interval.

### DIFF
--- a/src/Charts/PythonChart.cpp
+++ b/src/Charts/PythonChart.cpp
@@ -197,7 +197,7 @@ void PythonConsole::keyPressEvent(QKeyEvent *e)
                 line = line.replace("$$", chartid);
 
                 python->cancelled = false;
-                python->runline(context, line);
+                python->runline(ScriptContext(context), line);
 
                 // the run command should result in some messages being generated
                 putData(GColor(CPLOTMARKER), python->messages.join(""));
@@ -471,7 +471,7 @@ PythonChart::setState(QString)
 void
 PythonChart::execScript(PythonChart *chart)
 {
-    python->runline(chart->context, chart->script->toPlainText());
+    python->runline(ScriptContext(chart->context), chart->script->toPlainText());
 }
 
 void
--- a/src/Charts/RChart.cpp
+++ b/src/Charts/RChart.cpp
@@ -179,7 +179,7 @@ void RConsole::keyPressEvent(QKeyEvent *e)
 
             // set the context for the call - used by all the
             // R functions to access the athlete data/model etc
-            rtool->context = context;
+            rtool->scriptContext.context = context;
             rtool->canvas = parent->canvas;
             rtool->chart = parent;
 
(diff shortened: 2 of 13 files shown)


amtriathlon:
Python is complete for now and version number incremented, I will give a look to RTool adaptions.

amtriathlon:
The changes to RTool seem simple, see the commit above, but there is an underlying problem: R API is not being called from Custom Metrics on computeMetrics, only when I use the Test button.

I'm using these 2 "My Duration" metrics for testing, one based on samples:

```
{
   value {

%%R
GC.result(length(GC.activity()$time))
%% ;

   }
}
```

and the other based on Metrics:

```
{
   value {

%%R
GC.result(GC.activity.metrics()$Duration)
%% ;

   }
}
```

liversedge:
Weird. Am now looking at it.

liversedge:
Looks like scripts are being executed -- simplify to GC.result(20) in the script and it works fine.
I think there may be a reentrancy issue calling the GC.API in a thread ?

Looking deeper.

liversedge:
Yes, this is a very interesting problem that may not be resolvable.

R insists its event loop is on the main thread, QT also has an event loop on the main thread.

They do not and cannot be meshed. As a result the R API call back to the main thread is blocked by the QT event loop.

For now, I would remove Rscripting in user metrics. The only way around this is to place R in a separate process and interact with it in the same way Jupyter notebooks do it.

Make sense?

liversedge:
actually, i can get this to work, so long as I don't call an API from an API call

this works for me:
```
{
    value { 
%%R 
df <- GC.activity.metrics()
GC.result(df$TSS) 
%%;

 }
}
```

amtriathlon:
Hi Mark, it doesn't work for me that way either.
The behavior I'm seeing is the same with or without nesting of API calls:
- recomputing the metric after doing a Test takes the value of the Test for all activities and intervals
- quit, remove RideDB.json and restart the metric value is zero for all activities and intervals
Adding debug code to RTool::result and RTools::activityMetrics seem to indicate they are never called from computeMetrics, only from Test button.

liversedge:
Yes, oddly I am now seeing the same too. Perhaps I was mistaken.

For the sake of everyone's sanity I would just remove R scripting from DataFilters and make it python specific !

liversedge:
No the issue is the call is being made in the wrong thread and R maintains its own stack frame, so this is just corruption of memory structures.

We could use signlals/slots to call  the interpreter and return values across threads but tbh, its a lot of work that will probably break. I still think its easier to remove support for R scripts in data filters.

amtriathlon:
Ahh, that’s makes sense but it’s a PITA. Thanks for the explanation!
I will close this PR and will create a new one for Python only plus the removal of R Scrits in DataFilter.

PR #2856, opened 2018-04-29
Add Lap Distance and Lap Distance Remaining dials and telemetry

bwalding:
Adds new dials for the training screen that show Lap Distance and Lap Distance Remaining.

There are some slight display irregularities under test conditions - e.g. you've got 10m laps and are burning through them quickly - the lap distance can exceed the lap lengths due to how the display is updated. But when using longer laps this becomes less of an issue.

Predominantly for CRS files.


bwalding:
Displays are standard distance displays.

image

liversedge:
NICE !

PR #2866, opened 2018-05-12
RideSummary - Enable horizontal scrolling

amtriathlon:
The problem was reported at the forum (https://groups.google.com/forum/#!topic/golden-cheetah-users/bxxeQ8zPUeA), I can reproduce it and this change seems to fix the issue on Linux.
It is not tested on MAC nor Windows and I don't know if there is better solutions.


liversedge:
Hi Ale,

This is not how it is supposed to work. Horiz scrollbars are terrible ux.

It should never NEED a horizontal scrollbar.

In this case there are too many PB metrics on one line - the code should put them across multiple lines instead (or limit the number to track).

PR #2867, opened 2018-05-13
Add telemetry for time to next erg point

mark-inderhees:
This adds a new workout telemetry to see the time to the next erg point. This will allow the user to see how much time is remaining at the current erg point value. This is useful to see how much longer the user must pedal at the current target power.

### DIFF
--- a/src/Train/DialWindow.cpp
+++ b/src/Train/DialWindow.cpp
@@ -211,6 +211,7 @@ DialWindow::telemetryUpdate(const RealtimeData &rtData)
         break;
 
     case RealtimeData::LapTimeRemaining:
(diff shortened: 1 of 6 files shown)


mark-inderhees:
Sounds good. I'll make the change to remove nextPoint.

mark-inderhees:
@liversedge Can this be merged into master or do I need to complete more steps? I'm not familiar with the commit process.

mark-inderhees:
Cool, thanks!

amtriathlon:
When nextPoint was removed and changed by rightPoint was this tested again?

It doesn't seem to work except for the first section, but I don't know yet if rightPoint behavior changed after this commit.

PR #2878, opened 2018-05-30
various minor bugfixes + warningfixes

qheath:
a few parentheses/braces/break had been forgotten or misused (some of the corresponding bugs were almost invisible, some could have lead to segfaults like in `src/FileIO/RideFileCommand.cpp`)

in the process of correcting this, I also changed a lot of small things (indentation, comments, unused variables, etc) that caused compiler warnings, to make real problems easier to spot

the third and fourth commits look like stylistic changes with no operational impact, but actually derive from fixing an inconsistency in the treatment of `Leaf::Parameters` (so it's *almost* a bugfix)

hopefully the style formatting changes and the bugfixes are separated enough

### DIFF
--- a/lmfit/lmmin.c
+++ b/lmfit/lmmin.c
@@ -235,7 +235,7 @@ void lmmin(
     if( C->verbosity&2 )
         lm_print_pars(nout, x, msgfile);
     (*evaluate)(x, m, data, fvec, &(S->userbreak));
-    if( C->verbosity&8 )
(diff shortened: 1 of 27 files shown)


liversedge:
This will take a bit of reviewing, quite an opening pull request!
Will try and get a review done in a week.

liversedge:
Actually, they're fine, I took a peek during my lunch.
The datafilter ones were particularly smelly, thanks.

PR #2890, opened 2018-06-26
Appimage in travis

LyzardKing:
Hi.
I added the appimage support in travis.
To do that I moved the macos scripts to files in travis/osx, and at buildtime
travis calls the correct files.
The appimage works on bionic (the appimage script is the same that was there previously)
It temporarily uploads to transfer.sh for testing purposes.
The only thing that fails is the integration with dropbox/google/strava/..., but I think it's because
I don't have access to the secrets, that's why they're commented in the yaml for now..


LyzardKing:
The git log is a bit of a mess..but it'll be fixed
The appimage starts properly, and I can add a gpx file without errors.
The second time I load the program it fails with a core dump.

PR #2914, opened 2018-08-27
Change GPX decimal separator to dot

### LINKED ISSUE #2913: GPX export uses locale-dependent decimal separator
The GPX export uses a decimal separator which depends on your system locale for the export on latitude and longitude values. If that locale is e.g. de_DE, you end up with a comma as decimal separator in the exported GPX files.

You end up with
`<trkpt lon="7,9312479719" lat="49,978498886">`
instead of
`<trkpt lon="7.9312479719" lat="49.978498886">`

The latitudeType and longitudeType from the GPX specification derive from xsd:decimal, which is specified as `(\+|-)?([0-9]+(\.[0-9]*)?|\.[0-9]+)`, i.e. the decimal separator is a dot.

Using the incorrect decimal separator leads to other programs failing to parse the GPX file, I observed this with Garmin Connect.

I observed this issue with the current master revision, 2a6083071f3c3a3757b8136f702d944c58a8a67a

cyberfuzzie:
Fixes #2913

### DIFF
--- a/src/FileIO/GpxRideFile.cpp
+++ b/src/FileIO/GpxRideFile.cpp
@@ -83,12 +83,16 @@ GpxFileReader::toByteArray(Context *, const RideFile *ride, bool withAlt, bool w
         QDomElement trkseg = doc.createElement("trkseg");
         trk.appendChild(trkseg);        
 
+        QLocale cLocale(QLocale::Language::C);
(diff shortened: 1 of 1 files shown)


amtriathlon:
Thanks!

PR #2918, opened 2018-08-29
fix missing break in switch/case statement

vlcvboyer:
when HRV have been added no break was inserted in the switch/case directive.
All other messages are thereby decoded as HRV messages


liversedge:
thanks, good to see you back !

PR #2927, opened 2018-09-14
Add ignore zeros LTM setting

riccioclista:
This enables plotting measurement data without negative spikes for missing measurements


liversedge:
Seems good, will play with it.

amtriathlon:
Currently Measures (Body and HRV) are plotted as gapped curves, likely this is the origin of the conflict.
Although this is more general and could be useful for metrics and formulas, one issue we should look is to honor this setting for curve smoothing and moving average, these don't work well on Measures now and will not work well with this without other changes.

amtriathlon:
Thank you!

PR #2935, opened 2018-09-22
ANT: HRV measurement improvements

erikboto:
Improved handling of data page 4 for ANT+ HR.

I'm a bit worried about the check for *MeasurementTime != 0. That should lead to a loss of measurement every 1024 message on average for page 0 measurements and every 512 for page 4. Maybe it's enough to check that the number of heartbeats has increased? Anyway, that can be improved later.


liversedge:
Thanks

PR #2946, opened 2018-10-20
Build without ical

ianmartin:
Fixes to allow building without ical


amtriathlon:
Build fails when you remove CalDAV.h include from LTMSidebar.cpp, see the appveyor log.

ianmartin:
Sorry, now fixed

Issue #2113, opened 2016-09-30
Edge 800 temp measurement garbeld in graphs


Doorshifter:
Edge 800 is slow to start measuring temperatures. The start data of many rides looks like this because of that.
![screen shot 4](https://cloud.githubusercontent.com/assets/22545406/18995538/e9a014de-872c-11e6-8978-d1b217aa70c1.png)
empty record fields are handle by GC in a undocumented way resulting in a temperature graph like this.
![screen shot 5](https://cloud.githubusercontent.com/assets/22545406/18995579/1d9c5964-872d-11e6-8e13-07e53e0745c9.png)
which is a hot day, in degrees C  for any day in the world :-)

Copy paste of some values in de editor to the empty fields resolve the issue in the graph. Since it it only a few seconds (in the beginning of the ride) this could be solved in a different way in GC, with cause major inflouences. Ignore empty fields in temp records, set them to 0, take the first real occurrence and copy it in the empty fields...

amtriathlon:
I have been testing #2951 and found it is easy to fix Garmin 800 imported files with this issue using a Python data processor s.t. the following:
```
# pad temperature (assuming -255 == unset)
activity = GC.activity()
if ('temperature' in activity.keys()):
	temp = activity['temperature']

	# find first Temp
	firstTemp = None
	for i, a in enumerate(temp):
		if a != -255:
			firstTemp = (i, a)
			break
	# pad missing vals
	for i in range(0, firstTemp[0]):
		temp[i] = firstTemp[1]
```
Once created it can be run using Edit menu like builtin data processors, very nice!

PR #2951, opened 2018-11-02
Python data processors

riccioclista:
This PR adds the ability to create and execute python scripts that manipulate activity data, much in the same way as native data processors do. This enables the user to create their own data processors more easily. This PR adds the following features:
 * UI for managing python fix scripts (list, open for edit and delete)
 * UI for creating or editing scripts. This contains a PythonConsole (the same as in PythonChart) and the RideEditor-View. Scripts can be executed via this UI or by selecting them in the via the Menu "Edit" > "Python fixes".
 * a folder '.pyfixes' in GCROOT that contains all scripts. Scripts can be imported by simply dropping them into this folder
 * Python-API: Extension of the PythonDataSeries data structure to allow for setting values. Setting values is only enabled when run in the context of python data processor (not in PythonChart)
 * Python-API: New Method GC.deleteActivitySample(index, activity) to delete data rows
 * Python-API: New Method GC.deleteSeries(type,activty) to delete data columns
 * Command-Support for editing actions: All python script edits are wrapped in command contexts (LUW) to enable Undo and Redo
 * Integrate python scripts in existing data processor framework so they can be called on Save and Import
 * Python-API: New Method GC.postProcess(processor) to execute a data processor (python or non-python)
 * Python-API: new Method GC.createXDataSeries(xdata, series, unit)
 * Edit support of xdata series (via GC.xdataSeries(xdata, series))


liversedge:
Looks really interesting.. be interesting to know when youre ready to merge.

riccioclista:
Hopefully by next weekend. I'd like to add 2 more features: Integration with existing data processors framework, so they can be configured as import actions and the ability to call other data processors from python

riccioclista:
- rebased on cur master
- implemented last 2 features mentioned above
- not yet extensively tested

amtriathlon:
I think the Windows and OSX CI builds are failing since they don't include Python integration, since Python is optional the new code should be conditioned using #ifdef GC_WANT_PYTHON to avoid this problem.

riccioclista:
- #ifdef GC_WANT_PYTHON and "Enable Python" setting are now being checked

riccioclista:
I'll test it thoroughly this week so we can merge next weekend

riccioclista:
I've fixed GC.postProcess python implementation, which requires special handling since it must invoke any child python data processors on the same thread, otherwise it will deadlock. I've tested all the new functionality in this PR and haven't found any issues so far. It can be merged.

liversedge:
Do you have an example of using this?
Is it possible to have >1 script?

riccioclista:
Number of scripts is not limited. Scripts are stored in a .pyfixes folder (goldencheetah/athlete/.pyfixes).

Here are 2 sample scripts:

- [convert_to_indoor.py.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/3083401/convert_to_indoor.py.txt) (removes lon/lat and alt series)

- [import_fix.py.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/3083402/import_fix.py.txt) (trims data at the start with HR or LON/LAT values of 0 and pads invalid altitude data)

Remove the .txt extension, put them into the .pyfixes folder and restart GC. Then you can see them under "Edit --> Python fixes".

amtriathlon:
I started to play with this very nice feature and used one of the samples to fix a pending issue (https://github.com/GoldenCheetah/GoldenCheetah/issues/2113) and found a more interesting candidate would be  https://github.com/GoldenCheetah/GoldenCheetah/issues/2958, all is working as described, excellent!
I have not looked at the code yet, so I have a question:
I think we should restrict the use of the new Python APIs able to change data just to the Data Processor context and forbid its use in filter formulas an Python charts, is this included or have you see a way to include it s.t. them failing or having no effect on the unwanted contexts?

PR #2952, opened 2018-11-05
add support for serial daum devices

thebaron06:
Add support for daum devices connected via serial port.
Such device could be:
   - Daum Vital
   - Daum ergo_bike series
       * ergo_bike_2002
       * ergo_bike 4004 TRS
       * ergo_bike cardio
       * ergo_bike vita
       * ergo_bike fitness
The only device I own is a Daum Vital, therfore this is the only device
that is known to work correctly.

The default behaviour is to send the load to the device only once to
allow the user to interact with the device. The profile field is used
to pass a _force_ flag in form of a string containing the word _force_.
Doing so will cause the implementation to send the load to the device
every _poll interval_ (default 1s), except the training is paused.

Signed-off-by: Florian Nairz <nairz.florian@gmail.com>

### DIFF
--- a/src/Train/AddDeviceWizard.cpp
+++ b/src/Train/AddDeviceWizard.cpp
@@ -204,6 +204,7 @@ DeviceScanner::quickScan(bool deep) // scan quickly or if true scan forever, as 
     case DEV_MONARK : wizard->controller = new MonarkController(NULL, NULL); break;
     case DEV_KETTLER : wizard->controller = new KettlerController(NULL, NULL); break;
     case DEV_KETTLER_RACER : wizard->controller = new KettlerRacerController (NULL, NULL); break;
+    case DEV_DAUM : wizard->controller = new DaumController(NULL, NULL); break;
 #endif
 #ifdef GC_HAVE_LIBUSB
     case DEV_FORTIUS : wizard->controller = new FortiusController(NULL, NULL); break;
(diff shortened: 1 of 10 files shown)


liversedge:
There is no way we can test this, so assuming you have such a device?
The code appears to be good and shouldn't introduce a regression.
Can you provide a little bit of evidence that it works ?

Thank you.

thebaron06:
Hi, yes the code is tested. As I stated in the commit message I do own a _Daum Vital_ and tested it with this device. The documents I found regarding the protocol used all states that the same protocol applies to some daum_ergo bikes as well, see [here](http://pavel7575.sweb.cz/Pics/DAUM.pdf) and [here](http://www.daum-electronic.de/de/support/supp02.html).

This code does _not_ work for Daum ergo_bike 8008 because here Daum uses a different protocol that might be implemented on basis of this code, documents could be found [here](http://www.daum-electronic.de/de/download/bedaprem/SP-KOMM-PM1.pdf) and [here](http://www.daum-electronic.de/de/support/cla-bk8008.html).

If further proofs are needed just tell me.

liversedge:
No thats good enough for me!

Are you happy to merge, or is there any further testing/development pending?

thebaron06:
Actually it is production ready.
But I noticed that some method access specifiers are not set-up for re-usability or deriving from this class.
Also I noticed that we should pass in the whole _DeviceConfiguration_ in order to be more flexible with additions or we add a method to query them from the base class (RealtimeController) because there they are stored anyway.

But all these things are more _nice to have_ for less work in the future to extend daum devices.

So, with this in mind it depends on you. I can add the former additions if you like.

liversedge:
Will merge - updates can come through if/when you are ready.

PR #2968, opened 2018-12-01
PythonConsole: Print expression value when executing simple statements

riccioclista:
This PR makes the Python console behave more like a proper REPL shell in that it now prints the value of any expression entered automatically on the next line, e.g.:
```
>>> 1+1
2
>>> "Hello World"
'Hello World'
>>>
```


amtriathlon:
This is a feature I really wanted to have when testing Python scripts, thanks!

PR #2986, opened 2018-12-30
Decode Firstbeat/Garmin session metrics (VO2max, Training Effect, etc.)

hver:
Hi, I wanted to get the Firstbeat/Garmin session metrics from the watch into GoldenCheetah.
I spent two evenings trying to find them in the session message, without success. But with help of Google I found the issue on Runalyze github page, where they discussed and solved the problem: https://github.com/Runalyze/Runalyze/issues/2264
It's a bit of a mystery, how Garmin encodes the metrics which are calculated by Firstbeat logic on the watch, but the data of the session VO2max seems to be reliable in the newer watches.
Also I found a ruby fit file reader project, and it looks like the developers there decoded a lot of undocumented messages: https://github.com/scrapper/fit4ruby/blob/24dd34277c699e8215b72eeb1d23fd2b89fa4dc2/lib/fit4ruby/GlobalFitMessages.rb

With this information I wrote the code, to read following 6 metrics from the blob 140:

- **VO2max**: https://www.firstbeat.com/en/consumer-feature/vo2max-fitness-level/
- **Training Effect: Aerobic** : https://www.firstbeat.com/en/consumer-feature/training-effect-aerobic/
- **Training Effect: Anaerobic**: https://www.firstbeat.com/en/consumer-feature/training-effect-anaerobic/
- **Recovery Time**: https://www.firstbeat.com/en/consumer-feature/recovery-time/
- **Performance Condition**: https://www.firstbeat.com/en/consumer-feature/real-time-performance/
- **Lactate Threshold HR** and **Lactate Threshold Running Speed** : https://www.firstbeat.com/en/consumer-feature/lactate-threshold/

This Pull Request is still work in progress, because I am not sure, how do I save a session level metric in a right way. As Ale suggested, I used metadata like this:
` rideFile->setTag("Aerobic Training Effect", QString::number(value/10.0));`

But it is saved as a string, without dimensions, etc. Still, this metric is shown on the "Extra" tab, so I was able to modify it in the settings "Data Fields" to become a decimal metric, and after that use it in graphs.

Could someone give me an advice, how to do it in a proper way?

### DIFF
--- a/src/FileIO/FitRideFile.cpp
+++ b/src/FileIO/FitRideFile.cpp
@@ -689,6 +689,62 @@ struct FitFileReaderState
         rideFile->setDeviceType(getManuProd(manu, prod));
     }
 
+
+    void decodePhysiologicalMetrics(const FitDefinition &def, int,
+                                    const std::vector<FitValue>& values) {
+        int i = 0;
+
+        foreach(const FitField &field, def.fields) {
+            fit_value_t value = values[i++].v;
+
+
+            if( value == NA_VALUE )
+                continue;
+
+            switch (field.num) {
+                case 7:   // METmax: 1 METmax = VO2max * 3.5, scale 65536
+                    rideFile->setTag("VO2max detected", QString::number(round(value / 65536.0 * 3.5 * 10.0) / 10.0));
+                    break;
+
+                case 4:   // Aerobic Training Effect, scale 10
+                    rideFile->setTag("Aerobic Training Effect", QString::number(value/10.0));
+                    break;
+
+                case 20:   // Anaerobic Training Effect, scale 10
+                    rideFile->setTag("Anaerobic Training Effect", QString::number(value/10.0));
+                    break;
+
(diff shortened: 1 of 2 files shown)


hver:
> That's right, metadata tags are currently strings (see https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/FileIO/RideFile.h#L346) which are interpreted according to settings in metadata.xml

Thanks! I modified the metadata.xml and rebased the changes into a single commit.

liversedge:
This is great. Could you include an example file with metrics ?
They go in the tests/workouts folder.

liversedge:
sorry, thats the test/rides folder !

grauser:
I tested with one file from Garmin 1000 and one from 920 with new Metadata.

grauser:
I added the two files in the test/rides folder.

grauser:
Thank you Hennadiy.

PR #2990, opened 2019-01-11
Allow precise adjustment of the time slider with a keyboard.

hver:
Currently, it is very complicated to adjust merged activities for bigger activities. Every mouse click moves the slider 20-100 seconds. In this PR I allowed the slider to get keyboard focus, so now user can use cursor keys for 1 sec adjustments and pgDown/pgUp for 10s steps.
P.S. I tested this behavior on a Mac, maybe Linux or Windows are not so jumpy.


liversedge:
thanks !

PR #2991, opened 2019-01-11
Export .fit with correct activity type instead of "Ride". Fixed exporting "Runs"

hver:
Set the type of activity to Ride(2), Run(1) or Swim(5), instead of always Ride(2).
If exporting a Run, use rcad instead of cad

Lrbalance was never exported - fixed it. This was an interesting bug - QT Creator told me: "bitwise comparison always evaluates to false" - it took me some time, to understand why 😉 
Look for yourself:  `(type&2)==1`


hver:
@liversedge This is only a bug fix without any additional functionality. It would be great, if you could merge to the next dev release.

grauser:
I will merge it this week.

PR #3037, opened 2019-03-03
VC2017/Qt 5.12.1 - Fix 64Bit Build Problem

Joern-R:
... with windows / VC2017 / 64Bit the code does not build / cannot resolve a overwritten function when using QT 5.12.1
... adding the newly introduced function similar to the other function resolves the compile problem

### DIFF
--- a/src/FileIO/LocationInterpolation.cpp
+++ b/src/FileIO/LocationInterpolation.cpp
@@ -240,15 +240,10 @@ xyz UnitCatmullRomInterpolator3D::Interpolate(double frac)
 
 geolocation GeoPointInterpolator::Interpolate(double distance)
 {
-    return DistancePointInterpolator::Interpolate(distance).togeolocation();
-}
-
-bool GeoPointInterpolator::GetBracket(double &d0, double &d1)
-{
-    return DistancePointInterpolator::GetBracket(d0, d1);
+    return DistancePointInterpolator<SphericalTwoPointInterpolator>::Interpolate(distance).togeolocation();
 }
 
 void GeoPointInterpolator::Push(double distance, geolocation point)
 {
-    DistancePointInterpolator::Push(distance, point.toxyz());
+    DistancePointInterpolator<SphericalTwoPointInterpolator>::Push(distance, point.toxyz());
 }
--- a/src/FileIO/LocationInterpolation.cpp
+++ b/src/FileIO/LocationInterpolation.cpp
@@ -243,6 +243,11 @@ geolocation GeoPointInterpolator::Interpolate(double distance)
     return DistancePointInterpolator::Interpolate(distance).togeolocation();
 }
 
+bool GeoPointInterpolator::GetBracket(double &d0, double &d1)
+{
+    return DistancePointInterpolator::GetBracket(d0, d1);
+}
+
 void GeoPointInterpolator::Push(double distance, geolocation point)
 {
     DistancePointInterpolator::Push(distance, point.toxyz());
--- a/src/FileIO/LocationInterpolation.h
+++ b/src/FileIO/LocationInterpolation.h
@@ -20,7 +20,7 @@
 #define _LOCATION_INTERPOLATION_H
 
 #include <tuple>
-#include <bitset>
+
 #include "qwt_math.h"
 
 struct geolocation;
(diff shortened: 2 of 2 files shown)


Joern-R:
@ericchristoffersen - can you please have a look since you have developed this piece. With the change it's compiling fine wtih Qt 5.9.6 (32/64bit) and Qt 5.12.1 (32bit), but get a compile error not to be able to determine the overwritten function "/" - so I suspect an error in the VC2017 / 64Bit Qt/Windows area. Latest updates on VC2017 did not fix it. But adding the method fixed the problem. Thanks. Joern

ericchristoffersen:
Hi Joern! Happy to help. I'm on the west coast of the us, I will start looking at this about 13 hours from now.

Eric

ericchristoffersen:
Hi Joern,

I am running into errors when attempting to install that latest qt. The online installer is exiting with access errors. I'm still trying but would appreciate any hot tips you might have to get started with vs2017 and qt 5.12.1.

In the meanwhile I recompiled the locationinterpolation files in a standalone project in vs2017, made use of the problem methods for x64 and saw no issues (like you reported with other versions of qt.)

Also in my project settings under 'language' I disabled 'permissive' mode with /permissive-, and also enabled /Za and this made no difference either.

For a long while vs has been more permissive than other compilers when it comes to resolving names from templated parent classes because it implemented dependent name lookup as a single pass and the spec allows for a 2 phase (definition/instantiation) lookup. The flags above should make the compiler stricter and expose non-portable language issues. By default visual studio will resolve names that other compilers cannot resolve so it is safer for cross platform development if the goldencheetah vs2017 build were to enable /permissive-. The /Za flag is much more strict and won't compile windows sdk headers.

I'm suspecting that some sort of header processing is being done in this new version of qt that doesn't handle this case correctly, I believe that name should be available by the spec and the change you made is a no-op.

One thing I noticed is that I previously wrote:

DistancePointInterpolator::

I think it is more portable to use the full name of the base class:

DistancePointInterpolator<SphericalTwoPointInterpolator>::

Eric

PS: As I wrote this mail my qt install started to run correctly but very slowly. Looks like it won't be done for several hours.

Joern-R:
Hi Eric,

I also had problems with the Qt Online Installer - so I downloaded the offline installer from here: http://download.qt.io/official_releases/qt/5.12/5.12.1/ which made  did the job.  Overall compile time for GC went up with VS2017 - I am using the community edition and also checked on the VS2015 Compiler to be installed (on a Win10 - 1809 box).

I will have a look a the compiler options you mentioned to see if the behaviour changed. It's anyway weird the 32bit does not have the same behaviour like 64bit. For the official build I still use QT 5.6.x with VS2015 - just started first tests.

Will let know know further results - thanks for your efforts.
Joern

ericchristoffersen:
That installer was slow but worked.

I found this:

https://bugreports.qt.io/browse/QTBUG-72073

Which is pretty much exactly what we are seeing.
[QTBUG-72073] [REG 5.11->5.12] MSVC 64bit: Compile error C2666 when using std::bitset with QtCore - Qt Bug Tracker<https://bugreports.qt.io/browse/QTBUG-72073>
No reviews matched the request. Check your Options in the drop-down menu of this sections header.
bugreports.qt.io

Which lead to this bug opened against visual studio:

https://developercommunity.visualstudio.com/content/problem/406329/compiler-error-c2666-when-using-stdbitset-and-cust.html
Compiler error C2666 when using std::bitset and custom float class (64 bit) - Developer Community<https://developercommunity.visualstudio.com/content/problem/406329/compiler-error-c2666-when-using-stdbitset-and-cust.html>
We have fixed this issue and it's available in VS 2019 preview 1.0, please download it from here: https://www.visualstudio.com/vs/preview/ Thank you for your feedback!
developercommunity.visualstudio.com

Microsoft says they've fixed this for vs2019 but don't intend to fix in vs2017.

So we need to workaround it... I think best is to stop using bitset since I was the only person that used it so far.

I have attached a patch that replaces my use of bitset with a custom rolled 4 element bitset. I tested it.

Eric

Joern-R:
Thanks Eric,

I should have searched for the bug myself (my bad). I just can't see the attached patch. Can you please open a regular PR - I will take care on merging.

Cheers.
Joern

ericchristoffersen:
I'm not 100% on the whole pull request technology. Something about my github permissions. I'll try again today but for now I've attached the patch here. It is a patch to your branch.

[privatebitset.patch.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/2931361/privatebitset.patch.txt)

ericchristoffersen:
Changed ctor param from char to unsigned...

[privatebitset.patch2.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/2931400/privatebitset.patch2.txt)

Joern-R:
Thank Eric - patch is applied - no problems. Merge Done.

ericchristoffersen:
Thank you! Is great to be able to use vs2017.

Issue #3057, opened 2019-03-25
VirtualSpeed calculation produces nans on steep hills


ericchristoffersen:
I was trying to reconcile unusual numbers based on a virtual ride and the problem is the virtual speed calculation. It is a huge expression that only handles some of the cases for finding a cubic root. If you have a steep enough hill the equation's discriminant becomes positive and the expression produces NaNs.

I've reimplemented with a cubic solver and wrapped with the equations shared by:

http://kreuzotter.de/english/espeed.htm

I'd like to use this same expression in the nulldevice to compute speed so I can reasonably test activities created from null device (today speed is randomized independently of power and slope.)

ericchristoffersen:
Fixed checked in.

PR #3064, opened 2019-03-31
Issue 3057 virtualcalc produces na n on steep hills

ericchristoffersen:
Add cubic solver based on Blinn's fabulous paper. This is actually a solver for homogeneous 2d roots.

New physics equation with comments to compute velocity from input params (using cubic solver.) This is the instantaneous limit velocity and does not model inertia (which was true of old equation also.)

NullController now uses new equation so simulated trainer rides have reasonable speed so analysis tools can make sense of saved gpx.


liversedge:
looks like you replaced the entire source?
can you just check line ending settings in git?
also, can you squash/rebase before creating pull requests in future?

its difficult to review the actual changes made.

ericchristoffersen:
Hi Mark, sorry to be a bother here. I'm learning more git every day. I think I've correctly squashed the previous commits. Only caused 3 minor heart attacks and was a load of typing. I pushed a new squashed commit and it looks ok. I'm really glad there were no actual conflicts.

I still am not sure how to handle the line ending stuff. It probably doesn't help that I'm red/green color blind, probably the line ending errors are there in red and I just can't see them. What is it that you do to 'fix' the line ending errors? Is there somewhere I can read about what the actual issue is? I'm a bit confused about this because the existing code seems to be rife with lines with trailing whitespace. What exactly does the whitespace break?

WRT the changes: yes, the old speed calc is deleted. The previous virtualcalc compute was I think pretty obfuscated as to what it was doing, and it was slightly less complete. I redid the physics calc and commented it, and put the physical constants into what I hope becomes a common header for that sort of stuff. Most important the results are now obtained with an actual solver because a simple closed form is fundamentally incomplete (the source of the NaNs.)

I'm happy to keep adjusting this until youre happy, let me know what you want.

amtriathlon:
Fixed #3057, added for easier searches.

PR #3097, opened 2019-05-12
FitRideFile: Add Lezyne MegaXL

frnky:
I've just added the mega xl to the list of recognized devices.


grauser:
I modified the code according to Claus Assmann proposition.

Issue #2955, opened 2018-11-10
About the L/R Balance computation


jgpallero:
Hello:

The way GC computes the Left/Right Balance can be seen in the file src/Metrics/LeftRightBalance.cpp and looks like
```
while (it.hasNext()) {
    struct RideFilePoint *point = it.next();
    if (point->cad && point->lrbalance > 0.0f) {
        total += point->lrbalance;
        ++count;
    }
}
```
i. e. only data with cadence greater than 0 and lrvalue greater than 0 is used.

But this way can be problematic, at least in my case. I own a Favero Assiona DUO pedals and I've noted that when I'm not pedaling sometimes the value 100 is assigned to the lrbalance field instead the expected 0 value, so the average computation is slightly biased towards the left pedal. This assignation of 100 comes more or less randomly, as I have confirmed via programming with the Garmin SDK, and also in the Edit tab of GC (actually I've not made a deep research, so this is the reason I say "more or less randomly", but it is obvious that the 100 values must be addressed as 0 values). As sometimes the cadence measure have some lag there are data with lrbalance=100 with cadende!=0, and also data with lrbalance=0 with cadende!=0, which is the reason the code has the check lrbalance > 0.0f instead of only the if (point->cad). I don't know if this occurs with other power meters or with Favero Assioma DUO with other firmware than 3.0, but at least in my case I've noted this situation. So a more refined code must be in my opinion
```
while (it.hasNext()) {
    struct RideFilePoint *point = it.next();
    if (point->cad && point->lrbalance > 0.0f && point->lrbalance < 100.0f) {
        total += point->lrbalance;
        ++count;
    }
}
```
Other possible solution could be to check for power greater than 0 instead cadence or lrbalance. Or maybe a complete check like
```
while (it.hasNext()) {
    struct RideFilePoint *point = it.next();
    if (point->watts > 0.0f && point->cad && point->lrbalance > 0.0f && point->lrbalance < 100.0f) {
        total += point->lrbalance;
        ++count;
    }
}
```
in order to be absolutely sure. These additions have no side effects and work for any power meter.

Other question to consider, which I support, is whether other metrics like pedal smoothness average, torque efectiveness average and maybe others should be computed taking into account these lrbalance values, i. e., if these metrics should be take into account only valid lrbalance values.

[...]

PR #3098, opened 2019-05-12
Update LeftRightBalance.cpp

jgpallero:
Change in order to manage correctly the Left/Right balance due to the issues detected with the Favero Assioma DUO pedals, as it was explained here: https://github.com/GoldenCheetah/GoldenCheetah/issues/2955

But after I detected the issue, this file was modified in order to accomplish the proportion of Ground Contact Time from each leg for runs, and a new check was added to the code (point->cad || point->rcad), while in the original code only the point->cad was tested. I'm not sure if the addition of my new checks, specially the (point->watts > 0.0f) one, are safe for the running part of the function

### DIFF
--- a/src/Metrics/LeftRightBalance.cpp
+++ b/src/Metrics/LeftRightBalance.cpp
@@ -60,11 +60,10 @@ class LeftRightBalance : public RideMetric {
         RideFileIterator it(item->ride(), spec);
         while (it.hasNext()) {
             struct RideFilePoint *point = it.next();
-            if ((point->cad || point->rcad) && point->lrbalance > 0.0f) {
+            if (point->watts > 0.0f && (point->cad || point->rcad) && point->lrbalance > 0.0f && point->lrbalance < 100.0f) {
                 total += point->lrbalance;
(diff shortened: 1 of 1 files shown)


amtriathlon:
You are right, the power>0 condition will defeat the use for runs and it is, likely, overkill.

jgpallero:
There is then two options:

1. Check if the data come from bike or run and perform different checks, the running one without the power>0 condition
2. Make the check without the power>0 test, but including the point->lrbalance < 100.0f part. But in this case, although the L/R balance estimation is better, there is worse as including power>0, as it can be seen running the Octave/Matlab script I uploaded in https://github.com/GoldenCheetah/GoldenCheetah/issues/2955

amtriathlon:
It makes sense to check the presence of the base data to which lrbalance correspond, power or GCT, so the full condition could be:
`(point->cad && point->watts > 0.0f || point->rcad && point->rcontact) && point->lrbalance > 0.0f && point->lrbalance < 100.0f`
I'm not sure about the < 100 part.

PR #3099, opened 2019-05-13
FitRideFile: Unify handling of invalid product id

frnky:
@grauser 
Now all invalid product ids are handled similarly.
Moreover the "BSX Insight 2" probably doesn't belong to igpsport?


grauser:
Thank you

Issue #3102, opened 2019-05-15
About the average torque efficiency and pedal smoothness


jgpallero:
Hello:

This issue is related with the https://github.com/GoldenCheetah/GoldenCheetah/issues/2955 one, where a modification in order to correct the way of the L/R balance with the Favero Assioma DUO pedals.

I think that a change about the torque effectiveness and pedal smoothness must be made in the sense of check also the power, cadence and balance, as these metrics are directly related. In the current code it os checked if the TE and PS is greater than zero, but with the Favero Assioma pedals values of TE and PS greater than zero are stores although zero values for power are registered (see the attached screenshot).

I could made a pull request if this issue is considered
![favero](https://user-images.githubusercontent.com/857060/57810852-18a13400-7769-11e9-8bb0-416765f40903.png)

PR #3103, opened 2019-05-16
Update AvgLTE, AvgRTE, AvgLPS, and AvgRPS in BasicRideMetrics.cpp

jgpallero:
As for the L/R balance (see https://github.com/GoldenCheetah/GoldenCheetah/pull/3098) there are issues in the Torque Effectiveness and Pedal Smoothness computation. In some cases, values for LTE, RTE, LPS, and/or RPS are stored in the file while the power and/or cadence data shows that the correspondent stroke was either void or no power generating (see the attached image in https://github.com/GoldenCheetah/GoldenCheetah/issues/3102). This was not taken into account in the code, as the only check was the existence of LTE, RTE, LPS, and RPS data greater than zero. I've made the necessary changes in order to check if the stroke was a power generator by verifying if the cadence, power and lrbalance values are valid (it is SUPPOSED that all files that contain LTE, RTE, LPS, and /or RPS contain also power, cadence, and lrbalance data, as they are necessary for the LTE, RTE, LPS, and RPS values computation)

### DIFF
--- a/src/Metrics/BasicRideMetrics.cpp
+++ b/src/Metrics/BasicRideMetrics.cpp
@@ -2787,7 +2787,7 @@ class AvgLTE : public RideMetric {
     void compute(RideItem *item, Specification spec, const QHash<QString,RideMetric*> &) {
 
         // no ride or no samples
-        if (spec.isEmpty(item->ride()) || !item->ride()->areDataPresent()->watts || !item->ride()->areDataPresent()->lte || item->ride()->dataPoints().count() == 0) {
+        if (spec.isEmpty(item->ride()) || !item->ride()->areDataPresent()->watts || !item->ride()->areDataPresent()->lte) {
             setValue(RideFile::NIL);
             setCount(0);
             return;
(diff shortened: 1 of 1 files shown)


jgpallero:
I can see that an error occurs with the appveyor, but apparently there is nor related with any error in the src modifications

amtriathlon:
What’s the point of this condition?
`item->ride()->dataPoints().count() == 0`
It seems redundant to me

jgpallero:
I've copied the check from the similar statement in AvgPower, in the same BasicRideMetrics.cpp, where the checking for no ride or no samples is 

`if (item->ride() == NULL || !item->ride()->areDataPresent()->watts || item->ride()->dataPoints().count() == 0)`

In the case of AvgLTE, AvgRTE, AvgLPS, and AvgRPS spec.isEmpty(item->ride()) is used instead of item->ride() == NULL. I suppose the item->ride()->dataPoints().count() == 0 check control for a possible non null ride without data points (if it is possible), but I'm not sure if this situation can occurs

amtriathlon:
See https://github.com/GoldenCheetah/GoldenCheetah/blob/a86b0e9b74882ed83af3eb9293d97331ea95a500/src/Core/Specification.cpp#L84

jgpallero:
I've deleted the redundant condition

amtriathlon:
It's better to consolidate these changes in one commit before merge, for clarity and convenience in case they need to be reverted.

jgpallero:
Ups, sorry. It's my first contribution to a free software project, so I'm a bit confused about the procedure. Should I made or revert then any change?

amtriathlon:
No problem, you can rebase your branch to consolidate both commits in one and force-push or close this PR and make a new one.

PR #3105, opened 2019-05-16
iBike CSV: Import recorded Air Density series as XDATA. Fix Timestamp sync.  Fix XDATA import when DFPM=0

axl13:
This stream is useful to estimate instant CdA for a ride.


axl13:
Here's an[ example of the data](https://docs.google.com/spreadsheets/d/1EoO5Hf_WZHwUBme5crgml_yH3cW-vI0vOOLaJfMacLw/edit?usp=sharing):

PR #3107, opened 2019-05-18
FIT : Add Hammerhead as manufacturer

bstegmaier:
Added Hammerhead (manu id 289) to the list of recognised devices.

They currently do not set product ids as there is only one device (Hammerhead Karoo). However, this might change in the future...

Sticking to the invalid product id handling scheme suggested in PR#3099.


bstegmaier:
Hey @grauser, would you mind merging this or do you want me to change something? Thx ;)

grauser:
Yes, thank you.

PR #3110, opened 2019-05-25
Fix estimator not running for runs, if there are no bike rides

hver:
If there are no bike rides, the estimator:run() will exit. Instead of it, it should continue the loop to estimate runs


amtriathlon:
Thanks for the fixup, I think the empty estimate should be added at the end only when est.isEmpty() now

amtriathlon:
Thanks for the fixup!

PR #3111, opened 2019-05-26
Move tau setting from Athlete/Model to power zones settings

riccioclista:
W'bal Tau changes over time and should therefore be stored with the power zones for a given date.


amtriathlon:
Another reason to consider this is to extend W'bal to running power, there is a new paper discussing this application of recent publication:
[Vassallo, C., Gray, A., Cummins, C. et al. Exercise tolerance during flat over-ground intermittent running: modelling the expenditure and reconstitution kinetics of work done above critical power. Eur J Appl Physiol 120, 219–230 (2020). https://doi.org/10.1007/s00421-019-04266-8](https://rdcu.be/b0DAw)
This change is not enough, but it is necessary, IMHO.

amtriathlon:
I am closing this since it is likely overkill and it has a lot of unresolved conflicts.

PR #3112, opened 2019-06-01
Workout Editor: Support MRC flavoured qwkcode

riccioclista:
When loading an mrc workout file into the editor, FTP percentages are converted to absolute Watts in resp. to the current CP or FTP. This yields non-integer Watt values, which currently don't work as qwkcode (only integers are supported). We could make qwkcode understand doubles, but then: who creates workouts with fractional numbers?
To make it easier to read, it's better to stick to integers imho. Rounding the values to integers would result in significant rounding errors and thus is not a viable option. The only solution I see, is to support FTP/CP percentages natively in qwkcode.

What changes?
- Power numbers in qwkcode can now either mean Watts or percentage of FTP/CP depending on the format used (erg or mrc).
- A label above the code editor indicates which format is used.
- When creating a new workout one must specify the format


amtriathlon:
Thank you!

PR #3115, opened 2019-06-01
qwkcode: Remove trailing recovery block from interval sets

riccioclista:
When defining an interval set via qwkcode, a recovery interval is added to the end of the set. So for the specification 2x3m@200r1m@100 you would get

3m@200, 1m@100, 3m@200, 1m@100

This PR proposes to drop the trailing recovery interval for interval sets. The same specification would then result in

3m@200, 1m@100, 3m@200

### DIFF
--- a/src/Train/WorkoutWidget.cpp
+++ b/src/Train/WorkoutWidget.cpp
@@ -1885,8 +1885,55 @@ WorkoutWidget::qwkcode()
         for(int j=i+1; j<sections.count(); j++) {
             if (sections[j] == sections[i])
                 count++;
-            else
+            else {
+                // don't process any interval tails or such, if we don't have a set
+                if (count == 1)
+                    break;
+
+                // check whether j is the last interval of the set
+                bool isTrailingInterval = false;
+                QStringList section1Parts = sections[i].split('r');
+                if (section1Parts.length() > 1) {
+                    QStringList section2Parts = sections[j].split('r');
+                    QString section2Part1 = section2Parts[0];
+
+                    // if it has an r marker, remove it (only applies to one-part-only sections)
+                    section2Part1.remove('L');
+
+                    // compare the main interval parts of the sections
+                    if (section2Part1 == section1Parts[0]) {
+                        // it is indeed the trailing part of an interval set
+                        isTrailingInterval = true;
+                        count++;
+
+                        // now insert the recovery part (if there is one), so we don't loose it
+                        if (section2Parts.length() > 1) {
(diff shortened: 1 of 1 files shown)


liversedge:
I don't understand what you are trying to achieve.
If you don't want a recovery interval just dont use it, no need to remove the funcitonality?

riccioclista:
Sry, if I didn't express myself clearly. I like the functionality, I just like to change it, so that it doesn't create the last trailing part.

If I specify `3x3m@324r2m@250`, I'd prefer to get:
new
instead of the current implementation:
old

liversedge:
Ah I see. Did you check the code that determines the qwkcode from a hand drawn workout?

riccioclista:
The method where the qwkcode is generated is `WorkoutWidget::qwkcode()`. I've gone through it and altered the last part, where sections are composed to interval sets. I've documented each step thoroughly in code, but basically it handles 2 cases:

1. "Old" style interval sets are decomposed into the main part and the trailing bit. The trailing bit is inserted as a new line (section) after the main part in qwkcode. So, if we have a set like in the second picture posted above, this would become:
```
...
3x3m@324r2m@250
2m@250
...
```

2. If an interval set has been identified, it checks the next section, whether it contains a block that fits the last interval set. So, if we have something like in the first picture posted above, the current code will recognize the first 2 interval and recovery blocks as an interval set. The third, high watts block (+ recovery section after) will however be deemed to be part of a new section. With the new implementation, if this situation is encountered, the third section would be decomposed into 2 parts: The first block would be appended to the previous interval set; the recovery block would become a new section, inserted after the interval set.

The other place where changes where necessary, was the `WorkoutWidget::apply(QString code)` method, where changes in qwkcode are applied to the graph. When drawing an interval set, the last occurrence of a recovery block is now omitted.

riccioclista:
I've just added a commit that covers the edge case of an interval set being the last element of the workout. My previous change didn't cover post-processing of the last section in the workout, because the loop only ran to `j < sections.count()`, while I was referencing sections by `j - 1`. The loop now goes 1 element further, so that all elements are post-processed. Guards for IndexOutOfRange errors have been put in place.
If wished for, I can squash the 2 commits and force-push.

amtriathlon:
I think the currently implemented version is the more common interpretation, where the a rest interval follows each work interval, allowing better recovery monitoring.
Anyway, since it is a matter of preference, there has been no other requests for this feature in more than 2 years and introduces more code complexity, I will close it.

PR #3116, opened 2019-06-02
qwkcode: Express single interval-recovery section as 2 separate sections

riccioclista:
In the current implementation a section sequence like this:
single_int_rec_section
is expressed by the following qwkcode `2m@207r2m@134`. If this section is not part of an interval set, semantically it's not an interval + recovery period imho, but rather just 2 sections with different wattage. I'd therefore propose to express it simply as
```
2m@207
2m@134
```


amtriathlon:
Thanks for your contribution, but I think the added code complexity is not justified for what is like a cosmetic feature.

PR #3150, opened 2019-08-27
Allow defining time intervals in seconds for MRC workout format

Pi03k:
I've added defining interval times in seconds. I hope you find it helpfull.


ericchristoffersen:
The 'x' value on ergfile point represents time or distance, depending on the workout type.

Currently the 'x' field is a double so is not restricted to integer values, which means that seconds can already be represented, and there is no need to mark whether a time is in seconds or minutes.

amtriathlon:
I am closing this due to the comment above, thank you anyway.

PR #3163, opened 2019-09-09
Print Graphics and Reports 

Jroso:
Could you please let GoldenCheetah let you print the graphs and values ​​of each training in a customized report


amtriathlon:
Not a proper pull request, see https://help.github.com/en/articles/about-pull-requests

PR #3173, opened 2019-10-02
Add support for VO2 measurements and VMPro By VO2Masters

erikboto:
PR for tracking my work of adding support for VO2 measurements.


liversedge:
This looks really interesting. How close do you think it is to being ready?

erikboto:
The generic parts of adding the VO2 values I think is complete. For
the actual VM Pro device integration I think it should be mostly done,
but I want to get at least a few test sessions done to make sure I
haven't missed something. And there's some cleanup of qDebug stuff,
making sure all strings properly use tr() and having a fresh look at
all the comments so they make sense.

If life don't throw me curve balls though I hope I can get most of it
done this week.

erikboto:
So I believe this PR should be in a good state now. The functionality is in place and seem to work nicely now, and I removed a lot of the debug prints etc.

I left some changes to the existing qDebug lines, which adds printing deviceUuid() in addition to address(), so that those lines are useful both on mac and linux.

I think this would be a good time to review the changes to see if you want any changes made.

liversedge:
The patch looks clean and shouldn't bring in any regressions, couple of things:
1) can you add a .vo2 file to the tests/rides folder ?
2) can you squash and retain a commit message that covers all changes ?
3) have you tried using other BT40 devices at same time / after these changes applied ?

One quick Q, just for clarification:
4) How is the configurator widget used ? (I read the patch but didn't really grasp this)

erikboto:
> The patch looks clean and shouldn't bring in any regressions, couple of things:
> 
> 1. can you add a .vo2 file to the tests/rides folder ?
> 2. can you squash and retain a commit message that covers all changes ?
> 3. have you tried using other BT40 devices at same time / after these changes applied ?
> 
> One quick Q, just for clarification:
> 4) How is the configurator widget used ? (I read the patch but didn't really grasp this)

1,2,3 : Done now! 

4: It's a special widget that pops up as a separate window when a VM Pro device is connected. See screenshot.

vmpro-configuration

erikboto:
> I cannot test but it looks great in my opinion.
> The random generation of VO2 data in the Null Controller likely needs to be removed and agree to squash maintaining commit messages.
> Thanks Erik!

I removed the fake data from NullController, and the it's squashed into one commit now.

PR #3196, opened 2019-10-22
fixes #3193 (OSM MapWindow Marker)

### LINKED ISSUE #3193: Activity Start & End Markers out of position on Map view
I have just downloaded and installed GoldenCheetah_v3.5-RC1_64bit_Windows.exe
.
I looked at an activity to see if this version had solved my earlier raised issue on Route Segments.

On the Map view the route start and end markers are now out of position. That, the top left corner of the marker is at the intended position rather than the  pointer at the bottom of the marker.
image

sanfordhk:
OSM markers anchor defaults to top left but Google defaults to middle bottom. Hard coding marker size 16,37 for OSM.


sanfordhk:
Typo..

Issue #3211, opened 2019-11-13
Importing Garmin multisport fit files produces a single activity


thebaron06:
Importing a multisport fit file produced by a Garmin device causes GC to create only a single activity that has a wrong _Sport_ Tag. It gets the tag of the last extracted session. For Triathlons this is a _run_ activity.
This in turn leads to wrong analysis/metrics/...

This was already implemented and merged, see https://github.com/GoldenCheetah/GoldenCheetah/pull/1397.

The expected behavior is that GC splits the sessions included in a single multisport file into several activities.
Best case would be that there is a mechanism that allows to link them together.

amtriathlon:
IIRC that change was reverted due to its dependence on session header position, see https://github.com/GoldenCheetah/GoldenCheetah/commit/47d19d6f19c2062d21b0048c6b948c4b6c226363#diff-a6d4891b14953529b790a3fee31d80fe

amtriathlon:
Here you have a more detailed description of the problems with the code in #1397 and the reason it was reverted: https://github.com/GoldenCheetah/GoldenCheetah/issues/1424#issuecomment-118408146

thebaron06:
Oh, should have read the spec first. -.-
Thanks for the hints.

But, if we already know that Garmin files are working like this wouldn't it be a good idea to support at least files from them? We already know the vendor/product of the file. We can easily add a check in there to support only Garmin created files for this feature.
I do think this would fit almost all use-cases as the following statistics for sports watches used at ironman kona shows [[1](https://www.trinews.at/kona-count-compression-ist-out-garmin-dominiert/)] [[2](https://www.dcrainmaker.com/2017/10/insights-ironman-championships.html)]. I think it is needless to mention that kona is one of the most important event for any triathlete in the world.

---
2017
Garmin 1869
Polar 91
Suunto 20
Timex 13
Tomtom 9
Smart Watches 5

2016
Garmin – 1673
Polar – 85
Suunto – 47
Tom Tom – 24
Timex – 16
Fitbit – 1

thebaron06:
@amtriathlon can you tell me if there is any plan on what functionality should be supported for such a feature?
I had a quick look at the fit file spec and the activity spec mentioned earlier and I also had a look at the SDK code provided by Garmin.
[...]

PR #3212, opened 2019-11-13
WIP: fit file import: split multisport files into multiple activities

thebaron06:
Recently multisport files as produces by Garmin Fenix 3, some
Forerunners and so on are imported as a _Run_ activity, or whatever comes in as last session on parsing, that in turn is wrong.

This commit adds functionality that was already merged in https://github.com/GoldenCheetah/GoldenCheetah/pull/1397
This MR will split multisport fit files into multiple activities.
It was tested with a file recorded by a Garmin Fenix 3 HR that recorded a Triathlon.
I would declare it as WIP as long as it remains open if this has any side effects I am not aware of.

I will test this with more files produced by the same device and I will try to provide such examples.

My tests have already shown that we loose some device type information that was parsed for the multisport session embedded in the fit file. Also the HR records for the swim session did not show up, but this is a minor problem.

The major impulse for this MR was wrong metrics and analytics.

See issue https://github.com/GoldenCheetah/GoldenCheetah/issues/3211

### DIFF
--- a/src/FileIO/FitRideFile.cpp
+++ b/src/FileIO/FitRideFile.cpp
@@ -380,6 +380,7 @@ struct FitFileReaderState
                 case 3111: return "Garmin Fenix 5x +";
                 case 3112: return "Garmin Edge 520 +";
                 case 3113: return "Garmin FR945";
+                case 3126: return "Garmin Instinct";
                 case 3291: return "Garmin Fenix 6x";
                 case 20119: return "Garmin Training Center";
                 case 65532: return "Android ANT+ Plugin";


amtriathlon:
I tested minimally and this change fails to import the most basic single session FIT files for rides, runs or swims.

thebaron06:
Thanks for your feedback. I am wondering that this fails to do so because it only re-enables what has been done before.

But as discussed in https://github.com/GoldenCheetah/GoldenCheetah/issues/3211 I want to try to find a more general solution for this, therefore it stays WIP.

amtriathlon:
>I am wondering that this fails to do so because it only re-enables what has been done before.

Because you didn't include that part that was present in the original patch:
```
if (rides) 
            {
                // All our sessions are in the rides list, this one is not
                // needed.
                delete rideFile;

                return rides->first();
            }
            else
            {
                return rideFile;
            }
```

thebaron06:
> > I am wondering that this fails to do so because it only re-enables what has been done before.
> 
> Because you didn't include that part that was present in the original patch:
> 
> ```
> if (rides) 
>             {
>                 // All our sessions are in the rides list, this one is not
>                 // needed.
>                 delete rideFile;
> 
>                 return rides->first();
>             }
>             else
>             {
>                 return rideFile;
>             }
> ```

Hm, so I can't find any problem to the recent solution. I tried some single activity files and they import well. I had a look at the code handling the return params of this stuff and it is already aware of returning either a single file (rides ==nullptr) or multiple ride files (return param != nullptr && rides != nullptr). In latter case the returned ride file is not imported anyhow, it is ignored.

amtriathlon:
> Hm, so I can't find any problem to the recent solution. I tried some single activity files and they import well

If you have not found any issues with the current solution after adding that snippet, I would suggest some additional test cases to compare results with current master:
A lap swim file s.t. https://github.com/GoldenCheetah/GoldenCheetah/blob/master/test/swims/GarminSwimExample.fit
A swim with HRM s.t. https://github.com/GoldenCheetah/GoldenCheetah/blob/master/test/swims/GarminHRMswim.FIT
A ride with weather data s.t. https://github.com/GoldenCheetah/GoldenCheetah/blob/master/test/rides/Weather_XDATA_2016_02_06_13_01_22.fit
A run with Stryd data s.t. https://github.com/GoldenCheetah/GoldenCheetah/blob/master/test/runs/StrydPower-mutiple-float-values.zip

PR #3224, opened 2019-11-23
Fix 3222: Row append more efficient.

ericchristoffersen:
Fix 3222: Row append was doing incremental wildcard appends to main table string, causing QString.Arg to search entire string each time for wildcards. Is O(N^2) and caused big slowdown when lots of table rows. New code wildcard appends to rowstring, then += row's string when row is finished.


amtriathlon:
Much better Eric, thanks!

PR #3225, opened 2019-11-23
Strava data

qheath:
title and description for activities uploaded to Strava


amtriathlon:
Since Notes is one of the options for the name, I think we should check the user has not selected Notes for the name before to add Notes as description, to avoid duplication.

qheath:
removed the description code for now, only the charset fix is left

amtriathlon:
I tested and the first commit fixes a problem I had seen, some of my activities refused to upload, and the problem was accents in title. Description part also seems to work properly.
I will merge it, thanks!

PR #3228, opened 2019-11-24
3227: Further optimize trends tracker table display 

ericchristoffersen:
Additional speedups for trends tracker table display:

convert vectors to qlist before doing prepends.
Speedup test if unit is minutes: test static const hash instead of 10 string compares and translations.
Call arg on columnsummary to further reduce string that arg acts upon for wildcard substitution.

With these changes my largish table takes less than a second to display.

### DIFF
--- a/src/Charts/LTMPlot.cpp
+++ b/src/Charts/LTMPlot.cpp
@@ -4377,7 +4377,7 @@ void LTMPlot::refreshZoneLabels(QwtAxisId axisid)
 
 bool LTMPlot::isMinutes(QString units)
 {
-    static const QHash<QString, bool> MinutesHash = { {"minutes", true}, {tr("minutes"), true} };
+    static const QSet<QString> MinutesHash = { "minutes", tr("minutes") };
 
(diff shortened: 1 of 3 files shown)


src/Charts/LTMPlot.cpp
```diff
@@ -4377,5 +4377,7 @@ void LTMPlot::refreshZoneLabels(QwtAxisId axisid)
 
 bool LTMPlot::isMinutes(QString units)
 {
-    return units == "minutes" || units == tr("minutes") || PaceZones::isPaceUnit(units);
+    static const QHash<QString, bool> MinutesHash = { {"minutes", true}, {tr("minutes"), true} };
```
mtbkeith: nice solution

amtriathlon: Any reason to use a QHash instead of a QSet?

ericchristoffersen: > Any reason to use a QHash instead of a QSet?

Only my ignorance. Set is what I wanted. I'll change it.

PR #3235, opened 2019-11-30
Issue 3234: Speedup Load Time

ericchristoffersen:
Implement regex/hash based string substitution object to perform multiple substitutions in 2 passes. Speeds up athlete data load by 2x.

### DIFF
--- a/src/Core/RideDB.l
+++ b/src/Core/RideDB.l
@@ -44,13 +44,13 @@ static QString unprotect(char *string)
 
     // this is a lexer string so it will be enclosed
     // in quotes. Lets strip those first
-    QString s = string2.mid(1,string2.length()-2);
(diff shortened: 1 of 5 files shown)


amtriathlon:
TBH at first sight I didn't liked the added complexity but I gave it a try: my full history athlete with +6000 activities, which took a full minute to load the cache in a Windows 10 notebook with SSD and 16GB RAM, now loads it in around 30".

liversedge:
Crikey, had not realised how terribly slow this aspect was, I definiteley upvote this fixup !

ericchristoffersen:
Like shooting fish in a barrel. I changed to use QStringRef for substring before replace which was another 20% speedup. Added string handling cleanup to ridecache::save for 2x speedup to app exit.
Removed ifdefs. I think I'm done now with dumb string fixes.

PR #3242, opened 2019-12-06
Issue 3241: Fix Fix Power Spikes

ericchristoffersen:
Fix bug where variance criteria wasn't honored properly.
Add checkbox to permit function to interpolate slope,gear, cadence and speed for samples that have their power rewritten.

### DIFF
--- a/src/Charts/RideEditor.cpp
+++ b/src/Charts/RideEditor.cpp
@@ -600,8 +600,8 @@ AnomalyDialog::check()
     if (column >= 0 && rideEditor->ride->ride()->dataPoints().count() >= 30) {
 
         // get spike config
-        double max = appsettings->value(this, GC_DPFS_MAX, "1500").toDouble();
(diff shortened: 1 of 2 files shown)


amtriathlon:
Hi Eric, can you consolidate the changes in one commit to make easier to understand what it does?
GitHub squash merge can do the trick, but git rebase and a new PR will be more clear.
Thanks Ale.
PS: if it is a safe fix and it doesn’t include new strings to translate we could merge in v3.5, otherwise it would be better to pospone to v3.6.

ericchristoffersen:
> Hi Eric, can you consolidate the changes in one commit to make easier to understand what it does?
> GitHub squash merge can do the trick, but git rebase and a new PR will be more clear.

I went down the squash path but warn't pretty. Created a new PR.

> Thanks Ale.
> PS: if it is a safe fix and it doesn’t include new strings to translate we could merge in v3.5, otherwise it would be better to pospone to v3.6.

I think fix is safe and simple but no hard feelings if you want to delay to 3.6.

PR #3262, opened 2019-12-16
Issue 3261: Implement kinetic energy simulation

ericchristoffersen:
This implements kinetic-energy based bicycle simulation for trainer mode.

Is only active if check box is selected in trainer options.


ericchristoffersen:
Created new pull request against current head.

Issue #3261, opened 2019-12-16
Optionally use KE based speed when running in slope mode


ericchristoffersen:
I've noticed that I can do simulated rides on GC faster than I can do them in real life. I'm able to do more reported climbing per hour than in real life.

Root cause is that the training mode is using the speed reported by the trainer. The training mode then increments distance per speed and then derives altitude from location. It isn't even remotely suspicious of whether my paltry watts could reasonably achieve the ride.

It makes sense that reported speed isn't right for these rides. The trainer, at least my kickr, doesn't know most things that it would need in order to report a 'correct speed' based upon slope:
- my weight
- bike's weight
- crr
- frontal area
- wheel/drivetrain inertia
- etc

Instead implement kinetic energy simulation that determines speed bases on energy added (pedaling and hills) and subtracted (friction and hills) during each tick.

This feature is enabled via a checkbox in trainer options.

PR #3263, opened 2019-12-16
Issue 3261: Implement kinetic energy simulation

ericchristoffersen:
I've noticed that I can do simulated rides on GC faster than I can do them in real life. I'm able to do more reported climbing per hour than in real life.

Root cause is that the training mode is honoring the speed reported by the trainer. The training mode then increments distance per speed and then derives altitude from new location along path. It isn't even remotely suspicious of whether my paltry watts could reasonably achieve the ride.

It makes sense that reported speed isn't right for these rides. The trainer, at least my kickr, doesn't know most things that it would need in order to report a 'correct speed' based upon slope:
- my weight
- bike's weight
- crr
- caa
- wheel/drivetrain inertia
- etc

This change implements a kinetic energy simulation that determines speed bases on energy added (pedaling and hills) and subtracted (friction and hills) during each tick. Is currently setup to use an advanced numerical integrator to determine velocity and derive change in distance.

This feature is off by default and only enabled in slope mode via a checkbox in trainer options.


ericchristoffersen:
Second commit moves all the bicycle simulation properties into a new tab under athlete-trainer settings.

amtriathlon:
Hi Eric, we are near to v3.5 release and I think this feature could be merged after release to be part of v3.6.
In the meantime a very simple enhancement could be to use existing VirtualSpeed when Speed is not available and it would not require changes to UI, what do you think?

ericchristoffersen:
The virtual speed is the computed speed at the limit so it ends up pretty massively under and overestimating speed on climbs. For gpx rides the slope is pretty noisy which makes the speed crazy and leads to inappropriate progress up the hill. I was surprised to see i was climbing even faster. I went down the route of smoothing slope before i realized that whole approach should be energy based.

I also looked at relying on speed not available to tell when to compute it and the setup was impossible because options are ignored when there is a single device:

Ive been using this a few weeks on pc and really like it but it needs some bake time across platforms before we know it works across toolchains.

On my private branch ive also integrated peter brandts old rlv synchronization and fixed some video control issues and the virtual rides are starting to work pretty well.

Would be keen to do a big drop at the start of 3.6.




amtriathlon:
Agree, thanks for the detailed answer.

ericchristoffersen:
I think this is ready to go.

amtriathlon:
Linked issue #3261

PR #3275, opened 2019-12-27
Fix Issue 3273: Busy Hang in FixDeriveDistance

ericchristoffersen:


src/FileIO/FixDeriveDistance.cpp
```diff
@@ -164,21 +166,19 @@ FixDeriveDistance::postProcess(RideFile *ride, DataProcessorConfig *config=0, QS
                 }
 
                 if (ii >= ride->dataPoints().count()) {
-                    // Past end of ride points, continue pushing final point.
-                    RideFilePoint *pii = ride->dataPoints()[ii-1];
+                    RideFilePoint *pii = ride->dataPoints()[goodii];
```
amtriathlon: If there are no good points this will crash or is there other conditions that avoid to reach this code?

ericchristoffersen: You are right. I thought I'd initialized goodii to 0.

PR #3284, opened 2020-01-05
Issue 3283: Use tangent vector to compute interpolated grade.

ericchristoffersen:
This change has interpolated slope come from instant interpolated tangent vector instead of averaging slope across nearby samples. Along with correcting slope when riding ergfiles this massively improves the mergefile behavior in some cases.

Rename location interpolation method "Interpolate" to be "Location".

Change ErgFile::LocationAt and GradientAt to accept location as double instead of int. Location reporting was jumping all around when multiple gpx files were in a small area (interpolation now interpolates instead of snapping to nearest meter.)

This change helps refine training mode behavior. Should go in along with bicycle sim KE change.

### DIFF
--- a/src/Core/BlinnSolver.cpp
+++ b/src/Core/BlinnSolver.cpp
@@ -19,6 +19,28 @@
 #include "BlinnSolver.h"
 #include "cmath"
 
+// 0 == A*x + B
+Roots LinearSolver(double A, double B)
+{
+    if (A == 0) {
+        if (B == 0) return Roots({ 0,1 });
+        return Roots();
+    }
+
+    return Roots({ -B / A, 1 });
+}
+
+// 0 == A*x^2 + B*x + C
+Roots QuadraticSolver(double A, double B, double C)
+{
+    if (A == 0) return LinearSolver(B, C);
+
+    double det = sqrt(B*B - 4 * A*C);
+    double r0 = (-B + det) / (2 * A);
+    double r1 = (-B - det) / (2 * A);
+    return Roots({ r0, 1 }, { r1, 1 });
+}
+
 // Cubic solver as described by James F. Blinn's EPIC paper:
 // 'How to Solve A Cubic Equation', all 5 parts are awesome
 // but part 5 provides the algorithmic conclusion that I used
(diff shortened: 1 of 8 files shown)


ericchristoffersen:
Well... change is cool and correct but it extremely sensitive to having good gps points. I'm finding most gps data have stepwise altitude so route slope has too much variation. I am going to write a parameterized gps route smoothing filter (maybe based on NURBS.) This change should be considered only after there's a good way to smooth gps routes.

amtriathlon:
No problem, I tagged this for 3.6 but I will not merge it until is ready!

ericchristoffersen:
Gpx is set of locaiton points of route. When gpx is loaded the average slope is computed between the points. When gps data is noisy the averaging will make very extreme slopes, sometimes oscillating quickly and always discontinuously.

These extreme slopes are much higher than a trainer can honor, and occur much faster than a trainer can respond, so these slopes are mostly not felt by the rider. The new KE based trainer code does honor the extreme slopes so the rider has their speed penalized but they don't get to appreciate the true horror of the climb because they won't experience correct pedal resistance.

This altitude noise is the root cause of a lot of pain when riding gpx files so I think we should offer a way to nicely smooth gpx route data - especially the altitude data where a few vertical meters in a short span can make a big difference. I have something working nicely using b-splines but haven't integrated into gc.

Current GC only allows to query slope from known points on the route.

This change introduces the ability to query gradient for all locations on the interpolated route spline, and trainer mode then uses that data to update slope more frequently. The current interpolation is cubic splines which have continuous derivatives so the slope changes will be smooth and correctly integrate over distance to the correct altitude. In practice with a nicely smoothed gpx this works really well.

The downside of this change is that when gps points oscillate wildly the cubic spline can make the extreme slopes even worse.

ericchristoffersen:
I found the root cause of the slope fluctuations... the cubic interpolator is only correct when samples are uniformly spaced. Most of the time it doesn't matter but causes significant artifacts when computing slope with non-uniform samples. A bunch of my ride files work great but a few have intereresting artifacts.

I've implemented an update to the cubic interpolator so it will be correct with non-uniform sample spacing. It cleans up all the artifacts. Good news is that all users of cubic splines will only benefit.

Anyway... please wait for another commit to fix the non-uniform sample spacing issue.

liversedge:
similarly is this ready for merge?

ericchristoffersen:
Just wrote above. I am working on another commit for this.

ericchristoffersen:
Ok, this is good to go.

PR #3298, opened 2020-01-16
Remove ampersand from menu and button text before compare

### LINKED ISSUE #1852: Addview not working
Under dev GoldenCheetah, on Linux, Qt5.5.1 the Add Chart menu from the main menu doesn't work for me. The small toolbar menu option does still work.

On my machine/config it appears that this is because the call to text() in MainWindow::addChart is returning a string with accelerator markup. (i.e. it returns "&Workout" rather than "Workout")

`if (GcWindows[i].name == action->text()) {`

Replacing this with a call to toolTip() instead returns the correct text and fixes the issue for me since this defaults to the original message text unless explicitly set.

`if (GcWindows[i].name == action->toolTip()) {`

### LINKED ISSUE #2930: Cannot Close RideImportWizard Dialog
This was tested on Linux with Qt5.11.1 compiled from master, but originally also found from the 3.5 tag also compiled with the same settings.
Importing Activities the Abort and Finish button always runs Save as the QString comparison of the dialog text includes an & in front of the text which is not present from the string returned by tr(). Because of this the dialog can never be closed and the program must be killed to remove the dialog.

amtriathlon:
To avoid issues with kde injecting them, it is not pretty
but simple and safe.
Fixes #1852
Fixes #2930

### DIFF
--- a/src/Charts/HomeWindow.cpp
+++ b/src/Charts/HomeWindow.cpp
@@ -188,9 +188,11 @@ HomeWindow::rightClick(const QPoint & /*pos*/)
 void
 HomeWindow::addChartFromMenu(QAction*action)
 {
+    // & removed to avoid issues with kde AutoCheckAccelerators
+    QString actionText = QString(action->text()).replace("&", "");
     GcWinID id = GcWindowTypes::None;
     for (int i=0; GcWindows[i].relevance; i++) {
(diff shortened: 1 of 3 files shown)


amtriathlon:
Included in v3.5 Strava reissue.

Issue #1536, opened 2015-08-24
slope calculation is broken (and as the result, the power estimation too)


invy:
As you can on the screenshot, here is a climb (start: 47.996155 11.486262, end: 47.999702 11.498074). Distance 1.1km, height difference 73m, in avg slope should be 6% (at some places should be as high as 16%)

Goldencheetah shows -0.0 almost everywhere.

(Built from revision: 5d0a910)

![slope-0](https://cloud.githubusercontent.com/assets/833865/9453905/35e5d466-4abe-11e5-88ec-5266d5e2504d.png)

amtriathlon:
That slope channel is computed by GC or is present as data in the imported file? Can you post the original file for testing?

invy:
The slope channel is computet by GC. Here is gpx file:
https://github.com/invy/stuff/blob/master/climb1.gpx

amtriathlon:
Yes, I found the same, the smoothing algorithm with the -20/+20 cap doesn't work with this altitude data, OTOH if you run Fix Elevation Data the results seems more razonable:
![climb1](https://cloud.githubusercontent.com/assets/1444784/9482950/53e1b770-4b70-11e5-8bf6-c66e10e19fe9.png)
Trying to do this I found slope derived channel is not recalculated after Fix Elevation Data so I needed to remove the slope channel and restart GC to produce the above chart...

invy:
I've been playing around with slope computation. I could get somewhat reasonable results, using exponential smoothing.
![slope_power_exp_smoothing](https://cloud.githubusercontent.com/assets/833865/9562652/6590fd14-4e72-11e5-8fe6-e4849ea058b3.png)

liversedge:
can you provide a patch?

invy:
In general yes, however I have some questions regarding how it should be Implemented in GC.

At the moment in my 'proof of concept' I've simple added a menu entry "Edit->Fix slope errors" and modify elevation values. However adding new menu entries could be not the best approach. Eventually it would make more sense, if the smoothing is be done during importing. What do you think?

Here is my 'proof of concept':
https://github.com/invy/GoldenCheetah/compare/09099704c58e332e9540e5c33acc0d5c9a9ca5d1...1d689f1cf54e505d8646ad27ba160211a17c14e4

amtriathlon:
I think we could subsume this in https://github.com/GoldenCheetah/GoldenCheetah/issues/3302

PR #3310, opened 2020-01-25
Issue 3302: BSpline altitude smoothing to FixGPS.

ericchristoffersen:
This change adds altitude smoothing to the FixGPS dialog. Eventually I would like to extend this to also smooth location info as well but altitude is a priority in order to make gpx files ridable in training mode because I'm finding it is very noisy compared to lat/lon data.

For bspline implementation I gathered the code I needed from the GeometricTools library from https://www.geometrictools.com and pasted as a single file: GeometricTools_BSplineCurve.h. The geometric tools library is distributed under the boost license.

In order to support testing the smoothing setting before application I extended the ManualDataProcessor object to let me access the ride file.

recalculateRideFile function was not recomputing slope if slope existed. I changed it to always compute slope.

### DIFF
--- a/src/FileIO/FixGPS.cpp
+++ b/src/FileIO/FixGPS.cpp
@@ -22,7 +22,50 @@
 #include "HelpWhatsThis.h"
 #include <algorithm>
 #include <QVector>
+#include <QFormLayout>
+
 #include "LocationInterpolation.h"
+#include "GeometricTools_BSplineCurve.h"
+
+using namespace gte;
+
+bool smoothAltitude(RideFile *ride, int degree, double &minSlope, double &maxSlope, bool fDoRewrite);
+
+void ComputeRideFileStats(const RideFile * ride, double &minSlope, double &maxSlope, double &sampleDistanceVariance)
+{
+    minSlope = 0;
+    maxSlope = 0;
+    sampleDistanceVariance = 0;
+
+    const RideFilePoint * prevPi = NULL;
+    double prevKM = 0;
+    for (int i = 0; i < ride->dataPoints().count(); i++) {
+        const RideFilePoint * pi = (ride->dataPoints()[i]);
+
+        // Compute slope since ridefile slope is truncated.
+        double slope = 0;
+        if (i > 0) {
+            double rise = pi->alt - prevPi->alt;
+            double run = 1000. * (pi->km - prevPi->km);
+            slope = run ? (rise / run) : 0;
+            slope *= 100.;
+        }
+
+        if (slope < minSlope)
+            minSlope = slope;
+        if (slope > maxSlope)
+            maxSlope = slope;
+
+        double kmDelta = pi->km - prevKM;
+        prevKM = pi->km;
+
+        sampleDistanceVariance += (1000 * kmDelta * 1000 * kmDelta);
+
+        prevPi = pi;
+    }
+
+    sampleDistanceVariance /= ride->dataPoints().count();
+}
 
 // Config widget used by the Preferences/Options config panes
 class FixGPS;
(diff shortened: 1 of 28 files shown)


src/FileIO/FixGPS.cpp
```diff
@@ -31,25 +74,160 @@ class FixGPSConfig : public DataProcessorConfig
...
+
     public:
-        // there is no config
         FixGPSConfig(QWidget *parent) : DataProcessorConfig(parent) {
 
             HelpWhatsThis *help = new HelpWhatsThis(parent);
             parent->setWhatsThis(help->getWhatsThisText(HelpWhatsThis::MenuBar_Edit_FixGPSErrors));
-        }
 
+            // Should we consider passing ride data to all DataProcessorConfigs?
```
amtriathlon: No, RideItem already provides access to underlying RideFile, see https://github.com/GoldenCheetah/GoldenCheetah/blob/8a94666e91140d3b60537609c129ddb81c057a3d/src/Core/RideItem.h#L61

src/FileIO/FixGPS.cpp
```diff
@@ -31,25 +74,160 @@ class FixGPSConfig : public DataProcessorConfig
...
+            //
+            // Steal ride from ManualDataProcessorDialog. This is needed to test
+            // smoothing settings prior to application, so user can see what final
+            // min and max slopes will be.
+            ride = ((ManualDataProcessorDialog*)parent)->Ride()->ride();
```
amtriathlon: This will crash GC if this data processor is configured to run automatically, see https://github.com/GoldenCheetah/GoldenCheetah/issues/3311

ericchristoffersen: Thanks! I had no idea.

src/FileIO/FixGPS.cpp
ericchristoffersen: I mean should config be given explicit access to the RideItem on init? Casting to parent type to obtain the RideItem is sort of hokey. Is reasonable for lots of dataprocessor cases to see some results of the action before applying it. For now I am going to change the Ride access to return a const RideItem, config will be allowed to read RideItem but not to write it.

src/FileIO/FixGPS.cpp
amtriathlon: I was wrong, sorry.
The problem I was referring to doesn't happen here, but in postprocess were config is NULL when the data processor is called automatically, see https://github.com/GoldenCheetah/GoldenCheetah/commit/e64b6fb557090bd95296fab037fed771b28dc02f

ericchristoffersen: Right, I understand. I'm updating to support the import case.

src/FileIO/FixGPS.cpp
ericchristoffersen: Ok, latest commit the config is provided with a ridefile when possible, otherwise its null.

src/FileIO/FixGPS.cpp
ericchristoffersen: Ok, all better with latest commit. Even a ui in the fixgps processor settings row.

amtriathlon:
Hi Eric, this would need rebasing.
Linux and macOS travis-ci builds are failing due to the change in DataProcessorConfig constructor signature, likely AppVeyor builds are passing since they don't have Python Embedding enabled.

amtriathlon:
I did some minimal testing using this file https://raw.githubusercontent.com/invy/stuff/master/climb1.gpx from https://github.com/GoldenCheetah/GoldenCheetah/issues/1536, which seems to be a very bad case, and I can see the changes when I apply the fix.
What doesn't seem to be working is running on import, I enabled it with the checkbox activated and it doesn't change anything on import.

ericchristoffersen:
I just saw this for myself. I accidentally disabled when config was null. I'll update.

ericchristoffersen:
Ok. Works on import now. For that file I think first step is the "Fix Elevation Error" which recovers new altitude data from internet, then run the smoothing.

PR #3314, opened 2020-01-26
WIP: Fit multisport split support

thebaron06:
see #3212 and #3211 

This is another attempt to support multisport files to be imported and split up in their single sport activities.
Recently there are some things that have to be checked, but the data imported matches them Garmin Connect shows for my files.
Needs further testing!


thebaron06:
```
// NOTES:
        // - there are two temperature values (avg/max) - temperature and temp - where temp is correct and
        //   temperature seems to be the overall (avg/max)
        // - start altitude of first transition not correct (zero), leads to too high climb figure
        //   i think this is not really a big deal yet
        // - some metrics - like Gear Ratio for transitions - maintain a steady state due to auto-fill-up
        //   treatment. This does not really reflect the real world but imho its ok for now and more of
        //   a general problem
        // - it might be a good idea to treat transitions as a run activity
        // - devices are not added to session files
        // - laps are not imported correctly
```

thebaron06:
Laps are working now, rebasing this asap.

thebaron06:
should be rebase on master now

thebaron06:
As i have no clue what I was doing wrong I will re-open this with a much more cleaner branch.
The new branch is base on upstream/master, as this should be...

PR #3315, opened 2020-01-26
New Device IDs

MeisterP:
Add device IDs for Garmin Edge 830 and for Stages Cycling power meters


amtriathlon:
Can you add a test file?

MeisterP:
> Can you add a test file?

This is not a pristine Edge 830 file since I've stripped some data for privacy reasons. However, it does works as a test for the newly added device IDs

PR #3317, opened 2020-01-27
Support Multisport Fit Files

### LINKED ISSUE #3211: Importing Garmin multisport fit files produces a single activity

thebaron06:
This patch series will introduce support for splitting multisport fit files into their contained single sport activities. Eg. a multisport recorded triathlon file will split up into 5 files:
 - swim
 - transition
 - bike
 - transition
 - run

Laps and XData are adjusted and appended too, if present.

Note:
 - Recording devices are merged throughout the parsing process and therefore the whole set of devices will be appended to each session split.
 - This will change the default sport/subsport of activities because otherwise it is just guessing and will influence metrics calculation. The default changed to generic.

Fixes #3211
Note: #3212

### DIFF
--- a/src/FileIO/FitRideFile.cpp
+++ b/src/FileIO/FitRideFile.cpp
@@ -144,8 +144,15 @@ struct FitFileReaderState
     XDataSeries *deveXdata;
     XDataSeries *extraXdata;
     XDataSeries *hrvXdata;
-    QMap<int, QString> deviceInfos;
-    QList<QString> dataInfos;
+    QMap<int, QString> deviceInfos, session_device_info_;
+    QList<QString> dataInfos, session_data_info_;
+
+    // cache tags for each session
+    using SessionTagMap = QMap<QString, QVariant>;
+    SessionTagMap active_session_;
+    QList<SessionTagMap> session_tags_;
+    QList<QMap<int, QString>> session_device_info_list_;
+    QList<QList<QString>> session_data_info_list_;
 
     FitFileReaderState(QFile &file, QStringList &errors) :
         file(file), errors(errors), rideFile(NULL), start_time(0),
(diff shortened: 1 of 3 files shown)


liversedge:
I agree with Ale- do not use standard tags for specific purposes. By all means create a new one

thebaron06:
As requested the change was squashed into a single commit, at least the multisport part.
I added another commit that belongs to the XData classes and allows for deep copies. This one is sepperated.

The change now remove the debug output, introduces a new Tag called _BaseFileName_ that should represent the base filename the multisport activity is split from.

Device info is split too.

But I do not understand what you mean with the following
> the change of source filename tag to anything different from that

amtriathlon:
Please ignore the comment bout source filename, the mangling I was referring to is done by RideFileImport, not your code.

thebaron06:
I exported a couple of years of fit files from my archive and other platforms and all names have been unique across ~3k filenames given them by different Garmin devices.
But I fully agree that one shouldn't rely on this to be unique.

The aim of setting the tag was not to uniqually identify sthg. It was ment to display the fact that this activity originates from a multisport file and to have some kind of relationship between the activities that are part of the series.
If there is a better way to do this I will be happy to add this.
Is there another proposal to display that this ?

Btw. Data info is now splitted too. Have you spotted any other data that is candidate to be split up?

amtriathlon:
>The aim of setting the tag was not to uniquely identify sthg. It was ment to display the fact that this activity originates from a multisport file and to have some kind of relationship between the activities that are part of the series.

Yes, I agree, but being that identifier unique makes it easier to get all the components of a multisport activity with a simple filter.

>If there is a better way to do this I will be happy to add this.
Is there another proposal to display that this ?

The start date in ISO format is a good candidate, it is warranted to be unique in GC.

>Btw. Data info is now splitted too. Have you spotted any other data that is candidate to be split up?

No, all data and metadata seem to be ok as far I have tested.

thebaron06:
> > The aim of setting the tag was not to uniquely identify sthg. It was ment to display the fact that this activity originates from a multisport file and to have some kind of relationship between the activities that are part of the series.
> 
> Yes, I agree, but the uniqueness of this multisport ID would make it easier to get all the components of a multisport activity with a simple filter, without the need to look for collisions.

I agree on that too.

> > If there is a better way to do this I will be happy to add this.
> > Is there another proposal to display that this ?
> 
> The start date in ISO format is a good candidate, it is warranted to be unique in GC and readily available.

Thanks for the hint. The latest version uses this to identify files that belong together by their origin.

> > Btw. Data info is now splitted too. Have you spotted any other data that is candidate to be split up?
> 
> No, all data and metadata seem to be ok as far I have tested. Good job!
Thank you for your review and continuous work on this great project, too.

As mentioned above, the idea is to add a new Tag - _Multisport_. The presence of this tag indicates that there have been several sessions splitted from an single activity file. The value of this tag is the ISO formatted start time of the base file the sessions originate. This should allow for filtering/grouping files that belong together.

thebaron06:
Just saw that I forgot to add the above mentioned code segment. Will fix this later.

Issue #3320, opened 2020-01-27
FixGPS: Add ability to prune outliers


ericchristoffersen:
I found that adding ability to filter outliers allows better smoothing. Propose to extend ui to support an outlier criteria. Smoothing would work as:
1) build b-spline like today
2) compute variance between bspline and input, exclude outliers
3) build and interpolate final result using bspline made without outliers.

ericchristoffersen:
Here's an explanation of the new smoothing behavior.

I have a gpx route that I've downloaded from a routing site on the interwebz. It has a ton of altitude noise and also some kooky route points where route goes back on itself for no reason.

The fixgps is updated to have have two phases, each potentially building a pair of splines.

The first phase is altitude smoothing, the second is route smoothing. They are separate because altitude needs very different smoothing than the route. You can run either or both from the config dialog but if you run both the altitude will run first. The test stats button simulates what the current settings will do. If neither is set then current ride stats are computed and displayed.

Altitude noise is a more significant problem for virtual ride because the altitude noise is felt in your pedals and can make you suffer the wrong amount of pain. We want slope to be smooth and continuous and sensible like a mountain road. If route lat/lon location is a few meters to the left or right it just hurts how your route appears on a map but a meter up and down in the middle of a climb will cause your trainer to behave really strangely.

It is also important to smooth altitude first because the second route-smoothing phases sees the route as a 3d spline. If the noisy altitude data is not smoothed first the altitude changes will cause the outliers and generally you will not get results.

Each phase, both altitude and route smoothing consist of an outlier pass and then a final smoothing pass. The outlier pass is designed to discover and remove outliers so that the final smoothing pass can run without them. The outlier pass builds a b-spline with pass1 degree, then outlier is anything that differs from the smoothed spline by more than the variance criteria. The smoothing pass builds a b-spline with pass2 degree with the outliers removed, and rewrites the ride file to match that bspline. Generally you will want a high degree (smoother) bspline to discover outliers, then use a smaller degree to build a bspline to rewrite your ride.

The config allows you to perform both smoothing phases in one 'click', but I've run them separately here to show how they work.

Here is an example:

Here is my route's noisy altitude data. The extreme slope variance is clear.

![smoothing0](https://user-images.githubusercontent.com/46055145/73397567-ae013b80-4298-11ea-8e88-2b8b0ea7694c.JPG)

Run fixgps and get this dialog:

[...]

PR #3321, opened 2020-01-27
Issue 3320: FixGPS: Route Smoothing and Multipass Outlier Removal

ericchristoffersen:
Update fix gps to support outlier removal.


ericchristoffersen:
Hang on, I'm going to update this with the full route smoothing business.

ericchristoffersen:
Added an optional final processing that smooths route in xyz. Altitude and route smoothing both now support multipass smoothing where outliers are discarded on first pass.

liversedge:
is this now ready for merge?
IIRC a previous update crashed when set as a dataprocessor, have you fixed that?

ericchristoffersen:
This change is working great. Import works, no crash.

amtriathlon:
Fixed https://github.com/GoldenCheetah/GoldenCheetah/issues/3320

PR #3324, opened 2020-01-31
Add support for Ergofit Bike

mayerflash:
I've added support for a protocol used by Ergofit bikes with a serial interface.
It was tested on a "Ergofit Home 400", but should work on other Ergofit bikes as well.

That's my first contribution to GoldenCheetah, so if I made something wrong just tell me.


liversedge:
This looks like a good patch, will wait for CI to complete then merge.
Thank you for the patch !

mayerflash:
Great! Thank you!

PR #3332, opened 2020-02-03
Derived stats for sim bicycle config screen

ericchristoffersen:
I forgot I'd stashed these changes. Should have been part of the recent ke changes. This change causes the simulated bicycle config screen to show derived stats while setting numbers. Main thing is to show the kinetic energy mass (mass + equivalent mass) but also shows the values for wheels.

### DIFF
--- a/src/Gui/Pages.cpp
+++ b/src/Gui/Pages.cpp
@@ -2296,6 +2296,68 @@ SimBicyclePage::AddSpecBox(int ePart)
     m_SpinBoxArr[ePart] = pSpinBox;
 }
 
+void
+SimBicyclePage::SetStatsLabelArray(double d)
+{
+    double riderMassKG = 0;
+
+    const double bicycleMassWithoutWheelsG = m_SpinBoxArr[SimBicyclePage::BicycleWithoutWheelsG]->value();
+    const double bareFrontWheelG           = m_SpinBoxArr[SimBicyclePage::FrontWheelG          ]->value();
+    const double frontSpokeCount           = m_SpinBoxArr[SimBicyclePage::FrontSpokeCount      ]->value();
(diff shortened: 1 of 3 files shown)


liversedge:
The dialog reminds me of the pied piper ux, most users will struggle to understand most of this.
The philosophy in GC is to try and hide complexity until the user needs it. Also, it feels a little over engineered to me-- did you have another use case in mind when developing this e.g. solving for optimal pace for TT ???? might be worth looking at that?

liversedge:
pied piper ux :)
piedpiper

ericchristoffersen:
Let me know what you want.

I was hoping this ui was hidden enough. I think it is satisfying to be able to model my own bike and tires. Default is some modern tdf thing.

I am extremely happy with the new ke behavior, it is feeling perfect to me.

At home now I am actually am modeling my loaded Bikepacking setup and getting an idea of time to different parts of some himalaya climbs and what gears I want. Am thinking now I need to model power loss due to altitude...

WRT ui here: What would you suggest?

Tt pace: A precise race tt planner is doable but I suspect ‘good enough’ would require very precise user parameters for their power capabilities and aero. Maybe user could provide upper and lower bounds for w’ and how their aero really changes for different speeds? Feed into aero lab? Math doesn’t sound difficult.

WRT stats: I think it is interesting to see exactly how bike changes help and hurt climbing time. Whether those $5 spokes are a good value and how much pain my 1.5kg tires are actually causing.

liversedge:
ok .. so I guess if you're working on sim you need parameters. My feeling is mist users will be lost. but then they can just accept defaults.

Is this ok to merge?

ericchristoffersen:
Yes, safe to merge.

Understand this doesn't add more parameters, it adds display of some derived values. The displayed derived numbers are pretty helpful. I did it because there are so many and I was otherwise having to use a spreadsheet to add everything up and validate it.

Thanks for looking.

Issue #347, opened 2012-11-12
TTS file import


TriScott:
Has any work been done to allow direct import of the workout files from Tacx Trainer Software?  I know you can export from TTS as Polar .hrm files which can be imported, but being able to natively load the .tts format would be more convenient.

Part of the reason I ask is because I actually put a decent amount of work into cracking the .tts format myself previously.  I'd say I am 90% of the way there (stopped when I had the ability to extract the power data which was all I really cared about).  If the feature appeals to anyone I could probably help.

drockmd:
Any updates on this? I do not see the ability to import TTS files into Golden Cheetah in the current release. 

I recent switch to Wahoo, because of Tacx's terrible customer service, but have a ton of historical ride data and purchased courses still on the tacx platform that I would like to import into a program compatible with the kickr. 

Any info would be super helpful.

Thanks, 
Derek

mk4th:
Hi @TriScott ,
Do you still have your TTS decode somewhere?
I want to implement the convertor.
Thanks

ericchristoffersen:
Hi. I just started working on this. Can we please reopen? Some of my tts files have location data, some others don't. Would need it to be tested by a broader audience. Maybe have something to play with in 1 or 2 days.

liversedge:
send a note to the forums to request some, you may get a fair few.

ericchristoffersen:
I think this finished and ready to go. Would love to hear feedback from folks that can compare with the silo'd tacx experience.

PR #3333, opened 2020-02-06
Support for TTS files.

ericchristoffersen:
Add ability to read tts files into ergfile format for trainer mode. I've tested only with the few old tacx tts files I own.

Training mode now can record altitude even if lat/lon are not present (some older tts files.) This is done by deriving altitude from slope and distance.

Fix numerical stability issues in cubic and quadratic solvers caused when one coefficient dominates the others.

I was surprised to see the altitude data in my tts files is very noisy which leads to noisy slope. It is slightly distracting but for me it isn't a huge deal. An option we have is to ignore the altitude data in the tts and instead derive it from the tts' slope. Downside of this is that the altitude won't sum to the correct altitude for the ride, but their canned slopes are very smooth.

### DIFF
--- a/src/Core/BlinnSolver.cpp
+++ b/src/Core/BlinnSolver.cpp
@@ -16,28 +16,99 @@
 * Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#include <cmath>
+#include <algorithm>
 #include "BlinnSolver.h"
-#include "cmath"
+
+int GetExponent(double a) {
+    int exp;
+    std::frexp(a, &exp);
+    return exp;
+}
+
+template <typename T>
+int MaxExponent(T a) {
+    return GetExponent(a);
+}
+
+template <typename T, typename ... Args>
+int MaxExponent(T a, Args ... args) {
+    return std::max(MaxExponent(args...), GetExponent(a));
+}
+
+// Returns true if value is so small compared to args that it is effectively zero.
+// Very important that near-zero values are detected. Zero is defined relative to how large
+// the use-expression's other operands are.
+// The solver will be subtracting and losing precision. If one term is near zero the result
+// will be noise since result has no effective precision. Much better results if you disregard
+// a term with near zero coefficient and instead use the next lower order solver.
+template <unsigned T_MantissaBitsNeeded, typename ... Args>
+bool RangedZeroTest(double value, Args ... args) {
+    const static int s_ExpLimit = std::numeric_limits<double>::digits - T_MantissaBitsNeeded;
+
+    // Easy yes if value is actually zero.
+    if (value == 0.) return true;
+
+    int vExp = GetExponent(value);
+    int maxExp = MaxExponent(args...);
+    if (maxExp - vExp > s_ExpLimit)
+        return true;
+
+    return false;
+}
+
+// Define zero as 10 or fewer bits of mantissa precision shared between value and operand(s).
+// Less than that we have no precision. Relative size must differs by 2^(53-bits)
+const static int s_MantissaBitsNeeded = 10;
+
+template <typename ... Args>
+bool IsZero(double value, Args...args) {
+    return RangedZeroTest<s_MantissaBitsNeeded>(value, args...);
+}
 
 // 0 == A*x + B
-Roots LinearSolver(double A, double B)
-{
-    if (A == 0) {
-        if (B == 0) return Roots({ 0,1 });
-        return Roots();
+Roots LinearSolver(double A, double B) {
+    if (IsZero(A, B)) {
+        if (B == 0) return Roots({ 0,1 }); // choose 0
+        return Roots(); // no root
     }
 
     return Roots({ -B / A, 1 });
 }
 
 // 0 == A*x^2 + B*x + C
-Roots QuadraticSolver(double A, double B, double C)
-{
-    if (A == 0) return LinearSolver(B, C);
+Roots QuadraticSolver(double A, double B, double C) {
 
-    double det = sqrt(B*B - 4 * A*C);
-    double r0 = (-B + det) / (2 * A);
-    double r1 = (-B - det) / (2 * A);
+    // If A is zero then linear.
+    if (IsZero(A, B, C)) {
+        return LinearSolver(B, C);
+    }
+
+    // Make monic.
+    B /= A;
+    C /= A;
+    A = 1;
+
+    // No C term means a quick and accurate factor.
+    if (IsZero(C, B)) {
+        return Roots({ B, 1 }, { -B, 1 });
+    }
+
+    // Double root case.
+    double det2 = B * B - 4 * C;
+    if (IsZero(det2, B * B, 4 * C)) {
+        return Roots({ -B / 2, 1 });
+    }
+
+    // Negative determinate means no roots.
+    if (det2 < 0) {
+        return Roots();
+    }
+
+    // Otherwise standard double roots.
+    double r0 = (B + std::copysign(std::sqrt(det2), B)) / -2.;
+    double r1 = C / r0;
+
     return Roots({ r0, 1 }, { r1, 1 });
 }
 
@@ -60,10 +131,12 @@ Roots QuadraticSolver(double A, double B, double C)
 // To obtain solution for x with w == 1 simply divide
 // x by w.
 Roots
-BlinnCubicSolver(double A, double B, double C, double D)
-{
-    if (A == 0)
-    {
+BlinnCubicSolver(double A, double B, double C, double D) {
+    // Take care to detect near zero A coefficient since
+    // it can destroy precision and cause solver to produce
+    // the same result over and over. Much better to use 
+    // quadratic solver.
+    if (IsZero(A, B, C, D)) {
         return QuadraticSolver(B, C, D);
     }
 
@@ -100,8 +173,7 @@ BlinnCubicSolver(double A, double B, double C, double D)
             At = A;
             Cbar = d1;
             Dbar = -2 * B * d1 + A * d2;
-        }
-        else {
+        } else {
             // Depress params for algorithm 'D/A'
             At = D;
             Cbar = d3;
--- a/src/FileIO/JsonRideFile.y
+++ b/src/FileIO/JsonRideFile.y
@@ -627,7 +627,8 @@ JsonFileReader::toByteArray(Context *, const RideFile *ride, bool withAlt, bool 
             if (ride->areDataPresent()->cad && withCad) out += ", \"CAD\":" + QString("%1").arg(p->cad);
             if (ride->areDataPresent()->kph) out += ", \"KPH\":" + QString("%1").arg(p->kph);
             if (ride->areDataPresent()->hr && withHr) out += ", \"HR\":"  + QString("%1").arg(p->hr);
-            if (ride->areDataPresent()->alt && withAlt) out += ", \"ALT\":" + QString("%1").arg(p->alt);
+            if (ride->areDataPresent()->alt && withAlt)
+			    out += ", \"ALT\":" + QString("%1").arg(p->alt, 0, 'g', 11);
             if (ride->areDataPresent()->lat)
                 out += ", \"LAT\":" + QString("%1").arg(p->lat, 0, 'g', 11);
             if (ride->areDataPresent()->lon)
--- a/src/FileIO/LocationInterpolation.cpp
+++ b/src/FileIO/LocationInterpolation.cpp
@@ -16,6 +16,7 @@
 * Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#include <assert.h>
 #include "LocationInterpolation.h"
 #include "BlinnSolver.h"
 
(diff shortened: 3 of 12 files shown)


ericchristoffersen:
Second commit takes care of the problem with bad altitude data. ErgFiles from tts have their altitude data declared "untrustworthy". This means that train mode will just use the sample slopes instead of computing it from altitude. Restructured the altitude, slope and location grovelling to handle all the different cases.

Another fix: dont let sim bicycle roll backwards. Negative v goes to zero.

At this point the behavior is pretty good with my TTS files. Needs bake time.

ericchristoffersen:
#347

ericchristoffersen:
With last commit I've given up on the tts altitude info. It is simply too noisy. Now the tts loading always recomputes altitude based upon the gradient recorded in the tts file. Thus the altitude in the loaded file wont match reality but it will match the tts gradient, and metrics like VAM will be valid when riding tts files.

Ramoonus:
if you need help to test, i`ve got some TTS4 movies to verify on my W10 PC

ericchristoffersen:
> if you need help to test, i`ve got some TTS4 movies to verify on my W10 PC

Thanks Ramoonus. That would be great. I'm keen to hear what issues are present, especially how gc experience compares to the authentic tacx experience.

The ttsfile is loaded to an internal erg file which is then used for the training ride.

The ergfile from tts will always have altitude. This is a number that is faked using the gradient info present in the tts. If there is no location data then the altitude will always start at zero. For rides without locaiton data I'm keen to know if tacx has stored the stashed a starting altitude somewhere...

For tts with location data I am still deriving altitude from gradient. I do this because the altitude in the tts files I've seen is really terrible and doesn't match the tacx gradient. I figure it is better to show the actual vertical meters that you rode on your trainer instead of some fantasy number from the tts file.

The ergfile playback in gc always derives gradient from a continuous curve through the derived altitudes. This means gc will display a continuously changing gradient. Is that also how tacx works? Do you like it?

I have made no changes in how golden cheetah synchronizes distance to video. I would love to know if you think there is a bug here. Several sorts of bad synchronization I'd like to hear about:

sync always off by a constant offset
sync gets worse as ride goes on
sync varies through ride

ericchristoffersen:
Just found a really wicked error. The ttsreader was computing altitude from ride distance and gradient. But tangent vector during ride was computed from location on path spline constructed from location. The tangent vector gets stretched and squished when there is a big difference between ride distance and geo location... Fix is to compute altitude change using location distance delta, not ride distance delta.

ericchristoffersen:
Lot of good stuff here. Only issue I can see is a slight slope instability (+/- 0.2) on some switchbacks on the stelvio.

There are still the open questions of if there's a way to synchonize altitude on no altitude files and if the starting distances are correct. The mountain passes look pretty well synchronized to me.

ericchristoffersen:
Ok. Slope is now perfect. I had neglected to account for when parametric speed was different than the average. Tried a bunch of rides and they all seem to behave well.

I think this change is ready to go.

liversedge:
Perfect. Can you add a short description to the commit, nothing fancy just a brief description, a couple of lines will do. Then will merge.
PS: Make lines no longer than 50 characters in length.

ericchristoffersen:
Description: will do... I noticed something after doing a ride yesterday and have another small change that looks good.

Here is the current state, when loading erg file I spherical interpolate the duplicate points. Pretty much a linear interpolation.

slerp

I changed the duplicate interpolation to use cubic splines instead:

cubicInterp

In no way does this effect ride effort (doesnt change gradient or distance) but looks lots better I think. Given how noisy and inaccurate the gps data can be I literally have no trustworthy points so I'm tempted to do a bspline averaging instead, but that'll be another change.

So anyway, I will play with this new hotness a bit and re-push when ready. Will add a description at that point.

liversedge:
Why is altitude encoding in json changed?

I'm not sure that is relevant to TTS support -- we are super sensitive to changing the file format and encoding -- it generally requires a lot of testing. Was it in the patch previously? I don't recall seeing it.

ericchristoffersen:
I did just change altitude precision with this latest commit. This is the same issue I had before with lat/lon. The altitude store in json hasn't enough resolution. Currently only 2 digits. It makes the slope very rough and ugly on steep switchbacks when I'm riding slope. Slope is now beautiful with high resolution. Is this actually considered a file format change?

ericchristoffersen:
Here is same switchback on stelvio with the current 2 digits:

altitude_two_digits

And here is with change:

altitude_more_digits

ericchristoffersen:
My memory paged in. The issue is that the default fp write precision for the stream is only 2. The altitude is currently being written as float but only 2 digits. After the json change its being written as float with unbounded precision - exactly the same as lat/lon.

liversedge:
The fact you didn't need to change the parser on reading suggests this is ok. I just feel a little uncomfortable with it in this commit. But lets be pragmatic about it. I'm not gonna make you do a separate patch :)

Think its ok, seems churlish otherwise.

Sigfreid74:
@ericchristoffersen  - Great work on this project - I am very interested in being able to make the most of my old tts files and videos. Did you end up getting RLVstudio and RLVworkbench? I have copies if you still needed them.

ericchristoffersen:
> @ericchristoffersen - Great work on this project - I am very interested in being able to make the most of my old tts files and videos. Did you end up getting RLVstudio and RLVworkbench? I have copies if you still needed them.

Thanks for feedback. I'd love to hear how this tts support compares with what tacx app does. My tts files are working well for me but haven't heard anything critical (and I like criticism.)

I have pretty much given up on trying to produce rlv (and tts), I wrote something that can make an rlv but the editing experience is just really poor, and the math is indeterminate. I think the rlv format being speed/based is really unfortunate and hacky. They are closed formats and producing them is quite involved. The old rlv/pgmf system sucked because there were two files that had to be kept in sync, the newer tacx tts format is much better because there is only one stream of data: distance is the single axis which provides slope, location and video time, but tts is quite involved to write, I think lots of unnecessary fiddly bits.

So instead of requiring tts golden cheetah now supports playing saved json activities directly. To make a videosync ride you do the outdoor ride recording video and gps data. You can edit the ride as an activity using the spreadsheet style edit window. Because you recorded gps and video at the same time their time is synced. Edit the ends of the ride so the start/end times match up (easy if you always start video before gps record, end gps record before stopping video). If you have video sections you want to remove (like a rest stop), then edit them out in video editor, making careful note of start/end time of section you removed, now in golden cheetah edit window remove the samples that correspond to that time segment. Voila the time and distance and location remain synced. I've tried this for a few play rides, just to prove it, but I don't have a reasonable system for recording (I mostly mountain bike.)

Sigfreid74:
Hi Eric,
Thanks for getting back to me. I currently trying to get my old "non-smart" Tacx Genius trainer (T2020) to work with a different project (FortiusANT) so I can then connect up to Golden Cheetah and the like. (that team have successfully been able to hook up a "non-smart" Tacx Vortex. ) I am also wondering if you have a program that converts the TTS to RLV/PGMF or even GPX? Is your GC TTS file reader able to output the file to JSON or  GPX?
Overall I am hoping to future proof the day when TTS4 or the computer it is on fails me....

Thanks

Sigfreid74:
Hi Eric - The FortiusAnt group has finally cracked the Genius signals - and Golden Cheetah can now recognise the non-smart Genius as an FE+C trainer. I can now run my old tts videos on GC - It works thanks to your code! Do you know of a way to create segments of videos within the GC system - ie just the climbs for example?

ericchristoffersen:
That is great news!! Such a waste to have these old trainers gathering dust.

I'm so happy you are using the tts support. The tts unpacking was all thanks to the WattZap project on github, most of my work was just porting it from java to c++.

Segments of video: that sounds like something you should suggest on the golden cheetah user group.

Sigfreid74:
Yes it will make a lot of use of some pretty good hardware out there. A lot
of people are of course interested in Zwift - but I love the RLV's. The tts
files have courses (usually the climbs built into them. No way to access
them in GC? I'll ask the user group.
Thanks Eric.

ericchristoffersen:
Look what an informed user can achieve! Having never used the tacx software I had no idea there were 'courses' defined within the tts file. That sounds like another 'issue' you should bring up in the user group. Sounds neat.

There are a few tokens in the tts that we currently ignore. "Segments". It could be they are the info you want.

Can you please tell me a bit about courses? What does it achieve? Also please point me at a tts file that has 'segments'? And tell me where those segments lie on the course? Then I can figure out how to get that same data from the tts.

Sigfreid74:
OK, Thanks Eric, I'll get that to you in a day or so. In a longer video such as the "Route de Grand Alps" which may contain 3 or 4 climbs, the courses are usually the individual climbs. As far as I can tell there is no way to make smaller segments inside GC when loading a .tts file.

ericchristoffersen:
I don't really understand what the purpose of the 'sections' or 'courses' is, but if you think its neat explain why and maybe someone will add it. I'd like to understand the appeal.

Right now every press of the fwd and back buttons in train mode teleports you by 1km. If there's a section I like (the middle switchbacks on the stelvio) then I will just click back 5-6km to repeat that part. Similarly I often skip long descents by clicking ahead.

Sigfreid74:
Hi Eric,
Here is a collection of screenshots from tts4 with the courses / segments shown. As you can see on a long video like this one it is very handy if you want to quickly setup a ride for just one of the climbs (usually enough for me in one session). [https://drive.google.com/drive/folders/1Zsaf1wepWio0_GUjBqhb2JU9PxFXQjVu?usp=sharing](url)

ericchristoffersen:
Excellent. I had no idea. I never bothered to implement the string decoding. I just tried it out and indeed there are segments with the names. Its just like you said.

I'm not sure whatall we'll do to make these things accessible from train mode, there's nothing in the ergfile datastructure, nor in the ui that would let you choose the segments, but I might as well enable the full decode from the tts reader.

Here's an example:

[ segment name ]
[ "Route 1/8 - Gorges du Cians" ]
149490  [ 0x7ff7dd2d6640 149490 0x7ff7dd2d5770 ]:  0 . 1041  v 1000   10 x 1 
block type  1041  version  1000 
[segment range]  0 - 21.7388 / 1

For now I will scoop these all into a segment list in the tts reader.

ericchristoffersen:
Ok, I have submitted a change to do the parsing and bubble route title and description to the ergfile. At last the route name is now visible while riding. The segments and segment descriptions are parsed now but there's nowhere to put them. Baby steps!

#3668

Sigfreid74:
Brilliant Eric! Can I do anything to help / test changes?
Matthew

ericchristoffersen:
> Brilliant Eric! Can I do anything to help / test changes? Matthew

Unfortunately the empirical different with this change is to just see the route title while training. Not so exciting. But rest assured the datastructure inside now sees all the segments, etc. There's still quite a bit of work to be done to make it actionable. How exactly should we change the ui so that these segments can be used? The problem is that, as far as I know, its just tacx' tts that have these segments, so we can't turn the world upside down just for that.

It was super useful for me to see the screenshots. I've never seen the tts software.

Probably the most useful thing you could do is to document the stuff you want to see and explain it on the user group. Sell it. Then someone might step up and make the ui changes that are needed.

Sigfreid74:
Hi Eric, I will try and put a proposal forward. I do have another question though; when selecting train mode, what information is coming from the "workout" file vs the "videosync" file. Can you select different Workout files for any given videosync file?  Or do the titles have to match? Could a "workout" file be created to refer to certain segments of the videosync and avi files?
Matthew

ericchristoffersen:
File purposes: easiest explanation is what isn't in the workout file.
The movie is... the movie. It maps time to a picture.
The videosync file is a list of number pairs that are used to map distance to time.

With videosync file and a movie you can map distance to a picture. Yes?

The workout is everything that isn't the movie and the videosync file.

Workout files must contain distance and then either wattage or slope. That is the minumum. Sometimes workout files include time and they can be used as videosync files (because you can map distance to time with them.)

You are free to select any combination of workout, movie and videosync. If the distance or times don't match then the workout will end when you run off the end of one.

Custom Videosync to ride a piece of a ride. Sure, you could do that. Simply map distances to different times. You can map distance to movietime however you like as long as distance and time are both monotonic. We don't support reverse movie playback :).

But if you had a perfect workout slopes and videosync for a movie, I wouldn't touch it. Much easier to just ride the distance range that you want to ride (like how segments are described in the tts.) Would be simple to just say: "I want to ride from 100 to 150km", than to muck about making new videosync, there is just no point.

Perhaps the easiest thing to do for segments:
-fix how location is chosen, it isn't reasonable to click fwd 150 times to go ahead 150km. Maybe we could increase the step size as clicks are made quickly.
-it'd be nice to see the segment descriptions, maybe on the videoscreen, as you cross their boundaries.
- maybe have a tile that shows the description too...

I'm closing my eyes and imagining and I don't really imagine having segments available to choose would make my virtual rides better. I think I'd rather just click to where I want to ride and then ride it, instead of picking something first and needing to stick to it.

Perhaps there should be some sort of meta ride that drives the selection of workout, video,videosync, and it could be set with any sequence, including segments within a ride, and also to chain together other rides.

But I'm not really sold on it for what I do. Wrong place to discuss. Bring this to the user group!

Sigfreid74:
OK Eric, I have taken the plunge and tried to describe my "ideal" GC training setup on the google user group. Thanks for the detailed info, I understand a bit better now. Another thought: One other thing a segment could do would be to display distance to the top of a climb as a training incentive?

ericchristoffersen:
Distance to top of next climb... that is a cool idea! Maybe we could base climb upon the uci definition of climb and show the ranking. Sort of an auto-segment feature. Love it. I think you should open a new issue with this suggestion.

Hey the change to put the ride segments into the training mode is partly done now. You'll see lap boundaries at the segment boundaries. But the names are really hard or impossible to read and there's no way to select them by name. Try it out and then open issues about the things you think are broken.

THannoverS:
I've just installed 3.6-dev , but am unable to locate this exciting tts-import feature  .  
Currently i've looked after it 
- in activity/import from file .. but this do no list .tts as supported formats, and the impoter chokes if I select my .tts   from the wildcard *.* option.  (my tts file  is straight out of TTS4).
- in tools/ import workout , but that one 'chokes' on the tts-files as well

What am I missing out here ?

PR #3349, opened 2020-02-17
Optimize splines and bikesim dvdt.

ericchristoffersen:
Optimize cubic interpolator coefficient processing. Lazy coefficient processing is 4-10x faster. This has no effect on values.

Update bicycle sim dvdt calculation to return direct slope instead of interpolated slope. 2x faster and more accurate.

### DIFF
--- a/src/FileIO/FixGPS.cpp
+++ b/src/FileIO/FixGPS.cpp
@@ -672,52 +672,168 @@ bool IsReasonableGeoLocation(geolocation *ploc) {
              ploc->Alt() >= -1000 && ploc->Alt() < 10000);
 }
 
+class SaveState
+{
+    double m_t;
+    double m_step;
+    double m_achieved;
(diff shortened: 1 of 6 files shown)


ericchristoffersen:
Issue is here:

#3350

ericchristoffersen:
Travis failed because of some quadtree damage. Needs restart.

ericchristoffersen:
Pull is still pending so added more stuffs. Was smoothing a gpx with 480k entries. Found profound issues with file that size, mostly about detecting and dialing back the gradient descent if it runs of the rails.

This change also dials back distance length compute to no longer measure to the nanometer.

liversedge:
I don't feel confident in assessing this code, but since its isolated I'm happy to merge.
Are you happy with it at this point?

ericchristoffersen:
Yes. Good to merge. I'm happy with it.

PR #3352, opened 2020-02-19
Remove CRS_LOC

ericchristoffersen:
Remove CRS_LOC because it is just noise and adds no value.


ericchristoffersen:
Hello? Any objection to this?

PR #3354, opened 2020-02-20
Allow TTS Files for VideoSync.

ericchristoffersen:
Newer tacx dvd don't include rlv, the video sync info is in the tts. This change lets TTS act as videosync.

I tried with a few files and synchronization is ok but not the laser precision I was expecting. But I don't own any newer dvds. I'd love feedback from people who do.

### DIFF
--- a/src/FileIO/LocationInterpolation.cpp
+++ b/src/FileIO/LocationInterpolation.cpp
@@ -394,7 +394,6 @@ void GeoPointInterpolator::Push(double distance, double altitude)
 
     geolocation geo(0, 0, altitude);
     xyz point = geo.toxyz();
-    point.y() = distance;
 
     DistancePointInterpolator<SphericalTwoPointInterpolator>::Push(distance, point);
 }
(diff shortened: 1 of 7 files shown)


ericchristoffersen:
#3354

ericchristoffersen:
With latest commit I rewrite the original wattzap stream merging. The problem is that there is information in 3 arrays which can be totally unsynchronized. The previous code would always use the videosync data as the basis and then snap the other data into whatever was the point with the closest distance. Problem is that the amount of data in each array is independent. New approach uses basis with most entries, then cubic interpolates the other data onto the basis array. Other change is that while accumulating gps points I discard those that are duplicates of the prior point. This removes clusters of duplicate points which lets interpolator do a better job spreading the data out by route distance. Was able to remove the late gps de-duplication pass. There are still some tacx rides that are not perfectly synced but is getting better.

ericchristoffersen:
Hello? What can I do to help get this close?

PR #3360, opened 2020-02-26
Issue 3359: Increase time before considered fishy.

ericchristoffersen:
Merge ride file was rejecting all samples with time greater than 25 hours.

With this change I was able to merge all gpx for a 52 day trip... And result looks good.


liversedge:
I am not going to accept this at this time.
We have a limit of 24hrs for any activity.

ericchristoffersen:
What is the reason for the 25 hour limit? Is there something else not working?

PR #3370, opened 2020-03-08
CloudDB sharing of Custom Metrics

### LINKED ISSUE #3361: CloudDB sharing of Custom Metrics
Custom metrics can be exported and imported with recent commit, this is to also support upload and download to the cloud DB.

amtriathlon:
Upload/Download are similar to Export/Import in User Metrics config dialog.
Admin interfases are similar to CloudDB charts for both user (edit/delete)
and curator (set curated/edit/delete).
Name and Description are initialized to the corresponding items for the metric
but both can be edited to provide different/more complete information for the
CloudDB interfase.
Fixes #3361

### DIFF
--- a/src/Cloud/CloudDBUserMetric.h
+++ b/src/Cloud/CloudDBUserMetric.h
@@ -58,10 +58,8 @@ public:
     bool getUserMetricByID(qint64 id, UserMetricAPIv1 *usermetric);
     bool deleteUserMetricByID(qint64 id);
     bool curateUserMetricByID(qint64 id, bool newStatus);
-    void incrementDownloadCounterByID(qint64 id);
     bool getAllUserMetricHeader(QList<CommonAPIHeaderV1>* header);
 
-    void updateUserMetricInCache(qint64 id);
     bool sslLibMissing() { return noSSLlib; }
 
 public slots:
--- a/src/Cloud/CloudDBUserMetric.h
+++ b/src/Cloud/CloudDBUserMetric.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2015 Joern Rischmueller (joern.rm@gmail.com)
+ * Copyright (c) 2020 Ale Martinez (amtriathlon@gmail.com)
  *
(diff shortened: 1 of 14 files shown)


src/Cloud/CloudDBUserMetric.h
```diff
@@ -0,0 +1,246 @@
...
+
+    CloudDBUserMetricClient();
+    ~CloudDBUserMetricClient();
+
+    bool postUserMetric(UserMetricAPIv1 usermetric);
+    bool putUserMetric(UserMetricAPIv1 usermetric);
+    bool getUserMetricByID(qint64 id, UserMetricAPIv1 *usermetric);
+    bool deleteUserMetricByID(qint64 id);
+    bool curateUserMetricByID(qint64 id, bool newStatus);
+    void incrementDownloadCounterByID(qint64 id);
```
Joern-R: The DL Counter is not implemented in the User Metric API - and I also do not see a lot of need to do so. Even for Charts nobody actually looks at the data. So please remove the function and the call. Thanks.

src/Cloud/CloudDBUserMetric.h
```diff
@@ -0,0 +1,246 @@
...
+    bool getAllUserMetricHeader(QList<CommonAPIHeaderV1>* header);
+
+    void updateUserMetricInCache(qint64 id);
```
Joern-R: The function is both for UserMetric and for GChart not implemented. Looks like an idea from the past which was never required. So please remove in both objects (to avoid confusion). Thanks..

src/Cloud/CloudDBUserMetric.h
amtriathlon: I will remove them, thanks!

src/Cloud/CloudDBUserMetric.h
amtriathlon: I will remove them from both objects, thanks!

amtriathlon:
> Please have a look at my 2 comments.

Done, thanks for the review!

Issue #3373, opened 2020-03-15
Allow FixGPS to smooth altitude when there is no lat/lon present


ericchristoffersen:
This thread a user has a workout that needs altitude smoothing when there is no lat/lon data.

https://groups.google.com/d/msg/golden-cheetah-users/5d2YunLhpW0/y3nx06MNBwAJ

Currently FixGPS refuses to run in less there is lat/lon data.

Issue is fixgps should be able to perform altitude smoothing even when there is no lat/lon data present.

amtriathlon:
Complement of https://github.com/GoldenCheetah/GoldenCheetah/issues/3320

PR #3374, opened 2020-03-15
Issue 3373: FixGps to run on altitude alone.

ericchristoffersen:
Also prevent lots of mac build warnings by nulling the silent debug macros in geometrictools header.

### DIFF
--- a/src/FileIO/FixGPS.cpp
+++ b/src/FileIO/FixGPS.cpp
@@ -1096,55 +1096,61 @@ bool FixGPS::postProcess(RideFile *ride, DataProcessorConfig *config, QString op
     Q_UNUSED(config)
     Q_UNUSED(op)
 
-    // ignore null or files without GPS data
-    if (!ride || ride->areDataPresent()->lat == false || ride->areDataPresent()->lon == false)
-        return false;
+    if (!ride) return false;
 
-    // Interpolate altitude if its available.
     bool fHasAlt = ride->areDataPresent()->alt;
+    bool fHasLoc = ride->areDataPresent()->lat && ride->areDataPresent()->lon;
+
+    // We can operate on alt and/or loc, but cannot operate if there are neither.
+    if (!fHasAlt && !fHasLoc)
+        return false;
 
     int errors=0;
 
     ride->command->startLUW("Fix GPS Errors");
 
-    GeoPointInterpolator gpi;
-    int ii = 0; // interpolation input index
+    // Default behavior: Interpolate missing points.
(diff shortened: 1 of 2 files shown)


ericchristoffersen:
#3373

ericchristoffersen:
Was failing to smooth high frequency noise. Need to be able to specify variance threshold lower than 1. Changed outlier spinboxes from QSpinBox to QDoubleSpinBox.

Original slope data:

precisionnoise

After smoothing with 0.01 outlier variance (and degree 20):

fractionalvariance

ericchristoffersen:
Any objection? Someone take a look? Thanks.

amtriathlon:
Hi Eric, the feature is interesting, but I think it is not intuitive for users to have it as an option for Fix GPS when there is no GPS data. May be Fix Elevation is a better place for this?

ericchristoffersen:
> Hi Eric, the feature is interesting, but I think it is not intuitive for users to have it as an option for Fix GPS when there is no GPS data. May be Fix Elevation is a better place for this?

It could be. The filtering is pretty complex, you really need the stats and the test button in order to find a good choice. When filtering location data the altitude filtering needs to be available here as a first pass. It needs to be taken into account before smoothing the actual location spline. The combined stats really help.

I agree discoverability is a problem. I think it'd be ugly and unfortunate to put the altitude smoothing in two places. Was frustrating for me in a few cases where this altitude-only ability wasn't possible.

How about we rename "FixGps" into "Filter Route Data"/"Repair Route Data" something like that?

The FixElevation is a very different sort of repair because it queries a map service's opinion. In the many cases the internet map data is the source of the noise that needs filtering.

Maybe Fix Elevation should be renamed to something like "Repair Route using Map Service."

I dunno... I'm all ears let me know what you'd like.

PS: Other aspect of this change is to allow sub-integer variance. That was essential in order to smooth fine noise.

PPS: In a future checkin I'd like to change the use of variance in the ui to stddev. Much more stable and intuitive number. From math perspective they have the same capability but advantage of stddev is that it could actually have units for altitude. For location it wouldn't have a consistent unit because ECEF... :(

I'm waiting for this change to clear before I start that next one.

amtriathlon:
Agree “Fix Route” may be a better name for the enhanced Fix GPS.
Also Fix Elevation could be renamed, since what it does is to fetch elevation from an external service, just “Fetch Elevation” could work, the source likely be explained in description.
This is just my opinion, I prefer @liversedge to have the last word on this.

PR #3377, opened 2020-03-20
Issue 3376: FIT files as workout.

ericchristoffersen:
Issue 3376. Add ability to use FIT file as workout.

This is trivial since both GPX and FIT are loaded to ridefile and parsegpx merely populates ergfile from ridefile. We can now easily support training with any other file types that RideFile can load.

My test case exposed errors in location interpolation where it would not tolerate immobile point data, causing slope and distance to become NaN.

Small change to improve slope calc precision.

Main problem was an existing issue in ErgFile::parseGpx where file was being opened once in parse and again in ridefile instantiation, on same object, with different read flags.

### DIFF
--- a/src/FileIO/LocationInterpolation.cpp
+++ b/src/FileIO/LocationInterpolation.cpp
@@ -321,14 +321,15 @@ geolocation GeoPointInterpolator::Location(double distance, double &slope)
     double rise = l1.Alt() - l0.Alt();
 
     // Tangent vector's magnitude is velocity in terms of distance spline.
-    // Will be unit vector when distance spline has constant rate.
+    // Will be unit vector when distance spline has constant rate. Can have
+    // zero length when processing points without motion.
     double hyp = tangentVector.magnitude();
 
     // Compute adjacent speed.
-    double run = sqrt(fabs((hyp * hyp) - (rise * rise)));
+    double adj = fabs((hyp * hyp) - (rise * rise));
 
     // Gradient.
-    slope = run ? rise / run : 0;
+    slope = (adj > 0.) ? rise / sqrt(adj) : 0.;
 
     // No matter what we still don't permit slopes above 40%.
     slope = std::min(slope,  .4);
--- a/src/FileIO/LocationInterpolation.h
+++ b/src/FileIO/LocationInterpolation.h
@@ -752,7 +752,7 @@ public:
         // Remove frame speed from tangent vector.
         double frameSpeed = m_DistanceWindow.FrameSpeed(bracketRatio);
         double tangentSpeed = tangentVector.magnitude();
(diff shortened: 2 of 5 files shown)


liversedge:
This is not how ergfile works. It is there to support workout file formats.
The create workout wizard can be used to create a workout from any activity-- feel free to improve it, but this is not the way to proceed. Sorry.

ericchristoffersen:
> This is not how ergfile works. It is there to support workout file formats.
> The create workout wizard can be used to create a workout from any activity-- feel free to improve it, but this is not the way to proceed. Sorry.

Hi Mark. Hopefully I can convince you that it is reasonable to support riding these various formats as workouts. If not then I will resubmit the needed fixes from this change but without the support for additional file type imports.

Understand I am working to improve support for virtual rides. A year or so ago I added ability to load gpx file into an erg file. I did this because the existing CRS format does not contain GPS location or altitude. Loading gpx directly was very convenient since many free public rlv come with GPX. The alternative would have been to invent a new file format, which I think we both agree would be insane. In fact I implemented an extended CRS I called CRS_LOC to support location information but erased it in favor of just using gpx directly.

Today when I create a simulated ride I start with a loaded activity file. I remove data types that aren't needed (like heart rate), run smoothings and sometimes adjust fields directly in the edit window. Finally I export to a format that can be loaded as a workout, which for me has usually been gpx.

Recently I have been playing with virtual rides that contain only altitude but no lat/lon data. 

A recent change allows smoothing of this data, but what I found was that there was no way to actually ride these files since export as gpx doesn't contain distance. I looked at the available formats and believe that FIT is a better option since it can contain distance, gps location or altitude alone. This I think is the better format since it covers all bases.

I realize that in this particular case (with distance, altitude and no location) I can get a reasonable CRS using the create workout wizard, but I think it is much more convenient to ride the file directly and not require people to craft an actual crs. Benefit for using the richer file format is that it always contains all the data and it can be reloaded and edited multiple times. CRS is over-stripped for most of my use cases.

While adding the support for FIT I realized that the current gpx to erg translation will work for everything that can be instantiated as a ridefile, so I added all the types that may contain gps or altitude data. I think it is a wonderful trait of golden cheetah that it has such wide format support and I think it is nifty that people can use whatever interchange file type they prefer. There is a nice symmetry to this approach and is literally is no extra work from our side.

I don't understand why you would wish to restrict people to creating workouts using the workout creator. Is fundamentally painful to be working from a dialog.

If the creator were extended to handle the scenarios:
- it would need to start emitting new file type to capture location data (probably today I'd vote for FIT.)
- creator's interface really isn't rich enough to manage all the work you need to do to create a good simulated ride file, I think it would end up converging to allow access to all the tools we have for activities, or relying on activity editing as I do today.

PR #3380, opened 2020-03-22
Fixes for interpolation and gpx file load

ericchristoffersen:
Resubmitting previous change will all additional file types disallowed. This change is now just a set of fixes to current behavior:
- Interpolation math fixes: Slope and interpolation behave correctly when ride point location doesnt change.
- In gpx read: do not open same file twice with different read flags.

Commented the changes for additional file support in case we decide to enable in the future.

### DIFF
--- a/src/FileIO/LocationInterpolation.cpp
+++ b/src/FileIO/LocationInterpolation.cpp
@@ -321,14 +321,15 @@ geolocation GeoPointInterpolator::Location(double distance, double &slope)
     double rise = l1.Alt() - l0.Alt();
 
     // Tangent vector's magnitude is velocity in terms of distance spline.
-    // Will be unit vector when distance spline has constant rate.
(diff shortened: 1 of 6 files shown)


ericchristoffersen:
Several important changes here. Can someone take a look? Thanks.

ericchristoffersen:
User complained about a problem with one of his tts files not workings. I debugged and found it is the same issue that I fixed in locationinterpolation with this change (zero length tangent vector at start.) Please can you take this change. Or tell me what you want? The other part of this change is using tcx and fit for ride and video sync, which is super convenient and works perfectly. Thanks.

amtriathlon:
Hi Eric, I understand the fixes referred to in the PR title are confined to the first commit, I am inclined to cherry-pick it removing the fit parts to apply the critical fixes.
The second part, to use other formats as video-sync sources, I think it worths a separate pull request.

PR #3401, opened 2020-04-15
Enable BT on Windows

erikboto:
Here's a branch that enables BLE on Windows. It requires that the migration to Qt 5.14.2 is completed before it will work, since Qt 5.9 currently used don't have the backend QtConnectivity support.

Qt states the following on [its info page](https://doc.qt.io/qt-5/qtbluetooth-index.html):

> Qt 5.14 adds a native Win32 port supporting Classic Bluetooth on Windows 7 or newer, and Bluetooth LE on Windows 8 or newer. It must be enabled at build time by configuration option -native-win32-bluetooth. The UWP backend is used by default if this option is not set and the Win32 target platform supports the required UWP APIs (minimal requirement is Windows 10 version 1507, with slightly improved service discovery since Windows 10 version 1607).

I believe the UWP backend is the default, so I suspect that Windows 10 version 1607 or later is recommended.

### DIFF
--- a/src/Train/BT40Device.cpp
+++ b/src/Train/BT40Device.cpp
@@ -39,11 +39,11 @@ QMap<QBluetoothUuid, btle_sensor_type_t> BT40Device::supportedServices = {
 BT40Device::BT40Device(QObject *parent, QBluetoothDeviceInfo devinfo) : parent(parent), m_currentDevice(devinfo)
 {
     m_control = new QLowEnergyController(m_currentDevice, this);
-    connect(m_control, SIGNAL(connected()), this, SLOT(deviceConnected()));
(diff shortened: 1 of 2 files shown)


amtriathlon:
I tested on my local Windows 10 - 64bit local build with Qt5.14.2 and it is working for me using a Rhythm+ BTLE HR sensor. Excellent!

erikboto:
Just pushed a small fix-up. QLowEnergyController::Error wasn't registered as a metatype and couldn't be queued, so I reverted the move to QueuedConnection for the error signal. It just logs the error anyway, so it's not triggering the bug on Windows.

erikboto:
I see no problem with merging it now. I have a hard time seeing how it could break anything for existing users, and if any new issues with BT on Windows are discovered it just make sense to find them as early as possible so they can be sorted out.

PR #1698, opened 2015-11-24
Elevation widget to be used as an overlay during video training

vlcvboyer:
Next set of widgets that cover most of meters that may be expected by user.
Thus video can be scaled to full screen.
See screen capture below:
gc capture
Note: I forced "ready" for this screenshot. In fact "ready" is not shown when home trainer is running.

PR #3405, opened 2020-04-22
Elevation gidget

human705:
This is based on the work done by Vianney Voyer about 5 years ago. The original is posted here: https://github.com/GoldenCheetah/GoldenCheetah/pull/1698.  I've made some modification and removed dependencies on video sync as well as reading the slope from rtd rather than calculating it. I'm attaching a screen shot of the video window.

In my tests seems to work best with the changes posted here: https://github.com/GoldenCheetah/GoldenCheetah/pull/3380. Eric's changes allow for tcx files to be used for video sync.
Elevation giddget


amtriathlon:
Hi, thanks for your collaboration!
I think this needs some cleanup, the Additional sources commit adds sources s.t. Distance which already are incorporated, perhaps it would be better to restrict the PR just to Elevation.
Additionally it includes several non-source files s.t. .git-ignore and several VS json files which should not be included.

human705:
Thanks for the feedback. I will clean up the code and re-submit.

amtriathlon:
Perfect, an additional suggestion would be to add the changes to Resources/xml/video-layout.xml for the new widget to show up by default for new users.

Issue #3408, opened 2020-04-23
Part 1: Ability for user to specify their own trainer power curve


ericchristoffersen:
Many requests from people over the years to add support for their trainer. There are now 72 different hard coded functions implemented in realtime controller. Further those equations are sort of dubious looking.

We should add ability for user to specify their own power curve. Propose that they specify the curve with a set of speed/power pairs, this is fit and new virtual device made available.

Lots of moving parts here. Propose this goes in several steps.

Part 1:

Develop polynomial fit, polynomial form for speed/power curve.
Port all existing functions to new form.

Part 2:

Develop UI to allow creating and using new custom virtual device.

Part 3:

Support calibration or verification of chosen power curve.

PR #3410, opened 2020-04-23
Support Custom Virtual Power Curve

ericchristoffersen:
This is part 1 of issue 3408.

Implements all the math needed for custom power curve feature, then ports all current virtual curves in realtimetrainer to use new polyfit object.

Subsequent checkins will add ui to allow custom functions to be defined and used.

Added test function to pull in and compile the templates (not called.)

### DIFF
--- a/src/Core/BlinnSolver.cpp
+++ b/src/Core/BlinnSolver.cpp
@@ -66,6 +66,10 @@ bool IsZero(double value, Args...args) {
     return RangedZeroTest<s_MantissaBitsNeeded>(value, args...);
 }
 
+bool IsZero2(double value, double arg) {
+    return IsZero(value, arg);
+}
+
 // 0 == A*x + B
 Roots LinearSolver(double A, double B) {
     if (IsZero(A, B)) {
--- a/src/Core/BlinnSolver.h
+++ b/src/Core/BlinnSolver.h
@@ -19,6 +19,11 @@
 #if !defined(_BLINNSOLVER_H)
 #define _BLINNSOLVER_H
 
+// Is value zero wrt scale of arg? Returns true if value is indistinguishable
+// from zero with respect to magnitude precision of arg. Put another way,
+// is arg + value ~= arg?
+bool IsZero2(double value, double arg);
+
 // Holds up to 3 roots.
 template <typename T> struct T_resultstruct
 {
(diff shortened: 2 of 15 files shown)


ericchristoffersen:
Implemented dialog for adding virtual power curves.
Next step is to save and restore custom power fit as part of device.

ericchristoffersen:
Dang. Tool disagreement about qualified typename... working on it.

liversedge:
I get a SEGV

crash

liversedge:
You can't dereference via the stack;
std::vector<int>&v={0,0,0,0}
is illegal.

ericchristoffersen:
Hi Mark. Thanks for testing this. I am at a loss how I might deduce what the issue is.

I don't think the code above is illegal. The {0,0,0,0} is constructed to a std::vector temp, then a reference to that is assigned as a parameter of GetPolyFit. References to expression temps are permitted, but it is illegal to use that reference outside the lifetime of the expression.

As an example of what is being done above consider:

int foo(std::vector<int>&v){return v.size();}

It is certainly valid to call foo with:

foo({0,0,0,0});

The reference to the temp is fine.

I believe something else is happening here involving static init.

Can you tell me, is this a data access that is failing? Or is it failing on an indirect call instruction? It could be that the static function pointer array isn't initialized prior to this construction of the static predefinedPolyFit initialization.

How might I debug or at least repro this short of reproducing a linux build?

liversedge:
It is in the screenshot?
It fails when the method is called (the next frame on the stack).
It is illegal and undefined to initialise references with a stack object, 
try compiling:
```std:vector<int>&v={0,0,0,0};```
Mark

ericchristoffersen:
The screenshot says segfault. I think its because its an indirect call through null because the maker array hasnt been init yet.

I uploaded a new fix and think it is fixed now. There is a static init array of polyfit makers defined in polynomialregression.cpp. Then there is the static init array of predefined virtual trainers defined in realtimecontroller.cpp. They are in different compile units so theres no guarantee of their init order. I put the predefined trainer array into a function so isn't init until first use, it relies on the load time init of the polyfit maker array, which is fine. I also added back the constexpr to the polyfit construction but as a macro so it can be enabled in the future when were all using cxx17 tools. Constexpr will cause the maker array to live in the image with no runtime init.

Can you please give this update a try?

ericchristoffersen:
Also, with your above example use const:

    const std::vector<int>& v = { 0,0,0,0 };

liversedge:
Ah ok. Will do. But late here now but will check it out tomorrow. Cheers.

ericchristoffersen:
https://github.com/GoldenCheetah/GoldenCheetah/issues/3408

PR #3411, opened 2020-04-23
Elevation and grade gidgets

human705:
This is based on the work done by Vianney Voyer about 5 years ago. The original is posted here:  https://github.com/GoldenCheetah/GoldenCheetah/pull/1698 . I've made some modification and removed dependencies on video sync as well as reading the slope from rtd rather than calculating it. I'm attaching a screen shot of the video window. The widget is visible in slope mode.

Elevation giddget

In my tests seems to work best with the changes posted here: #3380 . Eric's changes allow for tcx files to be used for video sync.

### DIFF
--- a/src/Train/MeterWidget.cpp
+++ b/src/Train/MeterWidget.cpp
@@ -43,6 +43,7 @@ MeterWidget::MeterWidget(QString Name, QWidget *parent, QString Source) : QWidge
     m_Angle = 180.0;
     m_SubRange = 10;
     boundingRectVisibility = false;
+    forceSquareRatio = true;
 }
 
 void MeterWidget::SetRelativeSize(float RelativeWidth, float RelativeHeight)
@@ -75,7 +76,13 @@ void MeterWidget::AdjustSizePos()
 
 void MeterWidget::ComputeSize()
 {
-    m_Width = m_Height = (m_container->width() * m_RelativeWidth + m_container->height() * m_RelativeHeight) / 2;
+    if (forceSquareRatio)
+        m_Width = m_Height = (m_container->width() * m_RelativeWidth + m_container->height() * m_RelativeHeight) / 2;
+    else
+    {
+        m_Width = m_container->width() * m_RelativeWidth;
+        m_Height =  m_container->height() * m_RelativeHeight;
+    }
 }
 
 QSize MeterWidget::sizeHint() const
@@ -119,12 +126,7 @@ void MeterWidget::setBoundingRectVisibility(bool show, QColor  boundingRectColor
 
 TextMeterWidget::TextMeterWidget(QString Name, QWidget *parent, QString Source) : MeterWidget(Name, parent, Source)
 {
-}
-
-void TextMeterWidget::ComputeSize()
-{
-    m_Width = m_container->width() * m_RelativeWidth;
-    m_Height =  m_container->height() * m_RelativeHeight;
+    forceSquareRatio = false;
 }
 
 void TextMeterWidget::paintEvent(QPaintEvent* paintevent)
--- a/src/Train/MeterWidget.cpp
+++ b/src/Train/MeterWidget.cpp
@@ -37,7 +37,7 @@ MeterWidget::MeterWidget(QString Name, QWidget *parent, QString Source) : QWidge
     m_OutlineColor = QColor(128,128,128,180);
     m_MainFont = QFont(this->font().family(), 64);
     m_AltFont = QFont(this->font().family(), 48);
-    m_BackgroundColor = QColor(96, 96, 96, 200);
+    m_BackgroundColor = QColor(96, 96, 96, 0);
     m_RangeMin = 0;
     m_RangeMax = 100;
     m_Angle = 180.0;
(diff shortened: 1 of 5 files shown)


src/Train/VideoWindow.cpp
```diff
@@ -316,6 +316,59 @@ void VideoWindow::telemetryUpdate(RealtimeData rtd)
...
+        else if (p_meterWidget->Source() == QString("Gradient"))
+        {
+            // Do not show when not in ERG mode (mode=1)
+            if ( rtd.mode == 1)
+                {
+                    p_meterWidget->setWindowOpacity(0); // Hide the widget
+                }
+            // int curLap;
+            double slope = 0.0;
+            if (context && context->currentErgFile() )
```
ericchristoffersen, line 328: This test for existance of ergfile is not needed. Just trust rtd.

src/Train/VideoWindow.cpp
```diff
@@ -429,8 +482,12 @@ void VideoWindow::telemetryUpdate(RealtimeData rtd)
         if (curPosition > VideoSyncFiledataPoints.count() - 1 || curPosition < 1)
             curPosition = 1; // minimum curPosition is 1 as we will use [curPosition-1]
 
-        double CurrentDistance = qBound(0.0,  rtd.getDistance() + context->currentVideoSyncFile()->manualOffset, context->currentVideoSyncFile()->Distance);
-        context->currentVideoSyncFile()->km = CurrentDistance;
+        double CurrentDistance = 0.0;
+        if (context && context->currentVideoSyncFile())
```
ericchristoffersen, line 486: Why is this check for currentVideSyncFile() needed? Isn't it checked 10 lines above?

amtriathlon:
In current version the "Load" source mimics the Dial dual behavior: it is target power for erg workouts and slope for crs ones, this was add to avoid the need to hide/show the widget dependending on the workout type and it was not present on Vianney's original work, so to add "Gradient" as an additional source would be redundant, IMO.

src/Train/VideoWindow.cpp
human705: I removed this check.

src/Train/VideoWindow.cpp
human705: There was a comment by the Vianney that this section of code was checking for a null pointer. I removed it and tried a few tests and it seems to work fine. I thought it was something i didn't understand. I will remove it.

human705:
I will remove the "Gradient" source but the label "load" is confusing IMO. I don't like adding labels in code but since this source reports 2 different units, a label should be added? Something like "W" vs. "%"? Or even "Watts" vs. "% Grade"?

src/Train/VideoWindow.cpp
ericchristoffersen: Vinny was getting slope from the ergfile which was apparently optional, so he had to continuously check for the existance of ergfile. You're getting slope from rtdata which is not optional, there's no need to check about ergfile being present. Please remove all the other unnecessary checks for ergfile. If rtdata doesnt have slope set it will be zero, which is fine.

src/Train/VideoWindow.cpp
```diff
@@ -316,6 +316,59 @@ void VideoWindow::telemetryUpdate(RealtimeData rtd)
...
+            if (!elevationMeterWidget)
+                qDebug() << "Error: Elevation keyword used but widget is not elevation type";
+            else
+            {
+                elevationMeterWidget->setContext(context);
+                // TODO : do not use gradient when not in slope simulation mode
+                //int curLap;
+                double slope = 0.0;
+                if (context && context->currentErgFile())
+                    //slope = context->currentErgFile()->gradientAt((context->currentVideoSyncFile()->km + context->currentVideoSyncFile()->manualOffset) * 1000.0, curLap);
```
ericchristoffersen, line 355: Here.

src/Train/VideoWindow.cpp
```diff
@@ -316,6 +316,59 @@ void VideoWindow::telemetryUpdate(RealtimeData rtd)
...
+            p_meterWidget->AltText = QString(".") + QString::number(abs((int)(p_meterWidget->Value * 10.0) % 10)) + QString("%");
+        }
+        else if (p_meterWidget->Source() == QString("Elevation"))
+        {
+            // Do not show when not in ERG mode (mode=1)
+            if ( rtd.mode == 1)
+                {
+                    p_meterWidget->setWindowOpacity(0); // Hide the widget
+                }
+            if (context && context->currentErgFile())
```
ericchristoffersen, line 342: Here. Etc...

src/Train/VideoWindow.cpp
```diff
@@ -429,8 +446,12 @@ void VideoWindow::telemetryUpdate(RealtimeData rtd)
         if (curPosition > VideoSyncFiledataPoints.count() - 1 || curPosition < 1)
             curPosition = 1; // minimum curPosition is 1 as we will use [curPosition-1]
 
-        double CurrentDistance = qBound(0.0,  rtd.getDistance() + context->currentVideoSyncFile()->manualOffset, context->currentVideoSyncFile()->Distance);
-        context->currentVideoSyncFile()->km = CurrentDistance;
+        double CurrentDistance = 0.0;
+        //if (context && context->currentVideoSyncFile())
```
amtriathlon, line 450: Please, remove the commented out code and reindent the remaining

src/Train/MeterWidget.cpp
```diff
@@ -75,7 +78,13 @@ void MeterWidget::AdjustSizePos()
 
 void MeterWidget::ComputeSize()
 {
-    m_Width = m_Height = (m_container->width() * m_RelativeWidth + m_container->height() * m_RelativeHeight) / 2;
+    if (forceSquareRatio)
```
amtriathlon, line 81: I think it is more clear, and less prone to errors when modifying the code, to use a consistent pattern on if, when the else has {} is better to use them after the condition too.

ericchristoffersen:
Woo!! Congrats Peter!

PR #3415, opened 2020-04-24
Add VO2 and Ventilation graphs to WorkoutWidget and add settings

erikboto:
* When running tests with VO2 measurements, it can be useful to see a
  graph of at least VO2 but also Ventilation to determine when e.g. a
  ramp test results in a plateau and can be considered done.

* Make the color of some VO2 measurements configurable.

* Add chart settings to WorkoutWindow, so that plots can be enabled or
  disabled. Also add the possiblity to do averaging.

* Use the term Ventilation instead of Respiratory Minute Volume

### DIFF
--- a/src/FileIO/RideFile.h
+++ b/src/FileIO/RideFile.h
@@ -212,7 +212,7 @@ class RideFile : public QObject // QObject to emit signals
                           aPower, wprime, aTISS, anTISS, smo2, thb, 
                           rvert, rcad, rcontact, gear, o2hb, hhb,
                           lpco, rpco, lppb, rppb, lppe, rppe, lpppb, rpppb, lpppe, rpppe,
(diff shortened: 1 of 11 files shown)


erikboto:
This is how the settings dialog looks with this PR. 

ww-settings

liversedge:
I'm cool with all the changes except Ridefile.h .. is that necessary?

erikboto:
> I'm cool with all the changes except Ridefile.h .. is that necessary?

No it wasn't really :). I changed to using a custom enum for this in WorkoutWidget instead, which leaves RideFile.h as it was.

I kept it as a separate commit to make it easier for you to see the update, but can squash it into the first one before it's merge-time.

liversedge:
Thanks, that really helps. The Ridefile.h enum is used all over the codebase and likely those changes introduced a regression somewhere.

PR #3419, opened 2020-04-27
AppVeyor - Update to Qt 5.14.2

erikboto:


amtriathlon:
I merged this since the build was passing, but now it is failing with: “The process cannot access the file because it is being used by another process.” at compile time.
My guess is jom parallel building is triggering a bug in VS 2019 since using nmake fixes this, the ugly part is making the build slower and getting close to the 1h AppVeyor limit, but I can’t find alternatives.

erikboto:
I saw this too, and I'm not sure it's a VS2019 issue. I'm fairly certain I've hit this bug with jom and VS2017 too when building here locally. Then I just re-run jom and it's usually fine the next time.

I created a separate PR (#3420) which builds qwt single threaded, which is not ideal but at least allow for parallel build of the rest.

liversedge:
I notice that R 4.0 is being used too.

amtriathlon:
> I notice that R 4.0 is being used too.

Yes, the script if fetching the last R version to use for the build and last week it changed to 4.0, the test I did indicates it works if you have 3.6.3 installed, but we could freeze to a fix version s.t. 3.6.3, to avoid surprises.

liversedge:
Oh if that is the case then lets build with 4.0. So we can then support 3.6 thru 4.0.
It makes life easier for a lot of people

PR #3420, opened 2020-04-27
AppVeyor - Don't use parallel build for QWT

erikboto:
Seems like QWT sometimes fails when using parallel build, so use -j1
when building it to force non-parallel build.


amtriathlon:
I don't even tried this alternative thinking if it failed on QWT, the problem would be worse on the bigger GoldenCheetah source, I'm glad to have been proved wrong!

Issue #3132, opened 2019-07-08
Joule 1.0 'read timeout' 


bowdrillGit:
Sorry to re-open a closed bug (Joule 1.0 'read timeout' #3083), but it was closed with a solution that did not work.  I have the correct udev rule, and the FTDI device is recognized and loaded with correct permissions (I am a member of plugdev).  If I run GC as root, I still cannot download from the device, making me believe this isn't a simple permissions issue.  Please let me know if making a new issue was the wrong move.

When I try to download my ride file from my Joule 1.0 I keep getting a "read timeout" error:
Messages:
Getting list ...
Starting Download ...
Get Unit Software Version...
Download failed

Happens with both
GoldenCheetah DEV-V3.5 1710 (3964)
GoldenCheetah DEV-V3.5 1810 (3968)

lsb_release -a
Distributor ID: Ubuntu
Description: Ubuntu 18.10
Release: 18.10
Codename: cosmic

```ls -la /dev/ttyUSB0 
crw-rw-rw- 1 root plugdev 188, 0 Jul  8 11:04 /dev/ttyUSB0```

```groups
nate adm dialout cdrom sudo dip plugdev input lpadmin sambashare no-wan```

bowdrillGit:
Also happens with 1903 (build ID: 3970)

amtriathlon:
I think you could get more feedback in the forums since no developer seems to have your specific configuration for testing, but if you insist in reporting only here...

Have you seen https://github.com/GoldenCheetah/GoldenCheetah/wiki/FAQ-MOST-COMMON-QUESTIONS-OF-ALL#goldencheetah-doesnt-find-my-powertapjoulesrm-on-ubuntu-linux ?

The code has not changed, v3.4 neither works for you?

If you are able to build from source, enabling debug could help: https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/FileIO/JouleDevice.cpp#L27

bowdrillGit:
Thanks for the suggestion, I will try that

amtriathlon:
Have you done more tests?

PR #3425, opened 2020-04-29
Enable Joule debug traces

amtriathlon:
To help with #3132

**Do not merge!**


amtriathlon:
This was the user's test result:

> download Joule 1.0, GPS or GPS+Get Unit Software Version
> writing '\x10\x02\x00 \x00\x00 ' to device
> writing to device ok
> reading from device
> 
>  and I get a 'read timeout' error like usual

It doesn't seem related to GC code, more likely a drivers or hardware issue.

PR #3430, opened 2020-05-03
BT40: add support for Tacx's ANT over BLE to set gradient

aguinet:
This is based on https://github.com/abellono/tacx-ios-bluetooth-example/

This has only been tested so far with a Tacx Flow Smart T2240. I don't have access to a device with "rolling resistance", so I can't really test it.

### DIFF
--- a/src/Train/BT40Device.cpp
+++ b/src/Train/BT40Device.cpp
@@ -21,6 +21,8 @@
 #include "BT40Controller.h"
 #include "VMProWidget.h"
 
+#include "ANTMessage.h"
+
 #define VO2MASTERPRO_SERVICE_UUID "{00001523-1212-EFDE-1523-785FEABCD123}"
 #define VO2MASTERPRO_VENTILATORY_CHAR_UUID "{00001527-1212-EFDE-1523-785FEABCD123}"
 #define VO2MASTERPRO_GASEXCHANGE_CHAR_UUID "{00001528-1212-EFDE-1523-785FEABCD123}"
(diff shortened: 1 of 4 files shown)


aguinet:
It seems that there is a subset of ANT implemented on the service used, so maybe it would be interesting to use the actual code in `src/ANT` to read/write over this channel to support more features, but I'm just new to these technologies so it might be a bad idea!

amtriathlon:
Hi Adrien, I found this very interesting!

I think it would be nice to test the ergometer mode, to support both modes GC has.

A a matter of form, in the first commit it is very clear which the modifications are, but the second one includes an extensive code reformat which makes harder to isolate the changes, GC guidelines discourage this, see https://github.com/GoldenCheetah/GoldenCheetah/wiki/Guidelines-for-submitting-a-patch, please don’t do that or put it on a separate commit.

Thanks for your collaboration!

aguinet:
Hello! Thanks for the comment.
The second commit looks complicated due to the reindentation, but is quite small if you look at it with "git diff -w". I can remove it thought for this PR!

amtriathlon:
> I can remove it thought for this PR!

Yes, please, remove it or put in a separate, non-functional, commit.

aguinet:
Done!

src/Train/BT40Device.cpp
```diff
@@ -465,3 +475,19 @@ BT40Device::deviceInfo() const
 {
     return m_currentDevice;
 }
+
+void BT40Device::setGradient(double gradient)
+{
+  // Only supported for Tacx devices
+  if (!m_servTacxUART || !m_charTacxUART.isValid()) {
+    return;
+  }
```
michelrdagenais, line 484: I have code to submit shortly to "setGradient" for a Wahoo Kickr Core (and also setWeight since both are closely related). Should we start with a couple of if / else like this or define a class for LoadDevices with methods that are implemented differently for each subclass (Tacx, Kickr...)?

amtriathlon: In the ANT+ case a couple of variants are handled by ifs, it seems to be more variations in BTLE, but since we are just starting with BTLE control we could look at a refactoring in the future. For the moment I would be more interested in new devices handle the 2 modes GC support: slope (setGradient) and ergometer (setLoad).

amtriathlon:
Thanks for your collaboration Adrien!

mjunkmyjunk:
I was basing of the GitHub link in OP to also connect to my trainer which seems not to respond to most apps. I found out it was also using TACX's implementation of FEC Over BTLE.

I can verify that it works on my Magene Gravat 2 Direct Drive Trainer.

Issue #3416, opened 2020-04-24
Remove Distance State from VideoSyncFile


ericchristoffersen:
Reviewing code I came across the 'km' field on the videosyncfile object. Seems strange that an object that represents a resource would need a variable thats bound to current training state. Turns out it is part of a rube-goldburg device thats used to bound distance to the videosync range.

It works thusly: VideoSync is notified with an rtdata object. At that point it grabs the distance + manual offset and assigns that to itself, but not before first bounding the value by the range of the videosync that it represents.

Later trainsidebar grabs distance from rtdata, then... if videosyncfile is present, it grabs distance again from the syncfile.

We should remove the 'km' field from the videosyncfile and implement a method on videosyncfile that provides a range bound Then trainsidebar can simply call the method on videosyncfile to bound distance and the training state will live in a single place.

This is a simple change that will remove unnecessary complexity.

I'll implement this once current dust settles in that area.

ericchristoffersen:
ErgFile has the exact same issue. We've put state onto the object that holds the loaded file data which is confusing and also makes it inefficient to support multiple queries.

PR #3442, opened 2020-05-12
Code Cleanup - Remove Query State from ErgFile

ericchristoffersen:
Move ergfile query state from ergfile to callers.
Factor out the code that establishes the query bracket.
Make ergfile query methods const.

### DIFF
--- a/src/Metrics/WPrime.cpp
+++ b/src/Metrics/WPrime.cpp
@@ -496,11 +496,13 @@ WPrime::setErg(ErgFile *input)
         return; // needs to be a valid erg file...
     }
 
+    ErgFileQueryAdapter ergFileQueryAdapter(input);
+
     minY = maxY = WPRIME;
 
     if (integral) {
 
-        last = input->Duration / 1000; 
+        last = ergFileQueryAdapter.Duration() / 1000; 
         values.resize(last);
         xvalues.resize(last);
 
@@ -514,7 +516,7 @@ WPrime::setErg(ErgFile *input)
 
             // get watts at point in time
             int lap;
-            int value = input->wattsAt(i*1000, lap);
+            int value = ergFileQueryAdapter.wattsAt(i*1000, lap);
 
             powerValues[i] = value > CP ? value-CP : 0;
 
@@ -555,7 +557,7 @@ WPrime::setErg(ErgFile *input)
     } else {
 
         // how many points ?
-        last = input->Duration / 1000; 
+        last = ergFileQueryAdapter.Duration() / 1000; 
         values.resize(last);
         xvalues.resize(last);
 
@@ -563,10 +565,11 @@ WPrime::setErg(ErgFile *input)
         // and will also contain non-zero values
         double W = WPRIME;
         int lap; // passed by reference
+
         for (int i=0; i<last; i++) {
 
             // get watts at point in time
-            int value = input->wattsAt(i*1000, lap);
+            int value = ergFileQueryAdapter.wattsAt(i*1000, lap);
 
             if(value < CP) {
                 W  = W + (CP-value)*(WPRIME-W)/WPRIME;
--- a/src/Train/BicycleSim.cpp
+++ b/src/Train/BicycleSim.cpp
@@ -19,6 +19,7 @@
 #include "BicycleSim.h"
 #include "Settings.h"
 #include "Pages.h"
+#include "ErgFile.h"
 
 BicycleWheel::BicycleWheel(double outerR,         // outer radius in meters (for circumference)
                            double innerR,         // inner rim radius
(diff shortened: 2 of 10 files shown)


liversedge:
simulated rider class is not required for this?
should it really be included in this pull request?

ericchristoffersen:
SimulatedRider not used. A future step is to integrate it into trainsidebar, remove a bunch of cruft. Should really help to clean it up. Another guy is using the simulated rider to build training 'rabbits' and he was rebased off this branch.

ericchristoffersen:
I redid the recent fix for MRC/ERG files. Rather than test current execution state I changed trainsidebar to test the ergfile format. Added hasGradient() and hasWatts() so we dont need to test format itself. Logic is that altitude and gradient do no harm when running in ERGO mode, currently its impossible to run in ERGO mode without an MRC or ERG format file, but in future we might switch to ERGO mode in the midst of a simulated ride (forcing intervals while climbing the stelvio.) The real error in previous code was using the watts data in the mrc file as gradient data (they share same field in point.)

This change also comments out SimulatedRider in bicyclesim.

Main point of this change is still to pull the location state out of the ergfile so that the ergfile becomes data instead of an active participant in real time data state. This is mainly done so multiple locations can be queried without hosing the query state.

ericchristoffersen:
Latest (and probably last) commit:

Removes the 'km' and 'manualOffset' fields from the VideoSync object. Instead routeDistance is tracked using the 'routeDistance' field on the trainsidebar. This is a very nice simplification since now VideoSync object is purely stateless data.

While testing this change I found two related issues:
- route distance tile prints as km int, not floating point in user units
- distance tiles were rounding to nearest, not truncating to meter

This pull request now fixes these issues:

https://github.com/GoldenCheetah/GoldenCheetah/issues/3416
https://github.com/GoldenCheetah/GoldenCheetah/issues/3498
https://github.com/GoldenCheetah/GoldenCheetah/issues/3499

Both VideoSync and ErgFile are now stateless.

ericchristoffersen:
This change seems good and ready.

ericchristoffersen:
Resolved conflicts on rebase.

human705:
Hi Mark,

Is it possible to merge these changes? I have a prototype of a simulated rider pacing the user and attacking at random intervals to enhance the training experience for the user. But it is based on Eric's changes.

Thanks 

Peter

ericchristoffersen:
Ok, this is re-re-re-rebased so works again against latest sources. Can someone please review? Is healthy to get the duplicate state out of the shared ergfile and videosyncfile objects.

ericchristoffersen:
More cleanup and reorg. Iterator now maintained by adapter that implements all the query methods. Kept ErgFile* as const as possible. I think this is a pretty nice separation. I sure would like everyone to stop using global access to get to current ErgFile... While testing these changes I found and fixed #3580 , so that fix is here too.

ericchristoffersen:
When using intensity adjust on slope training file, intensity was allowed to reach zero, which zeros the slope and which is not recoverable. Fix is to prevent slope intensity adjustment from going below 1%. Elevation widget now draws new elevation graph when route elevation is adjusted. Change is detected by comparing saved ymin/ymax against ergfile ymin/ymax. While doing this discovered existing elevation widget code was iterating over entire file to recompute ymin/ymax. Ergfile didn't previously hold or compute ymin so added that field. Now make use of ergfile's ymin/ymax instead of iterating. Also use 0 and Duration instead of iterating to compute xmin and xmax...

ericchristoffersen:
#3582

ericchristoffersen:
Can someone (Mark?) please take a look at this?

ericchristoffersen:
Hi, I now have several trees that are relying on this and the rebasing is getting complicated. Is there a problem with this change or approach it takes? Please let me know what is wrong and I can fix it.

ericchristoffersen:
This also fixes #3589

ericchristoffersen:
@liversedge Hi Mark. Is it possible to get some feedback on this? Thanks!

ericchristoffersen:
Fix more merge conflicts.

ericchristoffersen:
FfwdLap works great, even rapid fire. Also a RewindLap method to ErgFile, ready for its button.

amtriathlon:
I reviewed and tested and the stateless ErgFile looks like a better solution, it is merged by https://github.com/GoldenCheetah/GoldenCheetah/commit/25bb4beaa50b0d2068ed333fd8494c312f3ac711 to include the core part (fixes #3416) and a fixed RewindLap implementation.
All other pieces were separately merged and simulated rider class better be introduced when (and if) it is effectively used.
Thanks for your hard work and let's see how it works in the wild.

PR #3443, opened 2020-05-12
Teach elevation widget to honor skipping

ericchristoffersen:
Elevation widget was showing progress based on distance, which is workout distance and didn't accout for distance change due to skip forward/skip back.

Add routedistance to realtimedata so elevation widget can access it
elevation using routedistance to show progress
Fix uninit iterator in elevation widget paint
Draw route distance in elevation widget

### DIFF
--- a/src/Train/MeterWidget.cpp
+++ b/src/Train/MeterWidget.cpp
@@ -21,6 +21,7 @@
 #include "MeterWidget.h"
 #include "ErgFile.h"
 #include "Context.h"
+#include "Units.h"
 
 MeterWidget::MeterWidget(QString Name, QWidget *parent, QString Source) : QWidget(parent), m_Name(Name), m_container(parent), m_Source(Source)
 {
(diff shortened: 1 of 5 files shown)


amtriathlon:
This is an important fix Eric, thanks!
On the cosmetic side, I think it would be more consistent to display route distance in the Athlete's preferred units instead of meters, like we already do for Distance:
https://github.com/GoldenCheetah/GoldenCheetah/blob/6c318059818eb6a002d84beb118c8c954c478196/src/Train/VideoWindow.cpp#L369
Preferred units are available via context.

amtriathlon:
We use km/mi for ride Distance all over GC, I think it is better to keep units consistent and, specifically, feet seems a weird choice for imperial system users.

ericchristoffersen:
I gotta say i love watching that meters measure. Is great to watch it tick up. Is fun to set goals for various round numbers.  Should i change to km and miles with 3 digits of precision? Id rather see feet than decimal miles, decimal miles make my head hurt. Hey: about yards? :)

amtriathlon:
> I gotta say i love watching that meters measure. Is great to watch it tick up. Is fun to set goals for various round numbers. Should i change to km and miles with 3 digits of precision?

This is what we are already doing with Distance metric:
https://github.com/GoldenCheetah/GoldenCheetah/blob/e8b1f69d633fe7aa2f692744a2fc41b43df3e78b/src/Metrics/BasicRideMetrics.cpp#L456

> Id rather see feet than decimal miles, decimal miles make my head hurt. Hey: about yards? :)

Miles plus the fractional part in yards (or feet) should likely a better solution, but I would prefer to keep it simple.

ericchristoffersen:
No room for a fraction. Ill make it match the other distance displays with km and mi. Goodbye feet! :)

src/Train/MeterWidget.cpp
```diff
@@ -409,14 +409,13 @@ void ElevationMeterWidget::paintEvent(QPaintEvent* paintevent)
...
 
-    QString distanceString;
-    if (context->athlete->useMetricUnits)
-        distanceString = QString::number((int)routeDistanceMeters) + QString("m");
-    else
-        distanceString = QString::number((int)(routeDistanceMeters*FEET_PER_METER)) + QString("ft");
+    routeDistance = ((int)(routeDistance * 1000.)) / 1000.;
+
+    QString distanceString = QString::number(routeDistance, 'f', 3) +
+        ((context->athlete->useMetricUnits) ? QString("km") : QString("mi"));
```
amtriathlon, line 418: We should use tr() instead of QString() to allow those strings be translated

ericchristoffersen:
Can you also take a look at: https://github.com/GoldenCheetah/GoldenCheetah/pull/3380

Its a pretty important fix. Some rlvs just dont work without it.

PR #3448, opened 2020-05-18
Live map

human705:
LiveMap overlay widget for train view.

image


amtriathlon:
Hi Peter, nice work! I have a couple of remarks:
- the conflict seems to be related to the newly merged Eric's fixes to Elevation widget, it should be easy to resolve them since they are unrelated changes.
- GOOGLE_KEY is defined and not use, you likely don't want to publish it either.
- #pragma optimize("", off), unless it is strictly required, it would be better to remove it.
- Map position on the top left corner hides the Trainer Status warning, which shows only when there are problems, status could be moved to top center perhaps.

amtriathlon:
Thanks Peter, there are 2 more changes affecting this:
1) WebKit support was removed in https://github.com/GoldenCheetah/GoldenCheetah/commit/f4fcc93693a3c68b58cb305aaee412fa444499bc, so the NOWEBKIT define is not used any more and WebEngine can be used inconditionally.
2) Code was added to clip the widgets to visible area in https://github.com/GoldenCheetah/GoldenCheetah/commit/0352e19e34c1fb651a4f41be1e123448b68d4e29, so painter.setClipRegion(videoContainerRegion); needs to be added also to this Widget.

And, we really need of #pragma optimize("", off)?

human705:
Hi Ale,
It should be OK now. Let me know if there are any other issues.
Peter

amtriathlon:
Thanks Peter, there are 2 more changes affecting this:

- WebKit support was removed in f4fcc93, so the NOWEBKIT define is not used any more and WebEngine can be used inconditionally.
- Code was added to clip the widgets to visible area in 0352e19, so painter.setClipRegion(videoContainerRegion); needs to be added also to this Widget.

And, we really need of #pragma optimize("", off)?

human705:
Hi Ale,

I removed the NOWEBKIT reference in the header file. I also included all the latest changes in my merge, so the widget clipping is already there. I also removed the "#pragma optimize("", off)" on May 11th as part of the "Added video-layout.xml" https://github.com/GoldenCheetah/GoldenCheetah/pull/3448/commits/1e7508f3837827649787a4f709a163f56ce79260. Do you still see it?
Peter

amtriathlon:
> I also included all the latest changes in my merge, so the widget clipping is already there. 

Merge is not enough, you need to add clipping to your widget like this:
https://github.com/GoldenCheetah/GoldenCheetah/blob/f51a6a6cb186013275bd6071f442343b4b0aa10d/src/Train/MeterWidget.cpp#L353

> I also removed the "#pragma optimize("", off)" on May 11th as part of the "Added video-layout.xml" [1e7508f](https://github.com/GoldenCheetah/GoldenCheetah/commit/1e7508f3837827649787a4f709a163f56ce79260). Do you still see it?

That’s right, but there’s another one in VideoWindow.cpp

human705:
Thanks Ale, i missed that reference. I'll remove it and add the clipping.

human705:
I will re-submit after rebase

PR #3452, opened 2020-05-21
Convert size from pixels to points for use in setPointSize

michelrdagenais:
Currently, size is computed from the QWidget geometry height in pixels and then used to set the font size in points. In many cases the result is not that bad because, depending on the screen size and resolution, the pizel size is not that far from one point. We now convert from pixels to inch (DPI) and then from inches to points (72 points/inch). This solves the problem of the text being clipped sometimes or the margin being too large.

### DIFF
--- a/src/Train/DialWindow.cpp
+++ b/src/Train/DialWindow.cpp
@@ -542,7 +542,7 @@ void DialWindow::resizeEvent(QResizeEvent * )
 
     } else {
         // set point size within reasonable limits for low dpi screens
-        int size = geometry().height()-24;
+        int size = (geometry().height() - 24) * 72 / logicalDpiY();
         if (size <= 0) size = 4;
         if (size >= 64) size = 64;
 


liversedge:
We also have a scaling factor set in main.cpp dpiXFactor and dpiYFactor.
I think the hardcoded limits of 4 and 64 should be scaled in there too?

michelrdagenais:
> We also have a scaling factor set in main.cpp dpiXFactor and dpiYFactor.
> I think the hardcoded limits of 4 and 64 should be scaled in there too?

The  modification proposed fixes the problem of incorrect size calculation that generates text that simply does not fit the window:
Screenshot from 2020-05-27 08-24-34

Otherwise, I am not sure what purpose those limits serve. When the dial window size is larger that 0.88 inches (64 points), the text will remain at that size, even if it only fills a small fraction of the window? On high resolution screens, no limits are set. You change for a slightly higher resolution screen and the behavior changes. However, pixelSizeForFont() tries to find the largest font that fits within the prescribed height, and stops its search at 200 pixels, which then becomes a high limit. I would be fine with removing the limits, replacing the whole if {IfBody} else {ElseBody} by the IfBody. For instance, the TextMeterWidget does not have such limits. I can update the patch accordingly if you want, removing the limits, or I can scale the limits.

Replace:
if (dpiXFactor > 1) {
        font.setPixelSize(pixelSizeForFont(font, geometry().height()-(24*dpiYFactor)));
} else {
...
}
by:
    font.setPixelSize(pixelSizeForFont(font, geometry().height()-(24*dpiYFactor)));

liversedge:
Cool- I think we are on the same page.
To fix this you can scale to fit the available space (taking into account width and height).
The arbitrary limits should be removed- although would be interested to see what happens, since IIRC the 64 limit was added as requesting fonts >64 was ignored (but could be wrong).

michelrdagenais:
The actual width depends on the text displayed. If you know the expected field width, you can account for width at resize time. Otherwise, how can you account for width unless you compute the scaling factor at paint time?
> To fix this you can scale to fit the available space (taking into account width and height).

liversedge:
There are 2 issues:
1) min/max font sizes need to scale to the dpi (via dpi[XY]Factor)
2) scaling text to "fill" the space needs to account for the width and height

The original problem you are trying to fix can be managed by the user by resizing the tile appropriately.

By seeking to adjust font sizes to fill space you introduce a new class of problem... knowing in advance what values will be shown -vs- scaling using a general rule (that may be broken) and allowing for width/height scaling too.

My preference is to behave in a simple manner and allow the user to adjust the size of the tiles....

Sorry :) !

michelrdagenais:
> 1. min/max font sizes need to scale to the dpi (via dpi[XY]Factor)
> 2. scaling text to "fill" the space needs to account for the width and height
> 
> The original problem you are trying to fix can be managed by the user by resizing the tile appropriately.

The current code queries the geometry in pixels, compares this to min/max 4 and 64, and sets the font size in points. This mixes units and is just plain wrong. You compute a font size in pixels to fit the geometry (minus a 12 + 12 pixels border) and set the font using the wrong function (point based). You thus get text that doe not fit. If you enlarge the tile, you will get a larger geometry and still mix units and obtain an incorrect result until you hit the maximum size.

What is the intent, work in points or work in pixels? If we want to work in pixels and the min/max (4 and 64) are in pixels, we would need to change font.setPointSize(size) to font.setPixelSize(size) and indeed it would make sense to scale these min/max with dpiYFactor. (Option A)

Because of the comment "set point size within reasonable limits for low dpi screens", my understanding was that the min/max (4 and 64) are in points (1/72 inch or 0.35mm) and thus my pull request converted the height() to points. In that case, the min/max do not need to be scaled, since they already account for the dpi. (Option B and current pull request)

However, I do not understand in what case one would like to have a large tile with a much smaller text inside (limited by max), or a small tile with clipped text (limited by min). Thus, my suggestion is to remove the test "if (dpiXFactor > 1)" and keep only the strategy currently used for high dpi screens. (Option C)

> My preference is to behave in a simple manner and allow the user to adjust the size of the tiles....

Option C does just that :-). Which option do you prefer?

PR #3453, opened 2020-05-22
Enable Meter Widgets overlaid on Video Window for Linux

michelrdagenais:
Meter Widgets, with a transparent background over the video window, use
little screen estate while providing all the needed information. They
were added for WIN32 first but actually work fine on Linux with minor
flag adjustments.

### DIFF
--- a/src/Train/MeterWidget.cpp
+++ b/src/Train/MeterWidget.cpp
@@ -25,11 +25,16 @@
 MeterWidget::MeterWidget(QString Name, QWidget *parent, QString Source) : QWidget(parent), m_Name(Name), m_container(parent), m_Source(Source)
 {
     setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
-    setWindowFlags(Qt::Window | Qt::FramelessWindowHint);
 
+#ifdef Q_OS_LINUX
+    setWindowFlags(Qt::Window | Qt::FramelessWindowHint | Qt::X11BypassWindowManagerHint);
+    setAttribute(Qt::WA_TranslucentBackground);
+#else
+    setWindowFlags(Qt::Window | Qt::FramelessWindowHint);
     setAttribute(Qt::WA_NoSystemBackground);
     setAttribute(Qt::WA_TranslucentBackground);
     setAttribute(Qt::WA_PaintOnScreen);
+#endif
 
     setAttribute(Qt::WA_TransparentForMouseEvents);
 
--- a/src/Train/VideoWindow.cpp
+++ b/src/Train/VideoWindow.cpp
@@ -87,22 +87,19 @@ VideoWindow::VideoWindow(Context *context)  :
         //vlc_exceptions(&exceptions);
 
 
-    /* This is a non working code that show how to hooks into a window,
-     * if we have a window around */
-#ifdef Q_OS_LINUX
-#if QT_VERSION > 0x50000
-        x11Container = new QWidget(this); //XXX PORT TO 5.1 BROKEN CODE XXX
+#if (defined WIN32) || (defined Q_OS_LINUX)
+#if (defined Q_OS_LINUX) && QT_VERSION <= 0x50000
+        container = new QX11EmbedContainer(this);
 #else
-        x11Container = new QX11EmbedContainer(this);
+        container = new QWidget(this); //XXX PORT TO 5.1 BROKEN CODE XXX
 #endif
-        layout->addWidget(x11Container);
-        libvlc_media_player_set_xwindow (mp, x11Container->winId());
-#endif
-
-#ifdef WIN32
-        container = new QWidget(this);
         layout->addWidget(container);
+
+#if (defined Q_OS_LINUX)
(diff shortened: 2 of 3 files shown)


amtriathlon:
Great! It works in my tests also.
I was trying to enable this on Linux unsuccessfully.
Now we have experimental VLC support on macOS would be interesting to see if we can enable the Overlay Widgets there, the settings for Windows didn't work for macOS either.

michelrdagenais:
I will rework the patch to resolve the conflict later today. A remaining problem is that MeterWidgets do not minimize or move with the main Window, but I have another patch for that using change and move events from the main window to adjust the MeterWidgets.

amtriathlon:
> A remaining problem is that MeterWidgets do not minimize or move with the main Window, but I have another patch for that using change and move events from the main window to adjust the MeterWidgets.

That would be interesting, the current scrolling behavior is annoying

michelrdagenais:
> > A remaining problem is that MeterWidgets do not minimize or move with the main Window, but I have another patch for that using change and move events from the main window to adjust the MeterWidgets.
> 
> That would be interesting, the current scrolling behavior is annoying

Is the same problem present on WIN32 or should I enable the code only on Linux. The idea is that the Main Window emits signals on move and state change (minimize/unminimize) and the VideoWindow subscribes in order to AdjustSizePos or to hide/show the MeterWidgets.

amtriathlon:
Yes, the same problem is present on Windows.
The size and position part, including scrolling, can be fixed just calling AdjustSizePos method in MeterWidget paintEvent before anything else, although I'm not sure this is the best solution, and it doesn't solve he miminize issue.

michelrdagenais:
> The size and position problems, including scrolling, are present on Windows too.
> They can be easily fixed just calling AdjustSizePos method in MeterWidget paintEvent before anything else, although I'm not sure this is the best solution

It is indeed more efficient and elegant to recompute the size and position only on move events from the main window.

> The minimization problem is specific to Linux.

I can enable it only for Linux if needed. MAC_OS may need it too. It should not hurt WIN32 if it is enabled on all platforms. I dont have any windows machines around to test it.

> BTW, removing Qt::X11BypassWindowManagerHint doesn't seem to change anything, is this flag really needed? Code gets simpler without it.

I tried almost all flag combinations. Without it, when you minimize GoldenCheetah, you see 4 or 5 minimized windows, depending on the number of MeterWidgets that you have, it really does not look nice. If there is a way to tell the system that these Meter Widget windows should follow the Main Window, without having to bypass the window manager and program it, that would be great but I have not found it yet. I am new to Qt though...

amtriathlon:
I agree to hide/show the widgets based on Main Window mimization/restore can be a good solution, but I am not sure about the moving and sizing issue because there can be other causes of movement, notably scrolling, and it seems complex and brittle to have to track and bind all of them.

amtriathlon:
> It is indeed more efficient and elegant to recompute the size and position only on move events from the main window.

The problem is there can be several other situations where moving and resizing is required, notably scrolling but may be others, and this solution requires to identify all of them and forward the corresponding signals, it seems complex and brittle to me.

OTOH, for the minimization issue, your proposal seems good for me.

WRT to bypassing the Window Manager, a nasty side effect is the widgets remain on top of whatever other window you place over them.

amtriathlon:
> It is indeed more efficient and elegant to recompute the size and position only on move events from the main window.

I think this doesn't cover all causes by which movement and resizing are required.

OTOH your proposal to solve the minimize/restore issue seems good for me.

A nasty consequence of bypassing the window manager is the widgets remain on top of other programs' windows when switching programs.

michelrdagenais:
> > It is indeed more efficient and elegant to recompute the size and position only on move events from the main window.
> 
> I think this doesn't cover all causes by which movement and resizing are required, being scrolling the Train View the worst offender.

I updated my patch for MeterWidgetFollowMainWindow #3458 to check if the position has changed on Video Window Telemetry updates. It could be an interesting compromise between trying to catch all possible move events, like scrolling, and recomputing the size and position on every update like you proposed. A remaining problem though, with either solution, is that when you scroll the Video Window, part of it may be hidden by the Scroll Area, while the Meter Widgets will remain visible over your desktop. To fix this, we need to know the visible part of the Video Window in the Scroll Area and clip painting of the Meter Widgets accordingly.
 
> OTOH your proposal to solve the minimize/restore issue seems good for me.

Kept, while an alternate method for the move is now proposed in #3458.

> A nasty consequence of bypassing the window manager is the widgets remain on top of other programs' windows when switching programs.

I retried several combinations. With just Qt::Window | Qt::FramelessWindowHint, the Meter Widgets disappear whenever you click on the GC window. You can fix this with Qt::WindowStaysOnTopHint but then you get the same nasty behavior that it stays on top of everything and it also shows one GC instance per Meter Widget on the GNOME panel. Thus Qt::X11BypassWindowManagerHint seems to be the lesser evil.

amtriathlon:
Thanks, I will give it a try to your updated PR.

PR #3454, opened 2020-05-22
Meter widget on linux

michelrdagenais:
Meter Widgets, with a transparent background over the video window, use
little screen estate while providing all the needed information. They
were added for WIN32 first but actually work fine on Linux with minor
flag adjustments.

### DIFF
--- a/src/Train/VideoWindow.cpp
+++ b/src/Train/VideoWindow.cpp
@@ -104,6 +104,7 @@ VideoWindow::VideoWindow(Context *context)  :
 #elif defined(Q_OS_LINUX)
         libvlc_media_player_set_xwindow (mp, container->winId());
 #endif
+#endif
 
 #if defined(WIN32) || defined(Q_OS_LINUX)
         // Video Overlays Initialization: if video config file is not present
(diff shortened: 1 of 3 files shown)


src/Train/VideoWindow.cpp
```diff
@@ -87,20 +87,25 @@ VideoWindow::VideoWindow(Context *context)  :
         //vlc_exceptions(&exceptions);
 
 
-#if (defined WIN32) || (defined Q_OS_LINUX)
+#if defined(WIN32) || defined (Q_OS_MAC) || defined(Q_OS_LINUX)
```
amtriathlon, line 90: The build is failing due to this being unclossed, since these 3 OS are all we support it is redundant and can be removed.

michelrdagenais: I had tested the patch and fixed the #endif but failed to re add the file for commit. It is done now. If the condition (defined(WIN32) || defined (Q_OS_MAC) || defined(Q_OS_LINUX)) is always true when  GC_VIDEO_VLC is defined, then indeed you can simply remove this line. Both ways are good.

amtriathlon: Yes, that condition is always true, no other OS is supported by GC, better to remove it for clarity sake.

amtriathlon:
Thanks!

PR #3456, opened 2020-05-22
Fix incorrect speed (halved) on BT40

michelrdagenais:
Unlike for the cadence value which uses 1/1024 second units, the wheel
revolution value is based on 1/2048 second units [1]. It is easy to
notice the problem when you ride downhill at 25 kph instead of 50kph! In
addition, the speed was initially incorrect because the previous wheel
position value was stale. This would sometimes give the speed of a
rocket for the initial interval and make a jump on the distance of
several km.

[1] Cycling Power Service, Bluetooth Service Specification, Date 2016-05-03, Revision CPS_v1.1, Prepared By Sports and Fitness Working Group, head of page 15.

https://www.bluetooth.org/DocMan/handlers/DownloadDoc.ashx?doc_id=412770

### DIFF
--- a/src/Train/BT40Device.cpp
+++ b/src/Train/BT40Device.cpp
@@ -48,6 +48,7 @@ BT40Device::BT40Device(QObject *parent, QBluetoothDeviceInfo devinfo) : parent(p
     connected = false;
     prevWheelTime = 0;
     prevWheelRevs = 0;
+    prevWheelStaleness = true;
     prevCrankTime = 0;
     prevCrankRevs = 0;
     prevCrankStaleness = -1; // indicates prev crank data values aren't measured values
@@ -451,9 +452,13 @@ BT40Device::getWheelRpm(QDataStream& ds)
     ds >> wheeltime;
 
     double rpm = 0.0;
-    quint16 time = wheeltime - prevWheelTime;
-    quint32 revs = wheelrevs - prevWheelRevs;
-    if (time) rpm = 1024*60*revs / time;
+
+    if(!prevWheelStaleness) {
+        quint16 time = wheeltime - prevWheelTime;
+        quint32 revs = wheelrevs - prevWheelRevs;
+        if (time) rpm = 2048*60*revs / time;
+    }
+    else prevWheelStaleness = false;
 
     prevWheelRevs = wheelrevs;
     prevWheelTime = wheeltime;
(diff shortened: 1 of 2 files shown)


ericchristoffersen:
Wow great! What sort of trainer are you using this on? I'm seeing very few predefined trainers in gc that rely on rpm. Difficult to test...

michelrdagenais:
I am using this on a bluetooth connected Wahoo Kickr Core. I was thrilled to see that it worked out of the box, but quickly became less thrilled when I realized that there was no gradient adjustment. I then figured out that the Power and Speed readings are standard Bluetooth services, well documented, while setting the load on that model is an undocumented service. I thereafter sniffed the bluetooth communication between the Wahoo application on my Android phone and the Kickr Core and could figure out the protocol. I did something similar with my Tacx Fortius many years ago. Once I got the setGradient working, I had a tough time the first time I tried it, climbing at 5 kph and going down at 25 kph, until I checked and realized that there was this small but annoying bug in the wheel RPM computation, halving the reported speed.

ericchristoffersen:
Interesting! I thought the channel from the device gave us speed directly. You mean that gc computes speed from rpm for a kickr core? I had no idea. I thought the core received wheel circumference. Where in gc is the rpm->speed conversion occurring?

michelrdagenais:
The Kickr Core indeed receives the wheel circumference to apply the relevant resistance for the gradient. However, with the Cycling Power service, it explicitly assumes that the speed and distance will be derived from the RPM. The computation takes place in BT40 Controller.cpp:
 
void
BT40Controller::setWheelRpm(double wrpm) {
    telemetry.setWheelRpm(wrpm);
    int wheel;
    if (localDc) wheel = localDc->wheelSize;
    else wheel = 2100;
    telemetry.setSpeed(wrpm * wheel / 1000 * 60 / 1000);
}

michelrdagenais:
Good point, the BT40Device::updateValue method is called for both cases ("Cycling Power" and "Cycling Speed Cadence") while apparently the two specifications differ slightly. I had not looked at the "Cycling Speed and Cadence" specification, since it is not used for the Wahoo. This means that a simple test on has_power should be sufficient to select between 1024 and 2048.

michelrdagenais:
I just submitted a pull request to fix it and then realised that in between my fetch and my pull request a similar fix was merged. I closed my pull request and will test your fix tomorrow but it looks good, except for a typo in a comment, 1048 should be 1024. 

// Power sensor uses 2048 time base and CSC sensor 1048

amtriathlon:
Thanks, I fixed the comment.

michelrdagenais:
> So speed needs to be handled differently according to the source: CSC or Power service

I tested your change, and the reading on the Wahoo remains correct.

Issue #3457, opened 2020-05-23
Lap 2 and subsequent laps start a minute early


ntadmin:
I recently returned to using golden cheetah to look at some data, which is imported from Strava. The data I was looking at had four laps "laps", which are correctly described in the summary on Golden Cheetah:

Lap 1 duration 22:52
Lap 2 duration 10:06 (by inference starts 22:52)
Lap 3 duration 20:01 (by inference starts 32:58)
Lap 4 duration 14:04 (by inference starts 42:59)

This matches the data on Strava. All the imported data (power, HR, etc ) seems correct. For this piece, the start of lap 2 and 4 are especially evident  from the power data.

However, when the data is viewed in, say "Ride" (or any there view). Lap 2 starts at 21:51, lap 3 at 31:58, lap 4 at 41:59, ie they all (apart from lap 1) start a minute early. This shows it pretty clearly:
![Screenshot 2020-05-23 at 12 16 40](https://user-images.githubusercontent.com/8012540/82729409-4495b480-9cef-11ea-9198-909d31019a52.png)

And the original data on strava, not super clear but the start of lap 2 and end of lap 4 bars at the top can be seen to correctly align with the start and end of the "higher power" section:
![Screenshot 2020-05-23 at 12 20 59](https://user-images.githubusercontent.com/8012540/82729486-def5f800-9cef-11ea-8fe8-df0e5ebee7e0.png)

I see the same effect on other rides.

Apologies if this is known or listed elsewhere. I had a look here and on the forum and couldn't find an issue or discussion with "lap" in the title.

ntadmin:
Sorry, should have put version information above, here it is:
![Screenshot 2020-05-23 at 12 23 49](https://user-images.githubusercontent.com/8012540/82729541-44e27f80-9cf0-11ea-90dc-0a8c9d6fe37b.png)

amtriathlon:
There was a previous report similar to this: https://github.com/GoldenCheetah/GoldenCheetah/issues/3254
If you attach the original (FIT) file we could upload it to Strava and sync to try to recreate the problem.

ntadmin:
I think that this is the file, taken directly from the Garmin - it has the right date in the title! The processing route was standard as I understand it: Device (Garmin Edge 510) -> phone (iPhone old SE)-> Garmin Connect app -> onwards.

[...]

PR #3461, opened 2020-05-26
Enable Strava Debug

amtriathlon:
To diagnose #3457
Please do not merge!


amtriathlon:
This was a failed attempt, I forgot secrets are not include when building PR for security reasons.

PR #3464, opened 2020-05-27
Remove Conditional Compilation for Qt < 5.9

amtriathlon:
This is a clean up to remove conditional compilation for all Qt versions
older than the last known to work: Qt 5.9 with Qt WebEngine and Qt Charts.
Includes an update note to INSTALL documents.


liversedge:
Really good idea and will simplify for new devs too.
Scanned through it and it looked correct. Feel free to merge when you are happy !

Issue #3466, opened 2020-05-30
Stored slope is overwritten without user knowledge


amtriathlon:
This behavior was reported at the forum and it is a regression introduced by: https://github.com/GoldenCheetah/GoldenCheetah/commit/6e5fb4217194d4b896e7db680cf8e40f14d11c1c#diff-23f3a036915f75349203d34b3789ad0c
Likely the idea was slope be recomputed after altitude smoothing was added in Fix GPS, but I think this is a mistake since it makes the imported or modified by the user slope data unavailable.
The right solution would be to invalidate slope data, if present, as Fix Elevation does:
https://github.com/GoldenCheetah/GoldenCheetah/blob/2d3eed0a98513f4f0e88d05dc73a64ec59e951a5/src/FileIO/FixElevation.cpp#L247

PR #3468, opened 2020-05-31
Issue: 3466. FixGPS Fixes, Dont Stomp Slope.

ericchristoffersen:
Fix issue where gc would unilateraly recompute slope from altitude, described on gc user group:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/golden-cheetah-users/6NIf5bCv5AY/cFGm1J9MCAAJ

Another issue in same part of ride file where slope was set to 0 if route distance didnt change, now we reuse previous slope.

Fixes in FixGPS:

- If altitude is modified the FixGPS now declares slope invalid so it will be recomputed (previously not needed because slope was always recomputed.)

- Change 'outlier constraint' from being percent of data/spline variance to percent of data/spline standard deviation. This is more intuitive since standard deviation has units in meters, and so percent scales linearly.

- Fix bugs in test stat calculations.

- Change all number printing to use QString::number with specified precision instead of .append of .arg of %1... Nobody wants to see scientific notation for a slope.

- Redo FixGPS ui labels to fit in available space and add more tooltips.

I played with alternative controls based on meters to make this smoothing more intuitive but still I needed to iterate with the map. I think ui needs to be extended to show route on a map tile so users can see the results when they click test. Today I apply fix, examine, then revert to saved and try again.

### DIFF
--- a/src/FileIO/FixGPS.cpp
+++ b/src/FileIO/FixGPS.cpp
@@ -1326,10 +1326,11 @@ bool FixGPS::postProcess(RideFile *ride, DataProcessorConfig *config, QString op
 
     if (smoothingSuccess) {
         // Invalidate slope data to be recomputed based on new altitude data
-        if (fHasSlope && fInvalidateSlope)
+        if (fHasSlope && fInvalidateSlope) {
+            ride->command->startLUW("Invalidate Slope");
             ride->command->setDataPresent(RideFile::slope, false);
-
-        ride->recalculateDerivedSeries(true);
+            ride->command->endLUW();
+        }
     }
 
     if (errors) {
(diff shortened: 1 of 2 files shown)


ericchristoffersen:
https://github.com/GoldenCheetah/GoldenCheetah/issues/3466

amtriathlon:
I haven't looked at full details so just 2 general observations:

1. for the user to be able to undo/redo changes the invalidation should be part of a LUW
2. no need to call recalculate derived series inside the data processor

ericchristoffersen:
Invalidate: maybe things have changed but that wasn't true when I coded it. Maybe its related to LUW? I'll try and see though. Agree its gross to call it.

Whats an LUW?

ericchristoffersen:
LUW: I see! Ok. Trying it now.

Issue #3376, opened 2020-03-20
Allow FIT files to be used for workouts


ericchristoffersen:
Small code change will allow fit files to be loaded as ergfile. Recent case on user group someone had a fit file with no lat/lon data. They can't export as gpx because gpx has no distance, relies on gps data to derive distance. By allowing fit file to be used as ride file we can support ride of any slope/altitude based workout. More flexibile than gpx.

liversedge:
this can be done via the create workout wizard, it will use the currently selected activity.

ericchristoffersen:
WRT workout wizard, issue is not what workout wizard can import from, but what workout wizard can export to. Its little tiny brain only thinks in terms of crs. I only saw it being able to export a crs which is a very limited slope only file. What we want to be able to do is load from anything reasonable into the in-memory ErgFile format. I'd rather keep my workouts as a single file type. Today I prefer tts when possible, then gpx, then crs. We can't generate tts, we can emit gpx. Even better we can emit tcx which holds all kinds of fields and is viewable and editable with a text editor.

I dont understand why we want to tell anyone what is the workout interchange format that they should be using. Its all data stored in different standard conventions, let the user pick what they want for their own reasons. If we force them to go through the workout wizard we'll have exactly the same problem having to second guess what they consider to be the orthodox file type.

liversedge:
My apologies, I didn't realise the workout wizard was so useless!
It needs a brain transplant...

liversedge:
There *is* a fit file protocol specifically for workouts (not recorded activities).

ericchristoffersen:
What we're trying to do here is use a sharable open format that can hold the data needed for a virtual ride. Sort of like tacx' tts but open and easily generated. I don't think its our business to dictate the file interchange format. TCX for example is also really nice to use. RLV really sucks because it stores distance as a set of rates. Gpx sucks because it has no explicit distance, distance is derived by the reader.

The benefit of a generic file like fit and tcx is that they can hold all the data needed for a virtual ride. The same file can be used for both ergfile and for video sync. Is lovely that there is no longer any possibility of asynchrony. Also makes file management much easier. I am now able to have a single file in each of my 'rlv' ride directories. For my tacx rides I still use the tts because they work well. They have the same concept of a single file that acts as ride and video sync.

Things we want:
- a clear unambiguous distance (unlike gpx where distance is derived.)
- gps location (or just altitude if thats all there is, (or no location if there isn't any.))
[...]

PR #3469, opened 2020-05-31
Additional file types for RideFile and VideoSyncFile

ericchristoffersen:
NOTE: Initial commit only adds code to support new file types but does not enable anything new.

Story:

There are no tools available to create tts or rlv files, and those encodings are complex and painful to read, let alone edit or create from scratch. Also it is unnecessarily pain to have a videosync file that is separate from a gpx.

This change allows the use of generic file types to act as both ErgFile and VideoSyncFile, so long as the particular instances have the data streams needed to do the job (for example distance and altitude to be an ErgFile, distance and time to act as a videosync.)

This is an extremely convenient way to operate since it means a single open and editable file with distance, location and time can act as an ergfile and a videosync file. In practice this means you can very easily create a videosynced virtual ride:
1) start your camera
2) start recording your training session
3) and go for a ride
4) stop recording training
5) stop recording video
6) crop the start and end of the video file so time matches start and end of training file
7) run route smoothing on the training file (at least to smooth altitude)
8) export activity to virtual ride folder alongside your video file
9) scan for workouts in gc and your new video will show along with your activity in both ergfile and videosyncfile sections
10) select and ride

A few people have tried this and say it works well.

Both the videosync and ridefile are loaded via the generic ridefilefactory.

### DIFF
--- a/src/Train/ErgFile.cpp
+++ b/src/Train/ErgFile.cpp
@@ -42,6 +42,19 @@ static bool setSupported()
     ::supported << ".zwo";
     ::supported << ".gpx";
     ::supported << ".tts";
+
+    // Additional supportable ridefile types. Uncomment to enable.
+    //::supported << ".bin";
+    //::supported << ".bin2";
+    //::supported << ".fit";
+    //::supported << ".fitlog";
+    //::supported << ".hrm";
+    //::supported << ".pwx";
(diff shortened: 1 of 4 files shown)


ericchristoffersen:
https://github.com/GoldenCheetah/GoldenCheetah/issues/3376

PR #3472, opened 2020-06-04
Add support for the Wahoo Kickr (Core) through BlueTooth LE

michelrdagenais:
All the commands to set the different parameters (weight, wheel size, gradient or load, wind speed...) are added to BT40Device and BT40Controller. This way, the load / gradient are set on the Wahoo Kickr as you ride. A separate pull request will add support to actually take advantage of the parameters beyond load and gradient, like the cyclist weight, now available in this and other devices like the Fortius.

### DIFF
--- a/src/Train/BT40Controller.cpp
+++ b/src/Train/BT40Controller.cpp
@@ -27,7 +27,7 @@ BT40Controller::BT40Controller(TrainSidebar *parent, DeviceConfiguration *dc) : 
     discoveryAgent = new QBluetoothDeviceDiscoveryAgent();
     localDc = dc;
     if (localDc) wheelSize = localDc->wheelSize;
-    else wheelSize = 2100;
+    else wheelSize = 2100.0;
 
     connect(discoveryAgent, SIGNAL(deviceDiscovered(const QBluetoothDeviceInfo&)),
 	    this, SLOT(addDevice(const QBluetoothDeviceInfo&)));
@@ -163,7 +163,7 @@ BT40Controller::deviceScanError(QBluetoothDeviceDiscoveryAgent::Error error)
 void
 BT40Controller::setWheelRpm(double wrpm) {
     telemetry.setWheelRpm(wrpm);
-    telemetry.setSpeed(wrpm * wheelSize / 1000 * 60 / 1000);
+    telemetry.setSpeed(wrpm * wheelSize / 1000.0 * 60.0 / 1000.0);
 }
 
 void BT40Controller::setLoad(double l)
--- a/src/Train/BT40Controller.cpp
+++ b/src/Train/BT40Controller.cpp
@@ -149,6 +149,7 @@ BT40Controller::addDevice(const QBluetoothDeviceInfo &info)
 void
 BT40Controller::scanFinished()
 {
+    emit setNotification(tr("Bluetooth scan finished"), 2);
     qDebug() << "BT scan finished";
 }
 
--- a/src/Train/BT40Controller.cpp
+++ b/src/Train/BT40Controller.cpp
@@ -26,6 +26,9 @@ BT40Controller::BT40Controller(TrainSidebar *parent, DeviceConfiguration *dc) : 
     localDevice = new QBluetoothLocalDevice(this);
     discoveryAgent = new QBluetoothDeviceDiscoveryAgent();
     localDc = dc;
+    if (localDc) wheelSize = localDc->wheelSize;
+    else wheelSize = 2100;
+
     connect(discoveryAgent, SIGNAL(deviceDiscovered(const QBluetoothDeviceInfo&)),
 	    this, SLOT(addDevice(const QBluetoothDeviceInfo&)));
     connect(discoveryAgent, SIGNAL(error(QBluetoothDeviceDiscoveryAgent::Error)),
@@ -138,6 +141,7 @@ BT40Controller::addDevice(const QBluetoothDeviceInfo &info)
         devices.append(dev);
         dev->connectDevice();
         connect(dev, &BT40Device::setNotification, this, &BT40Controller::setNotification);
+        dev->setWheelCircumference(wheelSize);
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/Train/BT40Controller.cpp
```diff
@@ -159,14 +163,62 @@ BT40Controller::deviceScanError(QBluetoothDeviceDiscoveryAgent::Error error)
 void
 BT40Controller::setWheelRpm(double wrpm) {
     telemetry.setWheelRpm(wrpm);
-    int wheel;
-    if (localDc) wheel = localDc->wheelSize;
-    else wheel = 2100;
-    telemetry.setSpeed(wrpm * wheel / 1000 * 60 / 1000);
+    telemetry.setSpeed(wrpm * wheelSize / 1000 * 60 / 1000);
```
ericchristoffersen, line 166: You should suffix the immediates with '.' to ensure the division is evaluated as double.

michelrdagenais: This is existing code, and is correct since wrpm is double, the operations are of the same precedence (* and /) and have left to right associativity ([C++ reference on operators precedence](https://en.cppreference.com/w/cpp/language/operator_precedence)). However, I agree that adding "." would prevent any questioning, and double checking like I did, about that. I can modify the original code as such if no one objects.

michelrdagenais: I made the conversion obvious and I checked the BT40Device and BT40Controller files for similar instances. There was in fact one integer division for the speed and the cadence which could loose a bit of accuracy. In addition, the speed was not reset upon disconnection.

amtriathlon:
Hi Michel, the code looks good to me and I don't think it introduces any conflict, but I have no chance to test it, so let me know when you think it is ready to merget.

As a side note, BT40Device and BTController can use `emit setNotification(tr("message to be displayed at Train Bottom"), seconds)` if you want the user be aware of the new device being discovered, or changed their state, this is not widely used yet likely because that facility was introduced later and by other developer, but I think it is something to consider, specially for BT sensors which are dynamically discovered at device start without a pairing step.

michelrdagenais:
> Hi Michel, the code looks good to me and I don't think it introduces any conflict, but I have no chance to test it, so let me know when you think it is ready to merget.
I believe that it is ready to merge now. I have been using a previous version of this patch to train for a few times in the past weeks. As I wrote, a followup to this patch will be to exploit the setWeight capability which should also be used for other devices like the Fortius. However, I wanted to get the current pull request to be merged first, to see if modifications were required, and build the followup patch on it accordingly. I was also waiting for feedback on my post about "Weight is not taken into account by training devices".
> 
> As a side note, BT40Device and BTController can use `emit setNotification(tr("message to be displayed at Train Bottom"), seconds)` if you want the user be aware of the new device being discovered, or changed their state, etc. 
Good suggestion, I will make a separate patch for this and check if the Fortius can also benefit from it.

mjunkmyjunk:
@michelrdagenais for the KICKR Core w/ the WAHOO_BRAKE_UUID, is there any steps needed to "unlock" the trainer? The Cycleops Hammer H1 also uses this UUID and sending a Unlock command (0x20, 0xee,  0xfc) [https://github.com/codeinversion/sensors-swift-trainers/blob/93db8dd000e47233435a41ba061c23541ee18bb4/Sources/SwiftySensorsTrainers/WahooTrainerSerializer.swift#L37] didn't work.

I'm not sure where the 0xee and 0xfc came about but I noticed that for GC, you've used 

https://github.com/michelrdagenais/GoldenCheetah/blob/5bc9d91981ed4c240d588b61db1e10f523eb1010/src/Train/BT40Device.cpp#L279

which is sending just 0x20 If not mistaken.

After which it can be controlled and target power can be sent?

michelrdagenais:
The write is to the notification descriptor, to request updates from the Wahoo of the "indicator" type. This is what the Wahoo application did when I looked at the packets exchanged between my phone and the trainer through bluetooth. I did not see any "unlock" sequence sent by the Wahoo application.

> @michelrdagenais for the KICKR Core w/ the WAHOO_BRAKE_UUID, is there any steps needed to "unlock" the trainer? The Cycleops Hammer H1 also uses this UUID and sending a Unlock command (0x20, 0xee, 0xfc) [ https://github.com/codeinversion/sensors-swift-trainers/blob/93db8dd000e47233435a41ba061c23541ee18bb4/Sources/SwiftySensorsTrainers/WahooTrainerSerializer.swift#L37 ] didn't work.
> 
> I'm not sure where the 0xee and 0xfc came about but I noticed that for GC, you've used
> 
> https://github.com/michelrdagenais/GoldenCheetah/blob/5bc9d91981ed4c240d588b61db1e10f523eb1010/src/Train/BT40Device.cpp#L279
> 
> `service->writeDescriptor(notificationDesc, QByteArray::fromHex("0200")); `
> which is sending just 0x20 If not mistaken.
> 
> After which it can be controlled and target power can be sent?
> 
> Thanks

michelrdagenais:
Yes the target power or the slope, and you also send the rider characteristics. I do not have the background about what you are trying to achieve. A good trick is to look at an existing application on your phone that commands the trainer. You change one parameter at a time (vary the slope, the rider weight, change the speed at which you are pedaling...). On Android, there is a debug mode to dump the bluetooth exchanges. You then analyze the packets exchanged and replicate that.

> So basically you're subscribing to the Notify Characteristic to get updates?
> 
> After which you start sending it Target Power Wattage to the Wahoo_BRAKE_UUID ? eg:
> HEX 0x426400 - 100w

mjunkmyjunk:
Thanks for the insight. Truly appreciate it. As for background, I'm trying to see how I can get a CycleOps Hammer H1 to also be controllable. I found that it also uses the WAHOO_BRAKE_UUID and was looking at the (your) code for the sequence. I have a packet logger and also some tools from Apple that is supposed to help, but since I don't have that trainer physically, (working with a tester across the world) its not easy going to say the least.

PS: I have a trainer that does the ANT+FEC Over BLE (the Tacx version) and managed to get it squared away via BTLE on iOS as well as using the fantastic web-ble Web-App Flux (https://github.com/dvmarinoff/Flux) which incidentally also doesn't work for Wahoo_Brake_UUID. also GoldenCheetah unable to control my trainer as well. (https://groups.google.com/g/golden-cheetah-users/c/A6yzKylcD2E/m/3Qtiv0B0CQAJ) 

> Yes the target power or the slope, and you also send the rider characteristics. I do not have the background about what you are trying to achieve. A good trick is to look at an existing application on your phone that commands the trainer. You change one parameter at a time (vary the slope, the rider weight, change the speed at which you are pedaling...). On Android, there is a debug mode to dump the bluetooth exchanges. You then analyze the packets exchanged and replicate that.

michelrdagenais:
It is unfortunate that Wahoo uses their proprietary descriptor to control the brake instead of Bluetooth FTMS. My reference model was the Wahoo application. Perhaps Zwift uses the Wahoo descriptors in a slightly different way. For instance, it may use notification instead of indication (notification with acknowledgement) and it is possible that Wahoo supports both while other models (CycleOps or Thinkrider) may not. You also need to insure that correct values are sent for the other parameters like the weight.

In my BTLE Wahoo code, the only tricky part is that if the send fails, it tries to resend commands. If the computer is a bit far from the trainer, or there are obstacles, the transmission may be pretty unreliable. The other thing is that the Wahoo controller seemed fairly slow, not ready to accept new commands at a high rate. Perhaps other models fare even worse and you have to be careful about the timing (i.e. if you do not wait long enough, the second command is ignored).

> Thanks for the insight. Truly appreciate it. As for background, I'm trying to see how I can get a CycleOps Hammer H1 to also be controllable. I found that it also uses the WAHOO_BRAKE_UUID and was looking at the (your) code for the sequence. I have a packet logger and also some tools from Apple that is supposed to help, but since I don't have that trainer physically, (working with a tester across the world) its not easy going to say the least.
> 
> PS: I have a trainer that does the ANT+FEC Over BLE (the Tacx version) and managed to get it squared away via BTLE on iOS as well as using the fantastic web-ble Web-App Flux (https://github.com/dvmarinoff/Flux) which incidentally also doesn't work for Wahoo_Brake_UUID. also GoldenCheetah unable to control my trainer as well. (https://groups.google.com/g/golden-cheetah-users/c/A6yzKylcD2E/m/3Qtiv0B0CQAJ)

PR #3477, opened 2020-06-05
Elevation Widget: Lazy Init Polygon to Avoid 2x Route Point Iteration

ericchristoffersen:
Debugging the vlc hangs I found significant time being spend in the elevation widget PaintEvent, it is doing two complete passes over the ergfile points in order to compute something that only changes on widget resize. Moved elevation polygon onto the elevation widget and only recompute when needed. This reduces load and eliminates elevation widget flickering on long route files.

Unfortunately this does nothing for the hangs.


ericchristoffersen:
https://github.com/GoldenCheetah/GoldenCheetah/issues/3478

PR #3482, opened 2020-06-06
Live map

human705:
Resubmit #3448 - LiveMap overlay widget for train view.

image

### DIFF
--- a/src/Train/MeterWidget.cpp
+++ b/src/Train/MeterWidget.cpp
@@ -15,7 +15,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc., 51
  * Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
-#pragma optimize("", off)
+
 #include <QtGui>
 #include <QGraphicsPathItem>
 #include "MeterWidget.h"
@@ -480,6 +480,7 @@ LiveMapWidget::LiveMapWidget(QString Name, QWidget *parent, QString Source, Cont
     liveMapView->setPage(webPage);
     liveMapInitialized = false;
     routeInitialized = false;
+    liveMapworkoutname = "";
  }
 
 void LiveMapWidget::paintEvent(QPaintEvent* paintevent)
@@ -516,6 +517,12 @@ void LiveMapWidget::paintEvent(QPaintEvent* paintevent)
 //***************************************************************************************************************
 void LiveMapWidget::initLiveMap ()
 {
+    if (liveMapworkoutname != "" && liveMapworkoutname != context->currentErgFile()->filename) {
+        liveMapInitialized = false;
+        routeInitialized = false;
+    }
+    liveMapworkoutname = context->currentErgFile()->filename;
+
     if ( ! this->liveMapInitialized ) {
 
         int mapZoom = 16;
@@ -523,33 +530,15 @@ void LiveMapWidget::initLiveMap ()
         createHtml(this->curr_lat, this->curr_lon, mapZoom);
         liveMapView->page()->setHtml(currentPage);
         liveMapView->show();
-        ////this->liveMapInitialized = true;
-
-
-        ////// createHTML2 ---
-        //int mapZoom = 16;
-        //createHtml2();
-        //liveMapView->page()->setHtml(currentPage);
-        //liveMapView->show();
-        ////// createHTML2 ---
     }
 }
 
 void LiveMapWidget::plotNewLatLng(double dLat, double dLon)
 {
     QString code ="";
-    QString mycenterMap = "";
-    std::string sMyStr;
-
-    //createHtml2();
-    //liveMapView->page()->setHtml(currentPage);
-    //code = QString("initMap (" + sLat + ", " + sLon + ", " + sMapZoom + ");");
-    //liveMapView->page()->runJavaScript(code);
-
     QString sLat = QVariant(dLat).toString();
     QString sLon = QVariant(dLon).toString();
 
-    //// createHTML2 ---
     if (!this->liveMapInitialized) {
         
         routeLatLngs = "[";
@@ -562,15 +551,7 @@ void LiveMapWidget::plotNewLatLng(double dLat, double dLon)
             routeLatLngs += "]";
         }
         routeLatLngs += "]";
-        //code = QString("centerMap(" + sLat + ", " + sLon + ", 16);");
-        //code += QString("showMyMarker(" + sLat + ", " + sLon + ");");
-        //code += QString("moveMarker(" + sLat + " , " + sLon + ");");
-        code = QString("showRoute(" + routeLatLngs + ");");
-        //liveMapView->page()->runJavaScript(code);
         this->liveMapInitialized = true;
-        //liveMapView->show();
-        //sMyStr = code.toStdString();
-        //qDebug("==> code @ 1 = %s", sMyStr);
     }
     else
     {
@@ -580,17 +561,8 @@ void LiveMapWidget::plotNewLatLng(double dLat, double dLon)
             this->routeInitialized = true;
         }
         code += QString("moveMarker(" + sLat + " , " + sLon + ");");
-        
-        //sMyStr = code.toStdString();
-        //qDebug("==> Code @ 2 = %s", sMyStr);
     }
-    
-    //view->page()->runJavaScript("function getelement(){return $('#elementid').val();} getelement();", [this](const QVariant& v) { qDebug() << v.toString(); });
-    //    view->page()->runJavaScript("jsfun();",[this](const QVariant &v) { qDebug()<<v.toString();});
-    //liveMapView->page()->runJavaScript(code, [this](const QVariant& v) { qDebug()<< v.toString(); });
     liveMapView->page()->runJavaScript(code);
-
-    //// createHTML2 ---
 }
 
 void LiveMapWidget::createHtml(double dLat, double dLon, int iMapZoom)
@@ -640,7 +612,7 @@ void LiveMapWidget::createHtml(double dLat, double dLon, int iMapZoom)
     "}).addTo(mymap);\n"
     "function showRoute(myRouteLatlngs) {\n"
         "    routepolyline = L.polyline(myRouteLatlngs, { color: 'red' }).addTo(mymap);\n"
-        "    mymap.fitBounds(routepolyline.getBounds());\n"
+ //       "    mymap.fitBounds(routepolyline.getBounds());\n"
         "};\n");
     // Move marker function
     currentPage += QString(    "function moveMarker(myLat, myLon) { \n"
--- a/src/Train/MeterWidget.cpp
+++ b/src/Train/MeterWidget.cpp
@@ -515,28 +515,41 @@ void LiveMapWidget::paintEvent(QPaintEvent* paintevent)
 
 //***************************************************************************************************************
 
-void LiveMapWidget::initLiveMap ()
+// Show Live map zoomed in at the current location
+void LiveMapWidget::lazyInitLiveMap (double dLat1, double dLon1)
 {
-    //Check if workout has changed and reset the route
+    int mapZoom = 16;
+    liveMapView->resize(m_Width, m_Height);
+    createHtml(dLat1, dLon1, mapZoom);
+    liveMapView->page()->setHtml(currentPage);
+    liveMapView->show();
+}
+
+// Build LatLon array for selected workout
+void LiveMapWidget::buildRouteArrayLatLngs() {
+    routeLatLngs = "[";
+    for (int pt = 0; pt < context->currentErgFile()->Points.size() - 1; pt++) {
+        if (pt == 0) { routeLatLngs += "["; }
+        else { routeLatLngs += ",["; }
+        routeLatLngs += QVariant(context->currentErgFile()->Points[pt].lat).toString();
+        routeLatLngs += ",";
+        routeLatLngs += QVariant(context->currentErgFile()->Points[pt].lon).toString();
+        routeLatLngs += "]";
+    }
+    routeLatLngs += "]";
+}
+
+//
+// Plot route and move marker if geolocation is reasonable
+void LiveMapWidget::plotNewLatLng(double dLat, double dLon, double dAlt)
+{
+    //Check if workout has changed and reset flags.
+    //This will rebuild the route and center the map at the new location
     if (liveMapworkoutname != "" && liveMapworkoutname != context->currentErgFile()->filename) {
         liveMapInitialized = false;
         routeInitialized = false;
     }
     liveMapworkoutname = context->currentErgFile()->filename;
-
-    if ( ! this->liveMapInitialized ) {
-
-        int mapZoom = 16;
-        liveMapView->resize(m_Width, m_Height);
-        createHtml(this->curr_lat, this->curr_lon, mapZoom);
-        liveMapView->page()->setHtml(currentPage);
-        liveMapView->show();
-    }
-}
-//
-// Plot route and move marker if geolocation is reasonable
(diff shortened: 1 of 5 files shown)


ericchristoffersen:
Just tried it out. Rad. Love the zoom buttons.

amtriathlon:
Hi Peter, I used this on both Windows and Linux and it seems to working very well and it is an interesting feature to add.

OTOH, I think the Zoom parameter should be configurable like size and position, there are methods and member variables which are not performing any useful function and they should be removed, the details about initialization are better handled in the widget itself, and resize can be handled in resizeEvent instead on each update.

Since these are a lot of changes, instead of commenting your code, I will push a new commit for evaluation and testing.

Context may be re-added at the future if required, for example to add the route, but I think it should be set once for all MeterWidget at creation time instead of at each update, or we could add an explicit initialization method to be called at workout start.

human705:
Thanks Ale. As you may have seen I've created a chart with the same functionality for users without video files. I've been looking for a suitable location to allow each user to personalize the map. In addition to the zoom factor, the map can allow dragging and double click zoom. Eric also suggested that the map starting point should also be configurable. I'm looking at the code in RideMapWindow.cpp where osmTSURL is set but i'm not sure that would be the appropriate location. Maybe a new tab in Options -> athlete?

Im also seen that my video is not synchronized with the map because i can't create an RLV file. Maybe we can all the TCX and/or FIT functionality created by Eric to make the experience even better.

I'll wait for your commit and see what propose. 

Peter

amtriathlon:
> Thanks Ale. As you may have seen I've created a chart with the same functionality for users without video files. I've been looking for a suitable location to allow each user to personalize the map. In addition to the zoom factor, the map can allow dragging and double click zoom. Eric also suggested that the map starting point should also be configurable. I'm looking at the code in RideMapWindow.cpp where osmTSURL is set but i'm not sure that would be the appropriate location. Maybe a new tab in Options -> athlete?

No, it should be a property of the chart, see how this is handled for RideMapWindow and the LiveMap chart can do something similar.
For this widget, if you want the Tile Server be configurable, it could be added to video-layout.xml, but I think we could start with the default you set up.

amtriathlon:
These are the changes I made: https://github.com/amtriathlon/GoldenCheetah/commit/bb64268f6590f6502df45158a6f528ebfd29ed24,  but I cannot push them because you did a forced push on this branch.

src/Train/MeterWidget.cpp
```diff
@@ -462,4 +466,225 @@ void ElevationMeterWidget::paintEvent(QPaintEvent* paintevent)
...
+    routeLatLngs += "]";
+}
+
+//
+// Plot route and move marker if geolocation is reasonable
+void LiveMapWidget::plotNewLatLng(double dLat, double dLon, double dAlt)
+{
+    //Check if workout has changed and reset flags.
+    //This will rebuild the route and center the map at the new location
+    if (liveMapworkoutname != "" && liveMapworkoutname != context->currentErgFile()->filename) {
```
ericchristoffersen, line 548: How to correctly determine that ergfile has changed? I believe here is a pointer compare, but even string.equal() will succeed when it should not if two workouts have the same name. Maybe its best to use string.equal and be sure to clear this name string when training mode stops?

src/Train/MeterWidget.cpp
```diff
@@ -462,4 +466,225 @@ void ElevationMeterWidget::paintEvent(QPaintEvent* paintevent)
...
+        liveMapInitialized = false;
+        routeInitialized = false;
+    }
+    liveMapworkoutname = context->currentErgFile()->filename;
+    geolocation geoloc(dLat, dLon, dAlt);
+    if (geoloc.IsReasonableGeoLocation()) {
```
ericchristoffersen, line 554: This checks that current query location is valid, thorough but maybe unnecessary since it came from interpolator. But below is route construction that is obtaining raw locations from route, they should all be tested to ensure they are valid.

src/Train/MeterWidget.cpp
ericchristoffersen: I think maybe easiest if you just handle the transition from no-name to name. When that transition occurs you populate the route. Make sure the name is cleared when the training is stopped. Then when transition occurs to new name you can reset everything.

src/Train/MeterWidget.cpp
```diff
@@ -462,4 +466,225 @@ void ElevationMeterWidget::paintEvent(QPaintEvent* paintevent)
...
+        "};\n");
+    // Move marker function
+    currentPage += QString(    "function moveMarker(myLat, myLon) { \n"
+    "   mymap.panTo(new L.LatLng(myLat, myLon));\n"
+    "    mymarker.setLatLng(new L.latLng(myLat, myLon));}\n"
+    "</script>\n"
+    "</body></html>\n");
+}
+
+void LiveMapWidget::createHtml2()
```
ericchristoffersen, line 636: Why are there two createHtml? At the least it looks like a lot of duplication between the two createHtmls. Consolidate common work into a function?

human705: I created the second one to test javascript functions, it will be removed.

src/Train/MeterWidget.cpp
human705: I agree that stopping the training is the correct place ti reset the name. I can do that in the chart. However, it appears that when training is stopped the video window (videoWindow.cpp code below) simply hides the widgets. I'm not sure how to clear the name from there. 

`void VideoWindow::stopPlayback()
{
    if (context->currentVideoSyncFile())
        context->currentVideoSyncFile()->manualOffset = 0.0;
#ifdef GC_VIDEO_VLC
    if (!m) return; // ignore if no media selected
    // stop playback & wipe player
    libvlc_media_player_stop (mp);
#endif
#ifdef GC_VIDEO_QT5
    mp->stop();
#endif
    foreach(MeterWidget* p_meterWidget , m_metersWidget)
        p_meterWidget->hide();
}`

human705:
Apologies for the forced push, still learning the intricacies of git. I merged your changes and working to add the shoe route functionality again. Should I close this PR and create a new one?

amtriathlon:
No problem Peter, the force push is only a problem when more than one is working on the same branch. I would say to continue with this, we can rebase later.

To add route show you will need Context and I don't like to set context on every update as we did for elevation, besides you may need to know when a workout starts/ends and it makes no sense to try to guest that situation on every update when we could add start and stop events to MeterWidget to be called at:

Workout starts: https://github.com/GoldenCheetah/GoldenCheetah/blob/04b4d945fa358467a7772669506964cdc0181779/src/Train/VideoWindow.cpp#L213

Workout ends: https://github.com/GoldenCheetah/GoldenCheetah/blob/04b4d945fa358467a7772669506964cdc0181779/src/Train/VideoWindow.cpp#L240

For the original widgets this methods don't need to do anything, but for LiveMap and Elevation, they can make the logic simpler and more clear, Context can be passed there to avoid going through the parser to set it at creation time.

human705:
I moved the map initialization to the workout start and end. I added Context back but only using it for start and stop. I added 3 more config options for the map in the video-layout.xml file.

amtriathlon:
> I moved the map initialization to the workout start and end. I added Context back but only using it for start and stop.

Perhaps I was not clear enough: my proposal was to add virtual Start/Stop methods to MeterMidget class to be called at workout start/stop passing context as a parameter, which do nothing for them, but for LiveMap can build the route, etc. not to spread LiveMap specific code all over VideoWindow, even exposing implementation details unnecessarily. 

> I added 3 more config options for the map in the video-layout.xml file.

Why do we need arbitrary initial lat/lon, just don't show the map when there are no coordinates.

human705:
I believe i misunderstood since i'm new to C++. I'll try to correct it. I was thinking that the initial coordinates would set the map to a "favorite"location for a user. I can remover them if they are unnecessary.

amtriathlon:
> I believe i misunderstood since i'm new to C++. I'll try to correct it.

Great, let me know if you need help.

> I was thinking that the initial coordinates would set the map to a "favorite"location for a user. I can remover them if they are unnecessary.

Yes, please.

human705:
If you can point me at an example of virtual methods it would be appreciated.

amtriathlon:
> If you can point me at an example of virtual methods it would be appreciated.

paintEvent is an example, the main difference here is the base class implementation will do nothing and derived classes don't need to call it. 

An alternative could be, once you consider this is finished and it will not change anymore, let me know and I can do that refactor as a separate commit.

human705:
Thanks Ale. Let me give it a try. If i can't do it, i'll clean up the unnecessary code and let you take over.

human705:
Hi Ale, one more question how does the virtual method  paintEvent get called? I don't see a reference of it in VideoWindow.cpp?

amtriathlon:
> Hi Ale, one more question how does the virtual method paintEvent get called? I don't see a reference of it in VideoWindow.cpp?

paintEvent is inherited from QWidget and called by Qt runtime, so the similarity ends here.

src/Train/VideoWindow.cpp
```diff
@@ -221,9 +221,9 @@ void VideoWindow::startPlayback()
         if (p_meterWidget->Source() == QString("LiveMap"))
         {
             LiveMapWidget* liveMapWidget = dynamic_cast<LiveMapWidget*>(p_meterWidget);
-            liveMapWidget->setContext(context);
-            liveMapWidget->initLiveMap();
+            liveMapWidget->startPlayback(context);
```
amtriathlon, line 224: p_meterWidget can be used to call startPlayback directly, no more need for conditional and the dynamic_cast

amtriathlon:
Now p_meterWidget can be used to call startPlayback/stopPlayback directly, no more need for conditional and dynamic_cast, that's the beauty of virtual methods.
This could be retrofitted to ElevationWidget to remove the need to call setContext on each update.

ericchristoffersen:
Good stuff. Peter, you should look up what dynamic_cast means... Its a really complicated mechanism to cast between classes. Stuff that is expensive and only extremely rarely necessary. Now that every meterwidget has a start and stop it should be called on them all. The base stub will just return. You can now move the videowindow calls to initlivemap, etc down into the livemap start method. The caller doesnt need to know details of the livemap, doesnt need to know what a livemapmeterwidget even is.

src/Train/VideoLayoutParser.cpp
```diff
@@ -213,6 +217,10 @@ bool VideoLayoutParser::endElement( const QString&, const QString&, const QStrin
             meterWidget->m_Angle = buffer.toFloat();
         else if (qName == "SubRange")
             meterWidget->m_SubRange = buffer.toInt();
+        else if (qName == "Zoom")
```
ericchristoffersen, line 220: Is ugly to have the livemap specific fields in the base class. They are there to be misused and sow confusion? Idea: move the livemap specific zoom and osmurl properties down into livemap object. Right here before you set use the dynamic_cast to get access to livemap object. If dynamic_cast gives you back a null then its not a livemap and you dont set anything. With this approach you should rename the properties to something specific to livemap, like LiveMap_zoom and LiveMap_osmURL, just so people are clear they are only for LiveMap widget.

src/Train/VideoWindow.cpp
```diff
@@ -218,6 +217,13 @@ void VideoWindow::startPlayback()
 
         p_meterWidget->raise();
         p_meterWidget->show();
+
+        if (p_meterWidget->Source() == QString("LiveMap"))
+        {
+            LiveMapWidget* liveMapWidget = dynamic_cast<LiveMapWidget*>(p_meterWidget);
+            liveMapWidget->startPlayback(context);
```
ericchristoffersen, line 224: Like ale says... you added a startPlayback method to the base. Now you should call startPlayback on every p_meterWidget, same as raise() and show(). No need to test or dynamic_cast.

src/Train/VideoWindow.cpp
```diff
@@ -237,8 +243,17 @@ void VideoWindow::stopPlayback()
 #ifdef GC_VIDEO_QT5
     mp->stop();
 #endif
-    foreach(MeterWidget* p_meterWidget , m_metersWidget)
+    foreach(MeterWidget * p_meterWidget, m_metersWidget)
+    {
+        if (p_meterWidget->Source() == QString("LiveMap"))
+        {
+            LiveMapWidget* liveMapWidget = dynamic_cast<LiveMapWidget*>(p_meterWidget);
+            liveMapWidget->stopPlayback();
```
ericchristoffersen, line 251: Question for Ale - anymore more familiar with qt. I suspect this is an architectural issue. The paint method is called from qt. Here the stopPlayback may be deleting data. Multithread issue with the delete? What is right way to synchronize?

liversedge: If a widget is being deleted in the event loop you can use deleteLater instead, it will be actually deleted after. 

I don't have time to look right now, but a widget processing a paint event after being deleted sounds very wrong.

ericchristoffersen: > If a widget is being deleted in the event loop you can use deleteLater instead, it will be actually deleted after.
> 
> I don't have time to look right now, but a widget processing a paint event after being deleted sounds very wrong.

Widget is not being deleted but there is specific data in the widget that is being deleted and may still be referenced by paint method.

liversedge: then set the pointer to null when deleted and check it is not null before using it?

ericchristoffersen: The problem is that one thread can check then grab the pointer and start using it (paint) and then the stop button is pressed and frees it. This is called a race around the check and is why there are things called atomics. Simple approach there needs to be a lock to protect readers from writers. There needs to be a convention. It could be there is a way for single access to be ensured by qt, but not if we call into the widgets directly. In my experience its not healthy to let this sort of stuff slide. It would be ok if we said no free/delete in media widget except for dtor. It would be ok if every resource held by our media widget was refcounted (delete when no more references). The easiest "no brainer" approach is to require every mediawidget to provide lock to be held when any method is called. Contention will be very rare. If contention ever became an issue we could do something smarter.

ericchristoffersen: Rule should be something like:

For any resource that is allocated or freed outside constructor or destructor, access (read or write) of that resource needs to be protected by a lock. General mediawidget doesn't need a lock but elevation and livemap widgets do.

human705:
Is there a reason Lat and Lon are set to zero when training is paused? It seems that other data fields int rtd like altitude maintain their value but LatLon is zeroed out. This causes the widget to get hidden when the workout is paused. I added code to fix it but it seems odd.

ericchristoffersen:
Zero lat/lon: i cant think why wed set to zero from non-zero on pause, but ive seen some routes with location gaps. Zero also means ‘no location’, good to test for so you dont annoyingly scan over to the equatorial atlantic. I think its ok to not handle that location well... Also Ive seen some routes start with a few zero locations.

amtriathlon:
Hi Peter, I did some tests and the maps doesn't show anymore, the error is something like "infinite tiles requested", besides this PR has become too complex, it cannot be automatically rebased anymore and now has conflicts with current master.
OTOH the initial functionality for tracking position works quite well, I used it several times, with different routes without issued, it even recovered automatically after an Internet disconnect of several minutes during a workout. I think it is a nice feature and is ready to merge.
So I decided to take that part with the code cleanup, adapt video-layout.xml providing a separate layout with the added LiveMap so the user can easily select to use it or not, and merged in https://github.com/GoldenCheetah/GoldenCheetah/commit/984d28abf0bd0aca6151d07d5531abab1ba44e4e
For the route functionality I think it is better to use another branch for a new PR, sorry for the inconvenience and thanks for your contributions!

human705:
Thanks Ale. I’ll create a new PR for the route and work on the errors. I must admit I haven’t seen that during my tests.

If you get a chance please review the LiveMap chart PR. It the same functionality in a chart.

PR #3487, opened 2020-06-08
Live map chart

human705:
This will provide the same functionality as the LiveMap overlay widget request 3482. 

image


src/Train/LiveMapWebPageWindow.cpp
```diff
@@ -106,6 +104,20 @@ LiveMapWebPageWindow::~LiveMapWebPageWindow()
 
 void LiveMapWebPageWindow::start()
 {
+    routeLatLngs = "[";
+    QString code = "";
+
+    for (int pt = 0; pt < context->currentErgFile()->Points.size()-1; pt++) {
```
ericchristoffersen, line 110: Where do you determine that the ergfile contains location information? It might be best to refuse to display map when there is no good location info in file. There's a function IsReasonableGeolocation in locationinterpolation.h.

human705: Good point Eric, i'll try to find an example of using "IsReasonableGeolocation".

ericchristoffersen:
Small nits: when I'm not riding it centers on 0,0 (atlantic off africa), and zoomed way out. Could we (a) make the default location better and (b) maybe not reset the values when a ride is finished?

Not sure what the best default would be, I'm sure there's a location people would enjoy. The top of the Stelvio pass? Maybe here: https://en.wikipedia.org/wiki/Poggio_di_San_Remo. Lots of possibilities that would bring people joy, at least better than the middle of the atlantic. Lan/Lon of the Poggio is: 43.827907, 7.814973.

human705:
Updated look with the route.
image

Chart settings to change map preferences 
image

amtriathlon:
Hi Peter, I resolved the conflict and merged current master to be able to test.
It works for me and I liked the general behavior except for the "default location" feature, I think it would be better just to indicate we have no data like RideMapWindow does, instead of showing an arbitrary location.
Related to this, we could use the ergFileSelected signal to draw the route -in a similar way to what WorkoutPlotWindow does with the altitude profile- instead to wait to the start signal, to allow route preview before starting the workout.
WRT to settings, GC charts have a general mechanism to remember them via properties without the need to use settings, this is simpler and settings are included if you export or upload the chart.

amtriathlon:
Here is how Workout chart uses ergFileSelected:
https://github.com/GoldenCheetah/GoldenCheetah/blob/b2da0afb83dcc1d897691942811e3b871247c999/src/Train/WorkoutPlotWindow.cpp#L38
Although, since context->currentErgFile() is updated before the signal is sent, you could connect that signal to a slot without parameters to minimize changes to your code.

amtriathlon:
And here is how WebPageWindow uses a property to store the URL:
https://github.com/GoldenCheetah/GoldenCheetah/blob/b2da0afb83dcc1d897691942811e3b871247c999/src/Train/WebPageWindow.h#L62

human705:
Hi Ale,

Should i continue the work on this branch or create a new one?

Peter

amtriathlon:
You can continue on this one if you want, we can squash all changes in a consolidated commit at merge time.

amtriathlon:
It works very well in my testing, I've found minor uses at chart creation and fixed initialization to ensure current settings are considered when the chart is added, also changed to use the standard blank chart when no lat/lon data available.
Automatic merging was no longer possible so rebased and pushed as a single commit: https://github.com/GoldenCheetah/GoldenCheetah/commit/4c19365e8b7588dd484401042b3b3852b3f230bf
Thank you for your work!

PR #3506, opened 2020-06-16
Allow Text meter widgets to be of same size and well aligned

michelrdagenais:
The various MeterWidgets are nice but I like my numbers in text and well aligned. I had several problems:

- Even with several entries of exactly the same specified size and font, some had smaller characters than others (notably the speed).
- The entries were centered and thus changed horizontal position with the number of digits.
- Even when the number of digits does not change, a 1 is thinner than a 5 and the text moves left and right.
- Some metrics do not come with the units and it is difficult to position a text for that properly, besides the metrics entry.
- Depending on the video underneath, some entries could be difficult to read, a transparent darker background helps readability.

In summary, here is a picture of the result:
gctextwidget

And the associated video-layout file:
[video-layout.xml.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/4787606/video-layout.xml.txt)

The y scale is computed from the font instead of only from the current content, to avoid having different scales between different Text widgets of the same size. For example, "kph" has a high k and low p, resulting in a text widget with bigger bounding box, and thus smaller scale than a text widget with "watts". 

Options are added for alignment and for text width, and the fonts are set to fixed width, to help align the different entries. The formatting of metrics, with the integer part as Text, and the fraction as AltText is more systematic, and it is possible to add an AltTextSuffix, to specify the units when they are not provided in VideoWindow. The visibility of the BoundingRect and Background can be controlled.

As another user suggested on the forums, the desired layout for a video and non-video workout may be very different. It would be useful to have the ability to quickly select between different train-layout files as a future feature.

### DIFF
--- a/src/Train/MeterWidget.cpp
+++ b/src/Train/MeterWidget.cpp
@@ -51,6 +51,7 @@ MeterWidget::MeterWidget(QString Name, QWidget *parent, QString Source) : QWidge
     m_Angle = 180.0;
     m_SubRange = 10;
     boundingRectVisibility = false;
+    backgroundVisibility = false;
     forceSquareRatio = true;
 }
 
@@ -114,19 +115,29 @@ QSize MeterWidget::minimumSize() const
 void MeterWidget::paintEvent(QPaintEvent* paintevent)
 {
     Q_UNUSED(paintevent);
+    if(!boundingRectVisibility && !backgroundVisibility) return;
+
+    QPainter painter(this);
+    painter.setClipRegion(videoContainerRegion);
+    painter.setRenderHint(QPainter::Antialiasing);
+
+    int radius = qMin(m_Width, m_Height) * 0.1;
+    if(backgroundVisibility) painter.setBrush(QBrush(m_BackgroundColor));
+    else painter.setBrush(Qt::NoBrush);
+
     if (boundingRectVisibility)
     {
-        m_OutlinePen = QPen(boundingRectColor);
+        m_OutlinePen = QPen(m_BoundingRectColor);
         m_OutlinePen.setWidth(2);
         m_OutlinePen.setStyle(Qt::SolidLine);
 
-        //painter
-        QPainter painter(this);
-        painter.setClipRegion(videoContainerRegion);
-        painter.setRenderHint(QPainter::Antialiasing);
-
         painter.setPen(m_OutlinePen);
-        painter.drawRect (1, 1, m_Width-2, m_Height-2);
+        painter.drawRoundedRect (1, 1, m_Width-2, m_Height-2, radius, radius);
+    }
+    else
+    {
+        painter.setPen(Qt::NoPen);
+        painter.drawRoundedRect (0, 0, m_Width, m_Height, radius, radius);
     }
 }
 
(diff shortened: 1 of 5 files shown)


human705:
I like that look!!! This leaves plenty of room for the other graphical widgets and still enjoy the video during the long winters. I'll try it ASAP.

amtriathlon:
> As another user suggested on the forums, the desired layout for a video and non-video workout may be very different. It would be useful to have the ability to quickly select between different train-layout files as a future feature.

The layout file allows to have multiple layouts, but this feature is not used, we could add configuration at the chart level (VideoWindow) to select between the available layouts and no widgets. I think this could be reasonable compromise between usability and implementation simplicity.

michelrdagenais:
> The layout file allows to have multiple layouts, but this feature is not used

Yes, I noticed. Doing a first pass to list the available layouts in the file would be easy. Then, parsing only the layout requested by the user would also be a simple change for the video-layout. For the train-layout, which is editable interactively and must be saved, replacing only the modified layout could be slightly more involved.

> we could add configuration at the chart level (VideoWindow) to select between the available layouts and no widgets. I think this could be reasonable compromise between usability and implementation simplicity.

Yes, once you have the list of layouts, it should be relatively easy. I finally found where chart settings were set, in each chart type as "controls" while the main forms for Chart Settings and Chart Setup are elsewhere.

My first plan, however, was to follow this pull request with documentation in the wiki under https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_ScopeBar_Views about how to define a video layout.

amtriathlon:
> Yes, I noticed. Doing a first pass to list the available layouts in the file would be easy. Then, parsing only the layout requested by the user would also be a simple change for the video-layout. 

I think this is worth considering, since most users will not be able or willing to edit the video-layout.xml, and to be able to choose between a set of predefined layouts would be a great enhancement.

>For the train-layout, which is editable interactively and must be saved, replacing only the modified layout could be slightly more involved.

I would suggest to wait for this since it is something it is evolving, the train layout is shared with the remaining views (main difference is Trend and Analysis default to Tabbed and Train to Tiled mode), and Mark is working on a replacement.

For the moment a simple working solution is to have a layout similar to the default one for workouts without videos occupying all screen space, plus a single Video Window with all the interesting overlay widgets for the video case, also occupying all screen space, and simply switch between the two with page up/down or scrolling.

> My first plan, however, was to follow this pull request with documentation in the wiki under https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_ScopeBar_Views about how to define a video layout.

Great! Documentation for the Train View is sparse, mostly in the form of video tutorials in http://www.goldencheetah.org/#section-tutorials, and recently a FAQ: https://github.com/GoldenCheetah/GoldenCheetah/wiki/FAQ#train, but the users guide section is empty.

michelrdagenais:
> I think this is worth considering, since most users will not be able or willing to edit the video-layout.xml, and to be able to choose between a set of predefined layouts would be a great enhancement.

I issued a pull request for that. It took some time to understand how the settings menu interacts with the charts, but otherwise the rest was fairly straightforward. There were problems as well when instantiating a video player while the video file was already selected, or when adding or editing a video player while the training session is running, but those could be solved with little additional code. Depending on which pull request goes in first, I may have to resolve small conflicts.

PR #3511, opened 2020-06-18
Untranslatable strings

MaksVasilev:


MaksVasilev:
Additional: cosmetic problems with displaying QLabel in grid:

GC-VBS-02
GC-VBS-01

PR #3515, opened 2020-06-19
Add control to Video Player to select among several Meter Widget layouts

michelrdagenais:
The video layout file is extended to contain possibly several named layouts. The file is read to list the layouts and offer a selection in the Video Player chart settings menu. The file is then read again to
instantiate the selected layout.

This was a suggestion of Alejandro Martinez. Interestingly, you can edit the settings and change the layout of the Video Viewer even while the training session is underway. You can add a video viewer as well while the training session is running. Your video layout choice is saved in your training layout. A video-layout.xml resource file is provided with 3 different layouts if you want to test it (move elsewhere your current ~/.goldencheetah/User/config/video-layout.xml and the new file will automatically be provided as default the next time you use a video viewer). The provided "Text Widget" layout does not look too good currently but should be nice with the Meter Widget enhancements pull request that I proposed a few days ago.

### DIFF
--- a/src/Core/Context.h
+++ b/src/Core/Context.h
@@ -97,6 +97,7 @@ class Context : public QObject
         DateRange dr_;
         ErgFile *workout; // the currently selected workout file
         VideoSyncFile *videosync; // the currently selected videosync file
+        QString videoFilename;
         long now; // point in time during train session
         SpecialFields specialFields;
 
@@ -155,7 +156,7 @@ class Context : public QObject
         void notifyVideoSyncFileSelected(VideoSyncFile *x) { videosync=x; videoSyncFileSelected(x); }
         ErgFile *currentErgFile() { return workout; }
         VideoSyncFile *currentVideoSyncFile() { return videosync; }
-        void notifyMediaSelected( QString x) { mediaSelected(x); }
+        void notifyMediaSelected( QString x) { videoFilename = x; mediaSelected(x); }
         void notifySelectVideo(QString x) { selectMedia(x); }
         void notifySelectWorkout(QString x) { selectWorkout(x); }
         void notifySelectVideoSync(QString x) { selectVideoSync(x); }
--- a/src/Resources/xml/video-layout.xml
+++ b/src/Resources/xml/video-layout.xml
@@ -1,4 +1,5 @@
-<layout name="preset_1">
+<layouts>
+<layout name="Graphical Meters">
    <meter name="Speedometer" container="Video" source="Speed" type="NeedleMeter">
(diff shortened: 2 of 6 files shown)


amtriathlon:
Thanks Michel, I merged this first so the others we have pending can be adapted.
A nice addition would be a "Reset to Default" button to remove config/video-layout.xml and load the default, mainly oriented to users not comfortable with config files who just want to use the newest layout.

michelrdagenais:
> Thanks Michel, I merged this first so the others we have pending can be adapted.

I will udate the TextMeterWidgetEnhancements. The ControllerSetWeight patch is not impacted.

> A nice addition would be a "Reset to Default" button to remove config/video-layout.xml and load the default, mainly oriented to users not comfortable with config files who just want to use the newest layout.

What is the "common" way to do it in GC? I can think of at least 3 plausible options:
- simply erase video-layout.xml and replace with builtin defaults
- remove video-layout.xml.bak if it exists, move video-layout.xml to video-layout.xml.bak and replace video-layout.xml with the builtin defaults. GC does something similar in src/Gui/SaveDialogs.cpp and src/Gui/SplitActivityWizard.cpp.
- mv video-layout.xml to video-layout.xml.date.bak where "date" is the current date with enough resolution to insure unicity. You do not loose anything but versions can accumulate.

amtriathlon:
> I will udate the TextMeterWidgetEnhancements. The ControllerSetWeight patch is not impacted.

The other affected was LiveMap, but it is already fixed and merged in its basic functionality for position tracking.
 
> > A nice addition would be a "Reset to Default" button to remove config/video-layout.xml and load the default, mainly oriented to users not comfortable with config files who just want to use the newest layout.
> 
> What is the "common" way to do it in GC? I can think of at least 3 plausible options:
> 
> * simply erase video-layout.xml and replace with builtin defaults
> * remove video-layout.xml.bak if it exists, move video-layout.xml to video-layout.xml.bak and replace video-layout.xml with the builtin defaults. GC does something similar in src/Gui/SaveDialogs.cpp and src/Gui/SplitActivityWizard.cpp.
> * mv video-layout.xml to video-layout.xml.date.bak where "date" is the current date with enough resolution to insure unicity. You do not loose anything but versions can accumulate.

For Reset Layout(s) we are using 1), see: https://github.com/GoldenCheetah/GoldenCheetah/blob/dbecc005b5b7572cacfe8d4333d9959f7a7cb71a/src/Charts/HomeWindow.cpp#L1371
with the ability to get (supposedly) more updated versions from goldencheetah.org
For activities 2), but I think it is overkill in this case.

Issue #3513, opened 2020-06-18
Untranslatable strings in Pages.cpp, RLibrary.cpp, can't simple fix by tr()


MaksVasilev:
I try use tr() in "static const SimBicyclePartEntry arr[] = {}" (on 2225 line), but have conversion string type error:

```
Gui\Pages.cpp(2251): error C2440: 'initializing': cannot convert from 'QString' to 'const char *'
Gui\Pages.cpp(2251): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
```

Commit with error:
https://github.com/MaksVasilev/GoldenCheetah/commit/1b26b223b64bc759a56f1b6eac1cc73ac5a58793

amtriathlon:
The problem is the code added by https://github.com/GoldenCheetah/GoldenCheetah/pull/3332 doesn’t comply with the first requirement to enable translations (Using QString for All User-Visible Text): https://doc.qt.io/qt-5/i18n-source-translation.html

MaksVasilev:
Addition:

Untranslatable strings in RLibrary.cpp like as:
`        errors << "R_HOME has not been configured in options or the system environment"
                  " so we looked in the common places to find the R install.\n\n";
`

In all operation with QString "errors".

MaksVasilev:
Addition:

Untranslatable strings in **Charts/Overview.cpp** like as:

`            add = new MetaOverviewItem(space, "Sport", "Sport");
            space->addItem(2,0,5, add);

            add = new MetaOverviewItem(space, "Workout Code", "Workout Code");
            space->addItem(3,0,5, add);

            add = new MetricOverviewItem(space, "Duration", "workout_time");
            space->addItem(4,0,9, add);
`

amtriathlon:
> Addition:
> 
> Untranslatable strings in RLibrary.cpp like as:
> ` errors << "R_HOME has not been configured in options or the system environment" " so we looked in the common places to find the R install.\n\n";`
> 
> In all operation with QString "errors".

This is explained in the same document under "Translating Non-Qt Classes", we have used that solution in other places to enable tr() macro:
```
class MyClass
{
    Q_DECLARE_TR_FUNCTIONS(MyClass)

public:
    MyClass();
    ...
};
```

amtriathlon:
> Addition:
> 
> Untranslatable strings in **Charts/Overview.cpp** like as:
> 
> ```
>             space->addItem(2,0,5, add);
> 
[...]

PR #3518, opened 2020-06-20
Make BicycleSim setting page translatable

ericchristoffersen:
Change visible strings to use QString and be obtained via tr() so that they can be translated.


ericchristoffersen:
This fixes part of 3513:

https://github.com/GoldenCheetah/GoldenCheetah/issues/3513

PR #3520, opened 2020-06-22
Add EPM - Ergo Planet Movie format import

amtriathlon:
To be used as workouts and videosync with free Ergo Planet Videos
EPM files contain video sync plus geolocation information in XML format
Library of free rides: http://wiki.ergoplanet.de/myor/roadmovie-galerie

The problem is altitude set to zero in those examples, likely because it is
redundant when you resort to the EPP (Ergo Planet Program) for slope data,
but the EPP is an unpublished binary format we don't support directly.

So the workflow to use the those synchronized videos is:
1) Import the .epm file to GoldenCheetah as an activity
2) Use Fix Elevation and Fix GPS to add and smooth elevation and route data
3) Export the conditioned activity to GoldenCheetah JSON format
4) Import the .json file as Workout and VideoSync files,
   plus the .avi as media in GC Train mode using Scan Workouts and Media
5) Train using a Video Window with overlay widgets and simulated speed for
   better experience.

### DIFF
--- a/src/FileIO/EpmParser.cpp
+++ b/src/FileIO/EpmParser.cpp
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2020 Alejandro Martinez (amtriathlon@gmail.com)
+ *               Based on TcxParser.cpp
+ *
+ * Copyright (c) 2008 Sean C. Rhea (srhea@srhea.net),
+ *                    J.T Conklin (jtc@acorntoolworks.com)
+ *
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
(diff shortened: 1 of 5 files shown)


TWINGSISTER:
I found that this epm does not get correct elevations with "Fix Elevation" just one value, the same,  for all point. 
[Davos-Albula.prf - Copy.epm.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/6445882/Davos-Albula.prf.-.Copy.epm.txt)
May be this could be of some interest.

TWINGSISTER:
Actually I found out that, most of the time, your suggestion works in converting videos in this HUGE repository of FREE VRLs  [here](https://web.archive.org/web/20210404222758/http://ergo.ub-online.de/extraHTML/rlvs-bikeboard.html). I found this so interesting that I turned this into a video [here](http://primi235711.altervista.org/moodle/mod/scorm/view.php?id=248) and [here on YT](https://youtu.be/ky60dOun_ck)

PR #3538, opened 2020-07-10
Computrainer Button tidy up

amtriathlon:
Since F3 was assigned to Calibration, its use as modifier was deprecated.
Also nextDisplay is not implemented so we can do some cleanup.
F1 is restored to their original function: start/pause and F2 is used
for new lap, in both cases with debouncing to avoid false activations.
This way the same functions enabled for ANT+ remotes are available via HBC:
Start&Pause: F1
Lap/Interval: F2
Load increase/decrease: +/-
Calibration: F3
Stop: Reset

### DIFF
--- a/src/Train/ComputrainerController.cpp
+++ b/src/Train/ComputrainerController.cpp
@@ -23,6 +23,8 @@
 ComputrainerController::ComputrainerController(TrainSidebar *parent,  DeviceConfiguration *dc) : RealtimeController(parent, dc)
 {
     myComputrainer = new Computrainer (parent, dc ? dc->portSpec : ""); // we may get NULL passed when configuring
+    f1Depressed = false;
+    f2Depressed = false;
     f3Depressed = false;
 }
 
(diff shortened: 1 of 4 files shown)


ericchristoffersen:
Are there screen buttons to achieve this same functionality? Downside of function keys is that they aren't available to press on windows 10 tablet, I don't know how to use them from the virtual keyboard.

amtriathlon:
> Are there screen buttons to achieve this same functionality?

Yes, in the Intensity Adjustments and Workout Control bottom bar:

image


> Downside of function keys is that they aren't available to press on windows 10 tablet, I don't know how to use them from the virtual keyboard.

F1-3, +/- and Reset here refer to the 6 buttons on the Computrainer Handlebar Controller, not the PC keyboard.

amtriathlon:
This works very well for me, but doesn't provide access to ffwd/rewind, an alternative would be to use F2 as a modifier:
ffwd: F2+plus
rewind: F2+minus
new lap: F2+F1(start)
I will give it a try.

PR #3546, opened 2020-07-27
make the date range of relative seasons (e.g. in Trends) update as time passes

### LINKED ISSUE #1751: Trends date ranges do not update as time passes
I very rarely close GoldenCheetah (maybe I should). I last opened it sometimes last week, so now when I click "this week" in the metrics tab I get last week.

qheath:
make the season limits (start and end) dynamic, and have 3 ways of specifying them:
- fixed season (e.g. "2018"):
  * start
  * end
- season relative to the end (e.g. "Last 28 days"):
  * length
  * end (defaults to today)
- season relative to the start (e.g. "Last Year"):
  * length
  * offset of the start with respect to a reference date (which defaults to today)

this has 2 benefits:
- the season logic is pushed in the season code, and we can call getStart and getEnd without checking first if the season is fixed or relative
- if the season is relative, getStart and getEnd are relative to the date when they are called, not when the season was created (fixes #1751)

also remove unused code related to seasons:
- CompareDateRange::days
- Season::days
- Seasons::seasonFor


amtriathlon:
I didn't look closely at the code, but the build error is likely due to the need to apply to src/Python/SIP/Bindings.cpp similar changes to those of src/R/RTool.cpp. This goes unnoticed when building without Python embedding.

PR #3573, opened 2020-08-15
Fix issue 3571: segment matching

conover-s:
Updates in Route.cpp to address the problem described in issue 3571.

[0001-Fix-issue-3571-segment-matching.patch.gz](https://github.com/GoldenCheetah/GoldenCheetah/files/5079807/0001-Fix-issue-3571-segment-matching.patch.gz)


amtriathlon:
Hi Scott, your solution seems a good one, I tested the provided file plus regression testing with previously working routes and it seems to be ok, so I merged it for inclusion in next builds.
Thanks for your contribution!

PR #3579, opened 2020-08-26
Adjust time units to compute speed properly for both Cycling Speed an…

michelrdagenais:
…d Cycling Power Bluetooth profiles

Both the CyclingSpeedAndCadence and the CyclingPower Bluetooth profiles have pretty much the same encoding for the wheel position and cadence data. Thus, the same function is used to read from Bluetooth and update those values for the telemetry. For the wheel position, this is in getWheelRpm. However, the time unit granularity differs between the two profiles, Cycling Speed reports wheel event times in 1/1024th seconds while Cycling Power uses 1/2048th seconds. We now define and use the wheelTimeUnit field to account for the difference.


michelrdagenais:
Another similar fix was merged just before this one

PR #3590, opened 2020-09-07
Bison 3.7

MeisterP:
This breaks backwards compatibility with bison versions older than 3.7
Probably not ready for merging. Still leaving it here for people who run into build issues.


flokli:
I can confirm this fixes things for Bison >= 3.7.

To fix the build, I applied the patch in nixpkgs on top of V3.5 - however it'd be nice if this would land upstream.

@liversedge, what's needed to get this merged?

amtriathlon:
> what's needed to get this merged?

This change brakes ci builds on the 3 supported platforms, until this is fixed it is not ready for merging, IMHO.

flokli:
Hm, CI currently is just lucky to be at an older version of Bison that works.
This will break not only developers with distros using a more recent version of bison (Ubuntu Groovy, Debian Sid, Archlinux, Gentoo), but also Travis, as soon as the package updates have trickled into there (which might take some more time).

Reading back on the thread linked from https://github.com/GoldenCheetah/GoldenCheetah/pull/3590/commits/0bedd5ac9213b72bea540d347db9cb12e6653aed and https://lists.gnu.org/archive/html/bug-bison/2020-08/msg00026.html, it seems it's not really possible to make the `.y` files compatible with multiple versions of bison (both pre 3.7 and >=3.7).

Their proposed course of action is committing the generated files alongside the `.y` files in the source repo, so `bison` is only needed to regenerate those files (in whatever version the project currently is using).

I agree that's not pretty, but it's probably better than keeping it broken on those distros.

liversedge:
Not gonna merge something guaranteed to break all our CI builds !

flokli:
> Not gonna merge something guaranteed to break all our CI builds !

Yes, I agree :-)

I did dig a bit further - turns out e1f42f8b3340eb4695ad73be764332e75b7bce90 should be all that's needed, and https://lists.gnu.org/archive/html/bug-bison/2020-08/msg00014.html seems to be a red herring, at least for GoldenCheetah.

I was able to successfully compile GoldenCheetah V3.5 with e1f42f8b3340eb4695ad73be764332e75b7bce90 applied on top, both with bison 3.5.4 and 3.7.1 in scope.

@MeisterP can you drop 0bedd5ac9213b72bea540d347db9cb12e6653aed from this PR and give CI another run?

liversedge:
Looks like a safe solution, thanks.. ping here when ready to merge a backwards compatible fix?

MeisterP:
> I was able to successfully compile GoldenCheetah V3.5 with (only) [e1f42f8](https://github.com/GoldenCheetah/GoldenCheetah/commit/e1f42f8b3340eb4695ad73be764332e75b7bce90) applied on top, both with bison 3.5.4 and 3.7.1 in scope.

I can't easily test version 3.5.4 but you are right, it is working with `sys-devel/bison-3.6.4`. So the version check could test for a lower version indeed. (probably 3.5? I'm not going to bother to test every bison release...)
It still fails here with `sys-devel/bison-3.1` though.

The ci currently uses the following versions:
- windows is using bison 2.7
- osx is using bison 2.3
- linux is using bison 3.0.4



with `sys-devel/bison-3.1` and **without** https://github.com/GoldenCheetah/GoldenCheetah/commit/0bedd5ac9213b72bea540d347db9cb12e6653aed
```
rm -f DataFilter_yacc.h DataFilter_yacc.cpp
yacc -d -p DataFilter -b DataFilter Core/DataFilter.y
Core/DataFilter.y: warning: 3 shift/reduce conflicts [-Wconflicts-sr]
Core/DataFilter.y:53.9-26: error: %define variable 'api.header.include' is not used
 %define api.header.include {"DataFilter_yacc.h"}
         ^^^^^^^^^^^^^^^^^^
make[1]: *** [Makefile:13794: DataFilter_yacc.h] Error 1
make[1]: *** Waiting for unfinished jobs....
```

with `sys-devel/bison-3.1` and **with** https://github.com/GoldenCheetah/GoldenCheetah/commit/0bedd5ac9213b72bea540d347db9cb12e6653aed
```
rm -f DataFilter_yacc.h DataFilter_yacc.cpp
yacc -d -p DataFilter -b DataFilter Core/DataFilter.y
Core/DataFilter.y:52.10-14: error: require bison 3.7, but have 3.1
 %require "3.7"
          ^^^^^
make[1]: *** [Makefile:13794: DataFilter_yacc.h] Error 63
make[1]: *** Waiting for unfinished jobs....
```

You can drop https://github.com/GoldenCheetah/GoldenCheetah/commit/0bedd5ac9213b72bea540d347db9cb12e6653aed. It's just there to give a more meaningful error message.

MeisterP:
> probably 3.5? I'm not going to bother to test every bison release...

bison 3.4 should work as well according to the documentation.

```
Directive: %define api.header.include {<header.h>}
History: Introduced in Bison 3.4. Defaults to ‘"basename.h"’ since Bison 3.7, unless the header file is y.tab.h. 
```

flokli:
2 of the three `.y` files say this:

> // This grammar should work with yacc and bison, but has
> // only been tested with bison. In addition, since qmake
> // uses the -p flag to rename all the yy functions to
> // enable multiple grammars in a single executable you
> // should make sure you use the very latest bison since it
> // has been known to be problematic in the past. It is
> // know to work well with bison v2.4.1.

which is a bit inconsistent with the versions of bison currently present in CI (thanks @MeisterP for digging):


> - windows is using bison 2.7
> - osx is using bison 2.3
> - linux is using bison 3.0.4

IMHO, CI should more closely mirror the versions it expects compatibility with.

Current Linux CI uses bison 3.0.4, because that's what comes with Ubuntu 16.04 / xenial, which will be EOL in April 2021 anyways.

It might be sufficient (and desirable anyways) to point travis to a more recent LTS Ubuntu, like Focal, which will come with Bison 3.5.1, which should be recent enough.

I'm not entirely sure about osx, and how that bison version is installed - given homebrew  provides bison 3.7.2, we might just be able to add `brew install bison@3.7` to `travis/osx/before_install.sh`.

Windows `bison` seems to be provided by a custom tarball, which got introduced in bcd0ccd56b1a386a60882d17e2b32ef548032d0d. Maybe we can revert this and add `choco install winflexbison3` again.

I'll give this a try and post a PR shortly to test it.

liversedge:
MacOS Xcode 12 developer tool installs Bison 2.3 (this is the latest on Catalina)

amtriathlon:
> Current Linux CI uses bison 3.0.4, because that's what comes with Ubuntu 16.04 / xenial, which will be EOL in April 2021 anyways.
> 
> It might be sufficient (and desirable anyways) to point travis to a more recent LTS Ubuntu, like Focal, which will come with Bison 3.5.1, which should be recent enough.

We are using CI not just for testing purposes but to generate the binaries for official releases, Ubuntu 16.04 is the recommended platform to generate an AppImage covering most of currently supported versions, when it reaches EOL the plan is to switch to 18.04, we could make that move in advance since v3.6 will not be released before, but I think to move to 20.04 is not a good idea.

liversedge:
18.04 uses bison 3.04, 20.04 is 3.5-- bottom line here is that for the most part, and for some time to come, this is pretty much a non-issue

flokli:
> We are using CI not just for testing purposes but to generate the binaries for official releases, Ubuntu 16.04 is the recommended platform to generate an AppImage covering most of currently supported versions, when it reaches EOL the plan is to switch to 18.04, we could make that move in advance since v3.6 will not be released before, but I think to move to 20.04 is not a good idea.

Thanks for the explanation.

>  bottom line here is that for the most part, and for some time to come, this is pretty much a non-issue

Building GoldenCheetah currently fails in environments with more recent (>=3.7) versions of bison (Debian Sid, Gentoo, Fedora Rawhide, Archlinux). This can be developer machines, downstream distro packaging environments etc.

The fixes proposed here will make it build on all environments with at least Bison 3.4.

I included the commits from this PR and updated CI to use Bison >=3.4 everywhere (keeping xenial for linux), alongside with some documentation updates and submitted https://github.com/GoldenCheetah/GoldenCheetah/pull/3628.

The CI builds are currently looking very promising. I propose closing this PR and carrying the discussion over there.

amtriathlon:
I am closing this because it cannot be merged without breaking CI builds and it is included in https://github.com/GoldenCheetah/GoldenCheetah/pull/3628

PR #3600, opened 2020-09-09
Lat and Lon now get DegreesToRadians

ericchristoffersen:
Convert lat/lon degrees to radians to determine rider bearing. Only matters when wind is specified.

Issue described in this thread:

https://groups.google.com/g/golden-cheetah-users/c/hECb3KdRPQU

### DIFF
--- a/src/FileIO/FixDerivePower.cpp
+++ b/src/FileIO/FixDerivePower.cpp
@@ -28,6 +28,12 @@
 #define MATHCONST_PI 		    3.141592653589793238462643383279502884L /* pi */
 #endif
 
+static double DegreesToRadians(double degrees) {
+    static const double degToRad = MATHCONST_PI / 180.;
+
+    return degrees * degToRad;
+}
+
 // Config widget used by the Preferences/Options config panes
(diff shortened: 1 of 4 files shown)


ericchristoffersen:
#3601

amtriathlon:
Thanks for the fix Eric!

The same error is present in Fix Running Power:
https://github.com/GoldenCheetah/GoldenCheetah/blob/01f1e0e9bd2c436f0d8af39302ee50e60ca4866d/src/FileIO/FixRunningPower.cpp#L223

May be we could add DegreesToRadians to Core/Utils and use it in both places.

ericchristoffersen:
> Thanks for the fix Eric!
> 
> The same error is present in Fix Running Power:
> https://github.com/GoldenCheetah/GoldenCheetah/blob/01f1e0e9bd2c436f0d8af39302ee50e60ca4866d/src/FileIO/FixRunningPower.cpp#L223
> 
> We have Core/Utils and Core/Units which could be used for common utilities.

I ended up consolidating the bearing calc into geolocation class.

PR #3636, opened 2020-10-17
Support Kurt Kinetic Inride and SmartControl

ericchristoffersen:
Support Kurt Kinetic Inride and SmartControl trainers over bluetooth.

I have only tested the Inride 2.0 support with windows 10.

### DIFF
--- a/src/Train/BT40Controller.cpp
+++ b/src/Train/BT40Controller.cpp
@@ -192,6 +192,61 @@ BT40Controller::getCalibrationType() {
     return CALIBRATION_TYPE_NOT_SUPPORTED;
 }
 
+uint8_t
+BT40Controller::getCalibrationState() {
+    for (auto* dev : devices) {
+        uint8_t caltype = dev->getCalibrationType();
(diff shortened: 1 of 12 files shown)


ericchristoffersen:
Latest change I put a CalibrationData object into the bt40 device and implement query through bt40controller. This lets standard trainsidebar calibration to work with bt40. Taught the kurt devices to populate the CalibrationData.

Added two new calibration_state_failure types, one for spindown too fast, another for spindown too slow. This info about spindown calibration attempt needs a way to be bubbled back to user.

This seems to work pretty well for kurt inride, and should also work for smart_control but I don't have that device to test.

Anyway, I've good confidence in the behavior of the inride support, even the calibration works really well.

ericchristoffersen:
#3641

amtriathlon:
Is this ready to merge, Eric?

ericchristoffersen:
> Is this ready to merge, Eric?

Yes, I think so. Good to get mileage on the changes to calibration. I know InRide V2 works great. I don't have a smart_control device to test - guy on group said it 'worked'.

Issue #3640, opened 2020-10-23
Elite Turbo Muin Speed v3.6 Sep 2020


G1032:
Current speed incorrect for Elite Turbo Muin
Probably 10x faster than actual
100rpm in 39x12 is over 200km/HR
Wheel circumference set correctly
Selecting either Turbo Muin 2013 or 2015 for preset virtual power curve doesn't fix. 
No issues with speed on version 3.5

amtriathlon:
This is a bug in Add Device Wizard, likely introduced with the new virtual power curves feature: when you use a custom wheel perimeter as required for this trainer, instead of select an standard rim/tire combination, wheel size is set to zero for the newly created device.

PR #3643, opened 2020-10-25
Fix Error Setting Custom Wheel Diameter

ericchristoffersen:
Virtual Power change #3410 introduced this error, where wheelsize value was only used if it was set via dropdowns. This change fixes the bad behavior, edit value is used if it is a valid number, whether it came from a dropdown or not.


ericchristoffersen:
#3640

Issue #3650, opened 2020-10-29
Virtual Power Trainers Should Track Mass Moment of Inertia


ericchristoffersen:
Currently the virtual power trainers are implemented with a straight mapping from speed to power. This is correct for constant speed but not for any time trainers rotational speed is changing.

Trainers have flywheels. Wheel-on trainers have wheels that spin. Those spinny things take energy to get them spinning. Propose to add a field to the virtualtrainer to hold a setup's Mass Moment Of Inertia. If that value is left at zero then current behavior is unchanged. If it is set to a number the virtual power calc will bias the power calculation by the acceleration that has occured since the last sample. This will increase reported watts when user is accelerating, and decrease reported watts when they are coasting.

If mass moment of inertia is set correctly, and speed/power curve is correct then coasting will show 0 watts emitted by user during a spindown.

PR #3651, opened 2020-10-29
rotational inertial for custom virtual power

ericchristoffersen:
Adds field to hold trainers rotational inertia. This allows power
calculation to track energy that passes in and out of trainers
flywheel so that acceleration and deceleration power can be
reported immediately.

All existing defined trainers are using 0kg for inertia. Probably
most should be updated with user data.

#3650

### DIFF
--- a/src/Train/AddDeviceWizard.cpp
+++ b/src/Train/AddDeviceWizard.cpp
@@ -47,6 +47,7 @@ AddDeviceWizard::AddDeviceWizard(Context *context) : QWizard(context->mainWindow
 
     virtualPowerIndex = 0;           // index of selected virtual power function
     wheelSize = 0;
+    rotationalInertiaKG = 0;
     strideLength = 0;
 
     // delete when done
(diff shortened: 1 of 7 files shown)


ericchristoffersen:
PS: This change also includes some polyfix improvements: slope calc for polyfit and some numeric sanity check unit tests.

ericchristoffersen:
Is totally wrong to put a rpm delta timer in the rtcontroller, need to find a better way.

ericchristoffersen:
Closing this until I get it working well.

amtriathlon:
> PS: This change also includes some polyfix improvements: slope calc for polyfit and some numeric sanity check unit tests.

Since you are reworking this, please separate fixes from new features.

ericchristoffersen:
> > PS: This change also includes some polyfix improvements: slope calc for polyfit and some numeric sanity check unit tests.
> 
> Since you are reworking this, please separate fixes from new features.

Yes, totally. I have some other code that isn't ready (spindown stuff) that relies on the slope calcs - but the tests are good. Do you think its ok to submit the slope calc and tests without anything that uses them?

amtriathlon:
>Do you think its ok to submit the slope calc and tests without anything that uses them?

No, I think it is better to introduce new code only when it is required to fix an error or implement a new feature, so it can be tested.

PR #3653, opened 2020-11-03
Rotational Inertial for Custom Virtual Power

ericchristoffersen:
Part 1: Rotational Inertial for Custom Virtual Power

This change is only enabled when device is defined with non-zero
Inertial Moment. When Inertial Moment is not defined or is
zero then this change has no effect.

Change adds field to device to hold trainer's inertial moment
in (KG M^2).

This allows power calculation to track energy that passes in
and out of trainer's flywheel so that acceleration and
deceleration power can be reported immediately.

A simple example of the benefit, suppose a trainer with
a speed->power curve that maps 100 watts to 20kph.
With this change if rider accelerates his trainer quickly from 0
to 20kph we will compute that rider emitted additional
watts for the time that they were accelerating the system.
In this example once 20kph is reached and maintained the
reported watts would drop to 100watts.

Whenever device sets a new rpm sample a time is
recorded alongside which allows the common controller
to compute change in rpm over time, which allows
calculation of kinetic energy that has entered or
left the trainers flywheels during the duration.

It is a bit of math to determine a correct 'I' for a trainer's
flywheels, especially if the flywheels have different rpms,
hopefully we can add reasonable 'I' values for the current
built-in virtual power trainers, but they can't be perfect for
wheel-on trainers since user's wheel's 'I' can vary.

Part 2: Use average power instead of point power.

Moved numerical integration into its own header, its now used by
bicyclesim and by virtual power. Changed default integrator to
Kahan-Li since in my testing it is stable and converges fastest.

Virtual Power now computes average power since previous sample,
which has the effect of reducing reported power during acceleration
and deceleration.

New advanced virtual power is on with this commit. It seems to work
very well in my testing.

Adds field to device to hold trainer's inertial moment
in (KG M^2).

This change allows power calculation to track energy
that passes in and out of trainer's flywheel so that
acceleration and deceleration power can be accounted
for correctly.

Whenever device sets a new rpm sample a time is
recorded alongside which allows the common controller
to compute change in rpm over time, which allows
calculation of kinetic energy that has entered or
left the trainers flywheels during the duration.

It is a bit of math to determine a correct I for a trainer's
flywheels, especially if the flywheels have different rpms,
hopefully we can add the I values for the current built-in
virtual power trainers so everyone will experience the benefit
automatically.


ericchristoffersen:
#3650

ericchristoffersen:
merge conflict - verified works great.

PR #3654, opened 2020-11-03
Add ANT+/BTLE serial number and antid into the device info.

cleeland:
Also added a poor-man's abstraction similar to `std::optional<>` that avoids messy `if` clauses that have to intuit whether something is set based on the values.  If you'd prefer using `std::optional<>` I can rework.

Work originally motivated by https://groups.google.com/g/golden-cheetah-users/c/LWhXNiLuS3k/m/3AChbB0-EAAJ

### DIFF
--- a/src/FileIO/FitRideFile.cpp
+++ b/src/FileIO/FitRideFile.cpp
@@ -1202,13 +1202,34 @@ struct FitFileReaderState
         session_data_info_.clear();
     }
 
+  template<typename T>
+  struct opt {
+    bool m_has_value;
+    T    m_value;
+    operator bool() { return m_has_value; }
+    opt() : m_has_value(false) {}
+    opt(const T& v) : m_has_value(true), m_value(v) {}
+    opt(const opt<T>&) = default;
+    opt<T>& operator=(const T& v) { m_has_value = true; m_value = v; return *this; }
+
+    T value() const {
+      if (!m_has_value)
+        throw std::exception();
+      return m_value;
+    }
+    operator T() { return value(); }
+    
+
+  };
+
     void decodeDeviceInfo(const FitDefinition &def, int,
                           const std::vector<FitValue>& values) {
         int i = 0;
 
         int index=-1;
         int manu = -1, prod = -1, version = -1, type = -1;
-        quint32 serial = 0;
+        opt<quint32> serial;
+        opt<quint32> antid = 0;
         quint8 battery_status = 0;
         fit_string_value name;
 
@@ -1231,6 +1252,9 @@ struct FitFileReaderState
                      break;
                 case 3:   // serial number
                      serial = value.v;
+                     break;
+                case 21: // ant/btle id
+                     antid = value.v;
                      break;
                 case 4:   // product
                      prod = value.v;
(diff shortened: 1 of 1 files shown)


amtriathlon:
Regrettably the highest possible value is also used to indicate serial number is not available for some sensors, that’s the reason for the test in https://github.com/GoldenCheetah/GoldenCheetah/commit/8c5b416b56c34a0cedb1481825156800e18a6f2d, removing that will include lot of invalid serial numbers, I would suggest to do some tests with the fit files in tests folder.

cleeland:
Are there unit tests for FitRideFile, or do I test simply by building the whole app and importing a fit file?

The code in 8c5b416 appears to just check that the value is within (0, max) of uint32, basically excluding 0 or max as values.  This implies that there are devices that bother to add bogus values to the DeviceInfo record?

amtriathlon:
> Are there unit tests for FitRideFile, or do I test simply by building the whole app and importing a fit file?

No, yes.

> The code in [8c5b416](https://github.com/GoldenCheetah/GoldenCheetah/commit/8c5b416b56c34a0cedb1481825156800e18a6f2d) appears to just check that the value is within (0, max) of uint32, basically excluding 0 or max as values. This implies that there are devices that bother to add bogus values to the DeviceInfo record?

That's what I found testing files, I don't know the reason.

Besides that, which would be the motivation for the change in the way serial number is handled, just to refactor the code?

cleeland:
> which would be the motivation for the change in the way serial number is handled, just to refactor the code?

When I originally made this change, `serial` was not a uint32, but an int, with a default value of -1 as an overloaded value to indicate "there is no serial value".  That seemed clumsy and error-prone, so I replaced it with an abstraction that encapsulates what we really want to know which is "did this ever get set?".  That abstraction could be extended to contain a range of valid values, or (better) a lambda for value validation.

Any change you remember which `.fit` file had the bogus values?

amtriathlon:
That changed was motivated by this thread since ID was being filtered: https://groups.google.com/d/msg/golden-cheetah-users/2Bm8NgBGtlQ/CCAKWnAtDAAJ

cleeland:
Thanks for the link to that thread.  I don't see the file that's attached to that thread anywhere in the `test/rides` directory, so I will add it.

My change was motivated by this thread: https://groups.google.com/g/golden-cheetah-users/c/LWhXNiLuS3k/m/3AChbB0-EAAJ

amtriathlon:
> My change was motivated by this thread: https://groups.google.com/g/golden-cheetah-users/c/LWhXNiLuS3k/m/3AChbB0-EAAJ

So you were referring to the previous version, that request is fulfilled in current master.

cleeland:
I'm not sure why you closed this, because the two changes, while similar, achieve different things.  There are two different fields that can be in a FIT file, and they can both appear.  One is the serial number, while the other is the ant ID.  They can be different values, too.

The change in this PR properly adds the ANT ID when that field appears in the device info record.

amtriathlon:
Let me be more explicit, this PR has one commit doing 3 separate things:
1) It supposedly fixes a bug the way serial is handled without proving it is one with a sample activity
2) It adds a feature (import antid) without providing a compelling use case
3) it refactors the code in a non-separable way, violating https://github.com/GoldenCheetah/GoldenCheetah/wiki/Guidelines-for-submitting-a-patch

cleeland:
Thanks for the clarification on the feedback related to this.  I will address all of them.

Issue #3656, opened 2020-11-08
Enable reordering of defined filters ("namedsearches")


ward:
I am able to reorder them by editing the xml file in `config/namedsearches.xml`, but that is not very user friendly. An up/down button in the "Manage filters" dialog box could be sufficient, dragging them up or down would also work.

ward:
I created a PR that adds an up/down button https://github.com/GoldenCheetah/GoldenCheetah/pull/3657

amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/e05cf52446309c01656e129d063218c933d2b5dd

PR #3657, opened 2020-11-08
Add up and down button to NamedSearch dialog

ward:
Also know as the Manage Filters dialog. Reordering was not possible in
the interface yet, you were forced to edit the xml file to do any
reordering.

Relevant issue: https://github.com/GoldenCheetah/GoldenCheetah/issues/3656

I used the updateClicked function as a "template". Let me know if I messed anything up.


amtriathlon:
Hi ward, I merged your PR moving the new buttons to bottom left, which is more consistent with the way they are used in other parts of GC. Welcome and thanks for your contribution!

Issue #2132, opened 2016-10-10
Ordering of segments


nigelvdv:
I would like to request following feature. At the moment it's possible to order user intervals. It would be nice if the detected segments could be ordered by time as well. Untill now they seem to be ordered by the date the segments have been first created.

Kind regards,

Nigel

mtbkeith:
Please describe the two pages; the one in which you can order intervals and the screen in which you want to order segments? A screenshot can also be really helpful to those who may be looking for something to investigate and may not know the code inside/out (seeing & then searching for text in the codebase can clue a developer in!)

nigelvdv:
Hi,

I attached 2 screenshots. The first one is the first climb I did during that activity after approximately 1h riding. The other one is the last climb I did during that activity after 5h. 

I created segments for both stretches of road during a previous activity. As you can see the segments are detected but not displayed in the correct order. The first one (Betsberg - Gontrode) should be displayed as the first of all segments, while the other one (Valkenberg - Brakel) should be last. At the moment I think they are ordered chronologically based on date created.

![first segment betsberg gontrode after 1h](https://cloud.githubusercontent.com/assets/14836910/24576931/e7e40a06-16c4-11e7-83a0-1b7dfe47e8ea.png)


![last segment valkenberg brakel after 5h](https://cloud.githubusercontent.com/assets/14836910/24576932/e7e573e6-16c4-11e7-8c4f-6e4e435641ac.png)


Kind regards,

Nigel

PR #3661, opened 2020-11-10
Sort routes when they are added to an activity

ward:
Before this change, when looking at an activity, the routes seemed to
appear in order of creation in the interval section. This patch instead
sorts them by their start time in the activity (or by their end
time if the start times are the same).

Note that the routes are also known as segments.

This fixes bug #2132.

### DIFF
--- a/src/Core/IntervalItem.h
+++ b/src/Core/IntervalItem.h
@@ -89,8 +89,8 @@ class IntervalItem
         RideFileInterval *rideInterval;
 
         // used by qSort()
-        bool operator< (IntervalItem right) const {
-            return (start < right.start);
+        bool operator< (const IntervalItem& right) const {
+            return std::tie(start, stop) < std::tie(right.start, right.stop);
         }
 };
 
--- a/src/Core/RideItem.cpp
+++ b/src/Core/RideItem.cpp
@@ -1447,6 +1447,15 @@ RideItem::updateIntervals()
         QList<IntervalItem*> here;
         context->athlete->routes->search(this, f, here);
 
+        // Sort routes so they are added by start time to the activity.
+        std::sort(
+            here.begin(),
+            here.end(),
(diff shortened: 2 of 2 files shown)


amtriathlon:
Hi Ward, IntervalItem class includes a < operator for sorting purposes you could use.

ward:
I had seen the operator, but I couldn't have `std::sort` rely on it directly since the `QList` consists of pointers. Now that you mention it though, I realise I should have at least made use of the operator in the function I defined.


What is the C++ target of GoldenCheetah btw? If it is more recent than C++11 I guess I could also directly use a lambda as an argument to sort.

ward:
I have updated it to use a lambda in the `std::sort` which in turn defers the comparison to `IntervalItem`'s `operator<`.

amtriathlon:
I think we could change the signature of the < operator to
`bool operator< (const IntervalItem& right) const`
to be able to use the simpler:
`std::sort(here.begin(), here.end());`
Which would be the pattern to replace deprecated qsort from qt5.2 onwards, BTW.

ward:
I have given the signature change a try, but `std::sort` seems to still compare the pointers instead. Thus I have kept the lambda in. I'll admit my C++ is a bit rusty, so maybe I'm overlooking something obvious, but searching the internet did not seem to hint at a way to make it work.

I also changed the `operator<` to use `std::tie` instead of the messier combination of comparisons I had used before.

ward:
> Build execution time has reached the maximum allowed time for your plan (60 minutes).

Is that something I did wrong or just a hiccup in AppVeyor? I do not see any error messages.

amtriathlon:
Sorry, I missed it is a list of pointers, your solution is fine, I will merge it.

AppVeyor failed due to 1h quota, it terminated Ok after restart.

Issue #3667, opened 2020-11-16
Mine Route, Segment, String Data from TTS Files


ericchristoffersen:
TTS Files have rich route name, route description, and may have richly described segments. Currently all ignored by the tts reader. We should process these and bubble them up so available to ergfile parser.

ericchristoffersen:
Currently ergfile have a route name and route description slot, so they can be populated immediatelly. Not sure how to bubble segments up into train mode. Appears the segments are used in tacx software to easily select pieces of very long files (>200km.)

amtriathlon:
Laps in ErgFile could be a good mapping for segments, name is displayed on bottom bar when entering a lap and there is logic for moving to next lap & previous lap

ericchristoffersen:
> Laps in ErgFile could be a good mapping for segments, there is logic for moving to next lap (FfwdLap). Texts are useful to show information while training, they are being used to show lap names from json files.

Woo. Good idea I'll give it a shot.

ericchristoffersen:
Ok, pushed new commit. I populate ergfile texts and laps. This led to fixing some problems with how laps were handled:
- only update lap time stuff when in ergo mode
- recalc lap metrics during fwd and rew.

As far as I can tell laps work ok now in slope mode.

The display of lap names is awful though. Needs a serious redo because they names are printed over each other. The tts segments can be fully or partially overlapped. Maybe a project-type view with bars for each segment?

![routenameproblem](https://user-images.githubusercontent.com/46055145/99472579-c745a800-28fd-11eb-8c9c-38cb73747f7e.JPG)

amtriathlon:
I added a button to Train bottom for FFwdLap method, it may be useful to play with this feature, also added lap markers from json files which makes easy to mark points where you may want to fast forward, s.t. climb starts.

ericchristoffersen:
Cool! Don't forget a "rwndLp" button too. No doubt folks will wish to replay their favorites. Hey in a different thread someone suggested we autogenerate lap markers for climbs. I think thats such a good idea. We could follow the uci categorization rules. What do you think?

ZajtiM:
@ericchristoffersen  are segments in TSS considered as laps? because i have few TSS files in which i can't pass certain segments with Next lap button. I can attach them later if you want.

amtriathlon:
> Cool! Don't forget a "rwndLp" button too. No doubt folks will wish to replay their favorites.

[...]

PR #3668, opened 2020-11-16
Teach ttsreader to process routes and segments

ericchristoffersen:
#3667 

TTS reader previously ignored routes, segments and strings.
With this change the route name, route description, segments,
segment descriptions are now all parsed into ttsreader object.

With this change the route name and description are assigned to
the ergfile, so route name now appears while riding a tts file.

With this change there is still no place to put segments in ergfile.

With this change none of the new information is assigned into the
activity file.

The segments and segment descriptions are used by tacx software
so user can select a named region of a long ride. Example ride
was >200km long and contained 12 named segments, each with a nice
description. Would be nice to bubble that info up to train mode.

### DIFF
--- a/src/FileIO/TTSReader.cpp
+++ b/src/FileIO/TTSReader.cpp
@@ -61,6 +61,21 @@ const log_disabled_output& operator << (const log_disabled_output& any, const st
 
 using namespace NS_TTSReader;
 
+// This is the static const string table that tts files may refer to by string key.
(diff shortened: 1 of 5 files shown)


ericchristoffersen:
Hold on: Found more stuff broken about laps...

ericchristoffersen:
Hold on found more stuff broken about laps...

PR #3671, opened 2020-11-18
Populate Details > Route field during Import - new feature.

### LINKED ISSUE #3630: Workout name as Details > Route
It would be nice to have Workout name automatically saved as Route in Details tab in activities.

Example:
My workout file names (for gpx and TTS) are route names. 
And I also use Route (in Details tab) as my Strava Activity name. So when you save your workout it could add filename / workout name as Route in Details tab.

I think Route is most logical for gpx and TTS files.

ZajtiM:
New feature which can populate Details > Route field. It works with
both "Import from file" and "Train view" workouts. Can be used when
importing multiple files with "Import from file". Details > Route
field will be populated with "Train view" workout filename or with
filename which will be imported with "Import from file".
Fixes #3630

Some extra info is posted here:
https://github.com/GoldenCheetah/GoldenCheetah/issues/3630#issuecomment-729945568

### DIFF
--- a/src/Gui/RideImportWizard.cpp
+++ b/src/Gui/RideImportWizard.cpp
@@ -16,6 +16,7 @@
  * Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
 
+#include "ErgFilePlot.h"
 #include "Context.h"
 #include "Athlete.h"
 #include "MainWindow.h"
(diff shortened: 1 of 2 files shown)


ericchristoffersen:
If this is fresh in your head it might be easy. We just recently added route name and description from tts files, also laps and lap names. Right now the text data is all available in train mode but not moved into activity when train mode ends. Attached a tts with beaucoup text descriptoins if you're interested:

[F_GrandAlps2011.zip](https://github.com/GoldenCheetah/GoldenCheetah/files/5563618/F_GrandAlps2011.zip)

ZajtiM:
@ericchristoffersen  i noticed it in my build yesterday, when doing my tests. Great feature. I'm very new to c++, so i can't promise you anything. But i will try it fur sure.

Any suggestions where and in what "format" should be added? I assume Notes would be good place to add them.

ericchristoffersen:
Where should the data go? Notes seems natural to me. User is free to overwrite it if they want.

Issue #3672, opened 2020-11-19
Crash Reading TTS File


ericchristoffersen:
Good to fix: there are a few cases where we read out of range when processing tts files that have no points.

amtriathlon:
Fixed by #3672

PR #3673, opened 2020-11-19
Fix3672 - Read out of range when tts has 0 points

ericchristoffersen:
#3672 Found it processing some new tts files that I found.

### DIFF
--- a/src/FileIO/TTSReader.cpp
+++ b/src/FileIO/TTSReader.cpp
@@ -462,6 +462,9 @@ bool TTSReader::deriveMinMaxSlopes(double &minSlope, double &maxSlope, double &v
 
 void TTSReader::recomputeAltitudeFromGradient() {
 
+    if (points.size() < 3)
+        return;
+
     // Some tts files I see have an altitude crisis around the start: Fixup altitude
     // of index 0 and 1 by linear interpolating.
     double d2d1delta = points[2].getDistanceFromStart() - points[1].getDistanceFromStart();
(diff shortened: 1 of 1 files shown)


amtriathlon:
Hi Erik, when compiling this I am seeing several warnings:
```
FileIO/TTSReader.cpp: In function ‘float AsFloat(unsigned int)’:
FileIO/TTSReader.cpp:85:38: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     return reinterpret_cast<float&>(u);
                                      ^
FileIO/TTSReader.cpp: In function ‘float AsFloat(int)’:
FileIO/TTSReader.cpp:94:38: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     return reinterpret_cast<float&>(i);
                                      ^
FileIO/TTSReader.cpp: In member function ‘bool NS_TTSReader::TTSReader::loadHeaders()’:
FileIO/TTSReader.cpp:693:20: warning: enumeration value ‘NONPRINTABLE’ not handled in switch [-Wswitch]
             switch (stringType) {
                    ^
FileIO/TTSReader.cpp:613:9: warning: variable ‘fingerprint’ set but not used [-Wunused-but-set-variable]
     int fingerprint = 0;
         ^

```

I think we should try to fix this when code is added or modified, even when they may be inocuos keeping them makes increasingly difficult to spot problematic ones and build logs hard to navigate.

PR #3681, opened 2020-11-26
Fix 3680: Text, Chim and Lap Issues

ericchristoffersen:
Several Minor Fixes for Texts, Laps and Audio Chimes.

- Some files have multiple 'texts' to be emitted at the same location. Currently we emit only the last.
- TTS Reader texts and lap names should use tr for common prefixes like start and end.
- Audio Chime and text emit limit should be reset on ffwd, rewind, ffwdlap, rewindlap and newlap.
- Add LapRangeNum to lap so lapmarkers can be grouped, for example into start/end ranges. This will be useful to connect lap boundaries so we can display lap ranges in chart.
- newLap method doesnt return on all paths. Fix so returns lapnum of new lap.

Also:
- Fix type punning error in ttsreader.

### DIFF
--- a/src/FileIO/TTSReader.cpp
+++ b/src/FileIO/TTSReader.cpp
@@ -82,8 +82,6 @@ static const std::map<int, const char*> tts_string_table = {
 
  // Handy utils
 float AsFloat(unsigned u) {
-    return reinterpret_cast<float&>(u);
-
     float t;
     static_assert(sizeof(u) == sizeof(t), "Error: mismatched reinterpret sizes");
     memcpy(&t, &u, sizeof(u));
@@ -91,8 +89,6 @@ float AsFloat(unsigned u) {
 }
 
 float AsFloat(int i) {
-    return reinterpret_cast<float&>(i);
-
     float t;
     static_assert(sizeof(i) == sizeof(t), "Error: mismatched reinterpret sizes");
     memcpy(&t, &i, sizeof(i));
--- a/src/Train/ErgFile.cpp
+++ b/src/Train/ErgFile.cpp
@@ -771,6 +771,10 @@ void ErgFile::parseFromRideFileFactory()
 
         prevPoint = point;
         point = nextPoint;
+
+        if (!point)
+            break;
+
         nextPoint = ((i + 1) < pointCount) ? (ride->dataPoints()[i + 1]) : NULL;
 
         // Determine slope to next point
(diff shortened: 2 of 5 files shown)


amtriathlon:
Hi Eric, I don't understand what is the role of LapRangeNum no why it is part of a PR intended to fix cosmetic errors.

ericchristoffersen:
> Hi Eric, I don't understand what is the role of LapRangeNum no why it is part of a PR intended to fix mostly cosmetic errors.

Ok. Tell me what you want here. Lots of different valid approaches. I was thinking its a priority that I don't muck with 'lap' navigation, but I also want a way for the display to look good. The groupId allows later lap consumers (aka, ergfileplot) to reconstruct the way in which the laps were related, specifically to handle the segment spans that we get from tts.

Idea is that the overview plot could rebuild the segments, then nicely arrange everything so the segments didn't overlap. All it takes is that little breadcrumb.

I implemented all of that on top of this pull request, but realized its a bit of weird and ugly work to reconstruct the laps, and would leave hanging what to do if a group had more than 2 members. So decided it is probably more readable and understandable to just add a new end field on the laps. So I redid it using an end field. The printing is a bit hokey but I think it is sufficient until some point in the future where we decide to rewrite the graphs in html.

The arrangement is sort of complicated because the laps have arbitrary partial overlap, and must mix with random non-spanning lap markers. Here is my current state showing a complex case:

image

It seems to be working pretty well. I'll try and push an update today.

amtriathlon:
> Tell me what you want here.

Currently we have lap markers and text cues, which are supported in most of the workout formats we use in both modes, they are simple concepts mapping well to current UI. If this PR is about to fix issues related to this features, please keep the scope restricted to that.

I am not convinced we should introduce the notion of "segment" just because one of the formats we can import has it without evaluate the impact on all other formats, the user interface and the added complexity, but in any case, it should be a separate PR.

ericchristoffersen:
> > Tell me what you want here.
> 
> Currently we have lap markers and text cues, which are supported in most of the workout formats we use in both modes, they are simple concepts mapping well to current UI. If this PR is about to fix issues related to this features, please keep the scope restricted to that.
> 
> I am not convinced we should introduce the notion of "segment" just because one of the formats we can import has it without evaluate the impact on all other formats, the user interface and the added complexity and the how useful are the new functions it enables. Just to be able to end the workout at segment end doesn't seem to merit this, IMHO.
> In any case, it should be a separate PR for the same reasons we talked in other cases.

Sure thing I'll pull the segment stuff out.

Issue #3685, opened 2020-11-29
Segment names are Unreadable in ErgFileOverview


ericchristoffersen:
Currently segment names stomp each other in overview.

![garblegarble](https://user-images.githubusercontent.com/46055145/100556665-aca5f400-3258-11eb-9683-4e129cfab103.JPG)

Mostly I care because 95% of my virtual rides have a bunch of named segments.Two parts to fix:

First is to add an 'end' field to the ergfilelap. For non-segment laps this has no meaning and is kept as a negative number (-1). This allows overview to know a segment's start and end.

Second part of fix is to distribute segments to rows in overview so they dont stomp each other.

![distributedsegments](https://user-images.githubusercontent.com/46055145/100556727-2d64f000-3259-11eb-918f-1d528f54201e.JPG)

Hopefully someone in the future will feel inspired to replace the ascii art in overview plot with something colored and graphical that takes into account the size of printed text. Knowing text size would allow named laps to be distributed to rows and so also be printed without their names stomping over each other.

PR #3686, opened 2020-11-29
Fix 3685: Readable Segments in ErgFileOverview

ericchristoffersen:
This change was built on top of fix for 3680.

For details see description and screenshots attached to issue 3685.

### DIFF
--- a/src/FileIO/TTSReader.cpp
+++ b/src/FileIO/TTSReader.cpp
@@ -324,25 +324,31 @@ void TTSReader::segmentRange(int version, ByteArray &data) {
 void TTSReader::segmentInfo(int version, ByteArray &data) {
 
     if ((version == 1104) && (data.size() == 8)) {
-        double startKM = (getUInt(data, 0) / 100000.0);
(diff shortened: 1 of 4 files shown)


ericchristoffersen:
#3685

ericchristoffersen:
I regret my change from rangeId to adding 'end' to ergfilelap. Is too much noise. I think the rangeId breadcrumb is a much better and easier approach for addressing the display issues. I'm going to redo this with rangeid.

ericchristoffersen:
Ok. It is more work for ErgFilePlot but I think much cleaner for everyone else. A lap with rangeId of 0 means it is not in a segment, is printed like laps are today. Non-zero rangeid indicates it is part of a segment group. The earliest in the group is the start, the lastest is the end. Moved the row distribution into its own class. Result is no different from previous changes. Non-segment laps still work great.

ericchristoffersen:
Rats. Can you take a look at #3688 ? Just minor code cleanups.

PR #3691, opened 2020-12-01
Daum: Adapt code for 8008 TRS pro

chrta:
For this a 50 msec delay is introduced before sending each serial command. This is tested on hardware and also used by jergotrainer.

Also during initialization there is a delay between each command, so even slow devices can keep up. As soon as the cockpit type is known, the delay is set according to the cockpit type.

The behavior for all other cockpit types should be unchanged.

### DIFF
--- a/src/Train/Daum.cpp
+++ b/src/Train/Daum.cpp
@@ -25,6 +25,8 @@ Daum::Daum(QObject *parent, QString device, QString profile) : QThread(parent),
     serial_dev_(0),
     deviceAddress_(-1),
     maxDeviceLoad_(800),
+    serialWriteDelay_(0),
+    playSound_(false),
     paused_(false),
     devicePower_(0),
     deviceHeartRate_(0),
@@ -180,6 +182,9 @@ void Daum::run() {
 }
 
 void Daum::initializeConnection() {
+
+    serialWriteDelay_ = 0;
+
     char addr = (char)GetAddress();
     {
         QMutexLocker locker(&pvars);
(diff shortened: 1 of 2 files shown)


chrta:
I have two questions:

1. I would expect that other similar cockpits do not work properly because of the missing delay. But i only added the delay of 50 msec to the ergometer i could test to not break anything. Should i leave it as it is now?
2. There is some redundant code in Daum.h and Daum.cpp (redundant member parent, redundant unused virtual keywords, redundant casts, redundant `this->`) Should i open a separate pr to fix this?

amtriathlon:
> 1. I would expect that other similar cockpits do not work properly because of the missing delay. But i only added the delay of 50 msec to the ergometer i could test to not break anything. Should i leave it as it is now?

A quick search shows no activity in the forums or here related to Daum devices so, they work perfectly or nobody but the original developer has used this part of the code. In the original commit message there is an email and he has contributed this year, perhaps you can contact him, otherwise better to restrict the changes to what you can test.

> 2. There is some redundant code in Daum.h and Daum.cpp (redundant member parent, redundant unused virtual keywords, redundant casts, redundant `this->`) Should i open a separate pr to fix this?

Yes, please.

thebaron06:
As I recently sold my Daum device, I am not able to test this. But I was using the code as is and it was working fine with my device.
But IIRC @afriess was using this with a similar device. Maybe he is willing to help here. I think he was also working on adding more device features/modes.

Regarding the timeouts I can only guess that this wouldn't have a negative effect on the _Vital_ device I was using.
Please feel free to clean up the code, I would appreciate your help.

Hint: Another possibility to enable users to slow down communication - maybe other serial adapters need this - would be to add an additional parameter to the profile data that could be set via Settings-UI as it was done with the _force_ flag, see [here](https://github.com/GoldenCheetah/GoldenCheetah/blob/90f0e419ab3f98f1be36e25bd1dd8a5b6bfefae9/src/Train/Daum.cpp#L37). This could also be applied to the initial communication timeouts.

amtriathlon:
I will merge then, thanks!

jotrax:
Using build 4003 from Jan 5 2021 I just tried to connect to my Daum TRS8008 under Ubuntu 18.04. I used the "Add device wizard", selected Daum Cockpit and it was found by the scan on the correct tty port. Shortly after connecting, it gave me an error msg: "Cannot Connect to Daum"

Can you confirm, that build 4003 has the 8008 support included?

amtriathlon:
Yes.

chrta:
@jotrax I will check today if the version works with my Daum device. I suspect, that your cockpit reports another id and therefore it does not add the required delay to the commands. If you can somehow activate the qt logger and set it to debug loglevel you should see a log line, like `Daum cockpit type: 0x40`. If it is 0x40, the delay is missing. If it is 0x64 (as mine) there is something else wrong.

I will check if i can make the delay configurable independent from the cockpit type (probably this week or the next week).

chrta:
@jotrax Maybe you can open an issue and ping me with the log information. I will have a look.

afriess:
I have done some stabilzation for daum devices here https://github.com/afriess/GoldenCheetah/tree/DAUM_Vita2002/src/Train 2 years ago. I have built a Daum Simulation https://github.com/afriess/ergobike here, i used this to test all commands (documented or not by daum). The reference programm was the orginal ErgoPlanet. The simulator can simulate a normale ride in ErgoPlanet, without any problem. So for tests it is possible to use a PC or RasPi to simulate a DaumBike. The Programm is written in Freepascal/Lazarus (NOT Delphi) and tested on Win10 PC and Rasbian.

Issue #3698, opened 2020-12-07
Simulate Hypoxia in Train Mode


ericchristoffersen:
Have you ever thought that it is no fun at all to make unrealistic virtual progress up a pass? Well if this has been bothering you I have a new feature you might like: Hypoxia Simulation.

There is a known decrease in aerobic capacity at altitude so it is no fair at all for us to treat your Puny Sea Level Watts as if they were Top of A High Pass Watts. Idea is to use a hypoxia factor to change the power you experience in your simulated ride.

For the hypoxia factor I am using the data from a paper:

    Prediction of Critical Power and W? in Hypoxia : Application to Work - Balance Modelling
    by Nathan E.Townsend1, David S. Nichols, Philip F. Skiba, Sebastien Racinais and Julien D. Périard

This is a description of the change in CP with altitude. Change to W' is not well understood today so not addressed. This approach does not take zones into account it basically whacks off a ratio of your power, including from your W' which arguably it should not.

Here's how it works:

simAltitude            : current altitude in your simulated ride
trainerAltitude        : a field in settings for you to enter the actual altitude of your indoor trainer
HypoxiaFactor(altitude): a function that maps altitude to hypoxia cp factor.

The watt adjustment is then:

    HypoxiaFactor(simAltitude) / HypoxiaFactor(trainerAltitude)

Some cases:

Trainer in Leh (altitude 3500m) and simulating a ride around the streets of Leh, then the factor is 1.
Trainer in Denver (altitude 1600m), and ride across Shingo-La (altitude 5000m) you'll be going real slow.
Trainer in Leh (altitude 3500m), simulated ride in Holland (altitude 0m), then you will receive a mighty speed boost.

This makes me happy because it is true that 200 leh-watts are much more painful than 200 sea level watts.

A number: If your trainer is at sea level then near the top of simulated alp d'huez (1800m) you effectively lose about 9% of your power.

For now I have hidden the adjustment inside the bicycle simulation. I don't think anyone really wants to see their simulated hypoxic watts, but I can add a field to rtdata if people would find it interesting. All this change effectively does is alter the power applied to the simulation meaning the time that is taken to transit a virtual ride will better reflect reality.

People are often surprised how the altitude effects their progress, so this feature might be useful for planning long higher-altitude rides, like from Manali to Leh.

PR #3699, opened 2020-12-07
Simulated Hypoxia for Train Mode

ericchristoffersen:
#3698 

This adds an option to trainer preferences to enable hypoxia sim. If its enabled power given to simulation will be adjusted by altitude.

There is a new field in bicycle sim to hold training altitude which is also fed into the altitude calc, so that people training at high altitude while doing virtual ride at high altitude will see no penalty.

Here is graph of Townsend's equation based on data from 0-4250m, and the extended equation I'm using.

image


amtriathlon:
I Eric, there are lots of warnings from Train/MultiRegressionizer.h and PolyFit, can we fix them?

The other contributor to log pollution is InRide change, but that is not related to this issue and I should have checked that before merging.

ericchristoffersen:
> I Eric, there are lots of warnings from Train/MultiRegressionizer.h and PolyFit, can we fix them?

Sure I'll take a look. I think I looked in the past and the warnings look like tool errors to me, overeager messages. I'll see.

ericchristoffersen:
> > I Eric, there are lots of warnings from Train/MultiRegressionizer.h and PolyFit, can we fix them?
> 
> Sure I'll take a look. I think I looked in the past and the warnings look like tool errors to me, overeager messages. I'll see.

Ok, I've created #3705 to address some of the build warnings. I can't test those other buils so will iterate on it until it is clean. 

Is not related to this change.

Issue #3712, opened 2020-12-13
Honor Simulated Speed Setting in Ergo Mode


ericchristoffersen:
Ergo mode is a smart trainer mode where trainer requires N watts to move, without regard to speed or cadence. Currently ergo mode in gc reports speed according to whatever trainer reports and this often just spends on what gear rider chooses for their workout. This gear choice can have the effect of significantly skewing athlete statistics.

Suggestion from Laurent on gc user group to run bicycle simulation in ergo mode so that erg mode workouts will have consistent speed and distance.

https://groups.google.com/g/golden-cheetah-users/c/o3_0vHeGc1c/m/n0490j0aCAAJ

amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/603c8cd651db5d3032a7ce137c9eb5f6c96b6595

PR #3713, opened 2020-12-13
Fix 3712: Simulated Speed setting now honored in Ergo Mode.

ericchristoffersen:
#3712

### DIFF
--- a/src/Gui/Pages.cpp
+++ b/src/Gui/Pages.cpp
@@ -735,11 +735,17 @@ bool deviceModel::setData(const QModelIndex &index, const QVariant &value, int r
 //
 TrainOptionsPage::TrainOptionsPage(QWidget *parent, Context *context) : QWidget(parent), context(context)
 {
-    useSimulatedSpeed = new QCheckBox(tr("Use simulated Speed in slope mode"), this);
+    useSimulatedSpeed = new QCheckBox(tr("Simulate Speed From Power"), this);
     useSimulatedSpeed->setChecked(appsettings->value(this, TRAIN_USESIMULATEDSPEED, false).toBool());
+    useSimulatedSpeed->setToolTip(tr("Simulation physics uses current athlete parameters and settings\n"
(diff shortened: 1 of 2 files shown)


amtriathlon:
Settings label should be changed accordingly: https://github.com/GoldenCheetah/GoldenCheetah/blob/983874d59963c05c06e8c4a4a2df69aad3a381da/src/Gui/Pages.cpp#L738

caronl:
Pretty clean solution @ericchristoffersen! Haven't tried the feature yet but I am imagining the fix would do exactly what I had in mind. As long as the label is adjusted accordingly (something like "Use simulated Speed in erg and slope modes") I think that solves it. Thanks! Can't wait to try the next build :)

amtriathlon:
There is another place which also refers to "slope mode": https://github.com/GoldenCheetah/GoldenCheetah/blob/af19f31b245014881e46d13454c424c9449c9624/src/Gui/Pages.cpp#L1075
From that text only the first line is displayed on Windows, but on Linux it looks crammed: 
VirtualBicycleSpecifications
This would need some reformat or to move the additional text to a tooltip

ericchristoffersen:
Thank you for checking Ale. Good catch. I messed with layout a bit, changed text, etc. Here is with latest. Now I'm sort of thinking to swap the left and right columns.

image

ericchristoffersen:
Noooo... I didn't update the pull yet...

I guess I resync, make new branch, new pull, etc?

I think its better with the swapped columns:

image

ericchristoffersen:
Ok, the latest commit has the last layout I posted above - moved the mass column to the right and fixed typos in some tooltips.

amtriathlon:
Yes, I thought it was ready from your picture.
It doesn't scale too well when the window is not tall enough, which is weird because it happens even when the column has free space at the bottom. If you want to update, please make a new PR.

ericchristoffersen:
Ok, made #3714 with the change to pages.cpp.

caronl:
Feature works exactly as imagined in the first place. Love it! Kudos @ericchristoffersen

PR #3725, opened 2020-12-18
Fortius Working In Golden Cheetah - Based on the FortAnt Technology

ericchristoffersen:
This is an experimental change. Adds resistanceWatts hint to controller setGradient. This is a hint for what watts level to use for training devices that wish to use ergo mode instead of slope mode.

This change plumbs the use of resistanceWatts down into fortius, whose current slope mode doesn't work well at all.

https://groups.google.com/g/golden-cheetah-users/c/8EDSFJiRLGU

PR #3731, opened 2020-12-22
TACX Fortius - new formulae and calibration routine

mattipee:
- Fortius device (ie USBheadunit+motorunit+motor)
  - does not implement slope or ergo mode itself
  - interface is "resistance" mode only
  - has a configurable "virtual flywheel"

- Fortius driver (Train/FortiusController.cpp and Train/Fortius.cpp)
  - this PR now implements both slope and ergo (corrected the previous, incorrect formulae)
  - device's virtual flywheel is employed by the driver's implementation of slope mode.
  - presents to GC a controllable trainer which operates in a closed loop.

### DIFF
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -17,7 +17,6 @@
  */
 
 #include "Fortius.h"
-
 //
 // Outbound control message has the format:
 // Byte          Value / Meaning
@@ -34,7 +33,7 @@
 // 10            Calibration Value - Lo Byte
 // 11            Calibration High - Hi Byte
 
-// Encoded Calibration is 130 x Calibration Value + 1040 so calibration of zero gives 0x0410
+// Encoded Calibration is 130 x Calibration Adjust + 1040 so calibration of zero gives 0x0410
 
 const static uint8_t ergo_command[12] = {
      // 0     1     2     3     4     5     6     7    8      9     10    11
@@ -46,21 +45,43 @@ const static uint8_t slope_command[12] = {
         0x01, 0x08, 0x01, 0x00, 0x6c, 0x01, 0x00, 0x00, 0x02, 0x48, 0x10, 0x04
 };
 
+const static uint8_t calibrate_command[12] = {
+     // 0     1     2     3     4     5     6     7    8      9     10    11
+        0x01, 0x08, 0x01, 0x00, 0xa2, 0x15, 0x00, 0x00, 0x03, 0x00, 0x00, 0x00
+};
 
+// From switchabl on FortAnt project, based on
+//  https://github.com/totalreverse/ttyT1941/wiki#newer-usb-1264-bytes-protocol
+//
+// Power resistance factor is 137 * 289.75 * 3.6 ~= 142905
+static const double s_wheelSpeedFactor = 289.75;
+static const double s_powerResistanceFactor = 137 * s_wheelSpeedFactor * 3.6;
+
+class Lock
+{
+    QMutex& mutex;
+public:
+    Lock(QMutex& m) : mutex(m) { mutex.lock(); }
+    ~Lock() { mutex.unlock(); }
+};
     
 /* ----------------------------------------------------------------------
  * CONSTRUCTOR/DESRTUCTOR
  * ---------------------------------------------------------------------- */
 Fortius::Fortius(QObject *parent) : QThread(parent)
 {
-    
-    devicePower = deviceHeartRate = deviceCadence = deviceSpeed = 0.00;
+
+    devicePower = deviceResistance = deviceHeartRate = deviceCadence = deviceSpeed = deviceWheelSpeed = 0.00;
     mode = FT_IDLE;
     load = DEFAULT_LOAD;
     gradient = DEFAULT_GRADIENT;
     weight = DEFAULT_WEIGHT;
-    brakeCalibrationFactor = DEFAULT_CALIBRATION;
+    brakeCalibrationFactor = DEFAULT_CALIBRATION_ADJUST;
+    brakeCalibrationValue = DEFAULT_CALIBRATION_VALUE;
     powerScaleFactor = DEFAULT_SCALING;
+    windSpeed_ms = DEFAULT_WIND_SPEED;
+    rollingResistance = DEFAULT_ROLLING_RESISTANCE;
+    windResistance = DEFAULT_WIND_RESISTANCE;
     deviceStatus=0;
     this->parent = parent;
 
@@ -72,6 +93,7 @@ Fortius::Fortius(QObject *parent) : QThread(parent)
      */
     memcpy(ERGO_Command, ergo_command, 12);
     memcpy(SLOPE_Command, slope_command, 12);
+    memcpy(CALIBRATE_Command, calibrate_command, 12);
 
     // for interacting over the USB port
     usb2 = new LibUsb(TYPE_FORTIUS);
@@ -87,17 +109,15 @@ Fortius::~Fortius()
  * ---------------------------------------------------------------------- */
 void Fortius::setMode(int mode)
 {
-    pvars.lock();
+    Lock lock(pvars);
     this->mode = mode;
-    pvars.unlock();
 }
 
 // Alters the relationship between brake setpoint at load.
 void Fortius::setBrakeCalibrationFactor(double brakeCalibrationFactor)
 {
-    pvars.lock();
+    Lock lock(pvars);
     this->brakeCalibrationFactor = brakeCalibrationFactor;
-    pvars.unlock();
 }
 
 // output power adjusted by this value so user can compare with hub or crank based readings
@@ -105,10 +125,9 @@ void Fortius::setPowerScaleFactor(double powerScaleFactor)
 {
     if (powerScaleFactor < 0.8) powerScaleFactor = 0.8;
     if (powerScaleFactor > 1.2) powerScaleFactor = 1.2;
-    
-    pvars.lock();
+
+    Lock lock(pvars);
     this->powerScaleFactor = powerScaleFactor;
-    pvars.unlock();
 }
 
 // User weight used by brake in slope mode
@@ -117,10 +136,37 @@ void Fortius::setWeight(double weight)
     // need to apply range as same byte used to signify erg mode
     if (weight < 50) weight = 50;
     if (weight > 120) weight = 120;
-        
-    pvars.lock();
+
+    Lock lock(pvars);
     this->weight = weight;
-    pvars.unlock();
+}
+
+void
+Fortius::setWindSpeed(double ws)
+{
+    Lock lock(pvars);
+    this->windSpeed_ms = ws;
+}
+
+void
+Fortius::setRollingResistance(double rr)
+{
+    Lock lock(pvars);
+    this->rollingResistance = rr;
+}
+
+void
+Fortius::setWindResistance(double wr)
+{
+    Lock lock(pvars);
+    this->windResistance = wr;
+}
+
+
+void Fortius::setCalibrationValue(uint16_t val)
+{
+    Lock lock(pvars);
+    this->brakeCalibrationValue = val;
 }
 
 // Load in watts when in power mode
@@ -130,9 +176,8 @@ void Fortius::setLoad(double load)
     if (load > 1000) load = 1000;
     if (load < 50) load = 50;
         
-    pvars.lock();
+    Lock lock(pvars);
     this->load = load;
-    pvars.unlock();
 }
 
 // Load as slope % when in slope mode
@@ -141,20 +186,19 @@ void Fortius::setGradient(double gradient)
     if (gradient > 20) gradient = 20;
     if (gradient < -5) gradient = -5;
     
-    pvars.lock();
+    Lock lock(pvars);
     this->gradient = gradient;
-    pvars.unlock();
 }
 
 
 /* ----------------------------------------------------------------------
  * GET
  * ---------------------------------------------------------------------- */
-void Fortius::getTelemetry(double &power, double &heartrate, double &cadence, double &speed, double &distance, int &buttons, int &steering, int &status)
+void Fortius::getTelemetry(double &power, double &resistance, double &heartrate, double &cadence, double &speed, double &distance, int &buttons, int &steering, int &status)
 {
-
-    pvars.lock();
+    Lock lock(pvars);
     power = devicePower;
+    resistance = deviceResistance;
     heartrate = deviceHeartRate;
     cadence = deviceCadence;
     speed = deviceSpeed;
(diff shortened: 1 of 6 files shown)


mattipee:
@ericchristoffersen @amtriathlon

ericchristoffersen:
Thinking out load, conversation not a suggestion:

An example of the worst implementation: I played for a bit with a kurt kinetic 'dumb' trainer. Its a turbo trainer with fixed power curve that reports watts. This is the same as having a fixed gradient. Climbing was not realistic at all because I had to simply spin at the speed needed to generate the power. The feel of climbing steeps at low speed out of saddle was ruined because the device was spinning too fast and had too much momentum so didn't slow properly between pedal strokes. My kickr runs pretty close to actual speed and feels terrific, but it can handle 500 watts at cadence of 25.

I'd call the dumb trainer a maximal divergence between sim and device feel.

Now suppose the trainer was init with a const handle to the sim's Bicycle (but not the simstate). That would provide the exact invariants and sim would be congruent. Suppose we also made altitude available to sim alongside slope. The numbers should then be the same except for inertia.

Currently there is no trainer that does this correctly. Every trainer has its own 'sim model' for slope mode, and their resistance cannot be correct because they aren't fed enough data.

Driver wants to honor device inertia, literally accept the device speed, while this will drive an increase in resistance it will almost certainly not match sim speed. Dollars to doughnuts the device will always accelerate/decelerate faster than the sim.

So even with the identical sim the speeds in the two worlds diverge, and I don't think there is any way to get them to agree short of duplicating the effective mass of bicycle and rider.

The sim generated resistance watts is absolutely the correct value for the trainer, but that doesn't allow for acceleration because Erg-Mode's job is to yield when extra power is applied. The hint out of that problem is that the device speed increases, which forces us to honor device speed and map speed to power when in slope mode.

I think correct here involves looking at graphs of 'function of devicespeed' vs displayspeed to see how quickly they can be made to converge under acceleration. I'm especially curious about cases with decreasing and increasing slope.

Note I say "function of device speed" because in correct implementation there is an offset that will be needed to support high torque and a scalar to correctly model inertia.

mattipee:
Consider this PR as "what it should have been, before introducing the sim".

The approach in this PR is consistent with FortiusANT, I imagine TTS4 as well, and alone, it should make Fortius usable.

The calculation of power required for current speed is correct and the trainer's virtual flywheel handles the acceleration/deceleration, as far as TACX went with their implementation of it, anyway. 

Thoughts on improving sim are fun and engaging, and I imagine there's an elegant mathematical solution. But I hope this PR is at least a self-consistent first step.

ericchristoffersen:
I realized somethiing on my freezing bike ride today. This fortius driver is averaging over the previous second. That might be why the mass variable isnt providing power from inertia: the history is adding latency?

src/Train/Fortius.cpp
```diff
@@ -539,35 +569,76 @@ int Fortius::sendCloseCommand()
...
-    double load = this->load;
-    double weight = this->weight;
-    double brakeCalibrationFactor = this->brakeCalibrationFactor;
-    pvars.unlock();
-    
+
+    {
+        Lock lock(pvars);
+
+        int mode = this->mode;
```
ericchristoffersen, line 587: These variables all die at end of scope...

mattipee: Good spot. Over-zealous.

mattipee:
ERG mode test
```

[COURSE HEADER]
VERSION = 2
UNITS = METRIC
DESCRIPTION = golden cheetah
FILE NAME = /home/majp/100to250to100.erg
MINUTES WATTS
[END COURSE HEADER]
[COURSE DATA]
0 100
0.5 100
0.5 150
1 150
1 200
1.5 200
1.5 250
2 250
2 250
2.5 250
2.5 200
3 200
3 150
3.5 150
3.5 100
4 100
[END COURSE DATA]
```

Bike: 56:16, 700x23, 60 rpm ~= 26.4 km/h, 90 rpm ~= 39.7 mh/k

30s each at 100W, 150W, 200W and 250W, cadence 60 rpm
30s each at 250W, 200W, 150W and 100W, cadence 90 rpm

Branch master:
image

Branch fortius_changes:
image

Analysis:
- power
  - master under-reports power at 60rpm, appears correct at 90rpm
  - fortius_changes reports correct power at 60rpm and 90rpm
- speed
  - master 60 rpm ... speed ~24 (vs 26.4 actual)
  - master 90 rpm ... speed ~36 (vs 39.7 actual)
  - fortius_changes 60 rpm ... speed ~22.5 (vs 26.4 actual)
  - fortius_changes 90 rpm ... speed ~35 (vs 39.7 actual)

Conclusion:
- power: master is bad (accuracy dependent on speed), PR is good (at both speeds tested)
- wheelscale: master is bad (under-reports speed), PR is worse (under-reports speed to a greater extent)

ericchristoffersen:
1) Speed comparison: Is that device speed vs simulated speed? I think priority should be to investigate the low rpm power issue. Does no good to match sim speed if the trainer can't hold. I have a computrainer that is infuriating because it can't hold my power at a slope of more than about 6%, I end up literally riding my rear wheel slipping for a full revolution against trainer flywheel. If I taught the computrainer to run the wheel faster than sim it would support the power I need. I imagine you will find the same issue here, which is why I don't think we should focus about sim and device speed matching. If it turns out there is no reasonable low power issue then that is great, device speed then doesn't (functionally) matter so we can endeavor to make it sim speed if that floats our boat. But lets test that before we decide that sim and device speed should match.

2) I really think you should disable the power smoothing from kinetic, at least for testing the slope mode. That latency will hide the brake response latency. In slope mode power variance is smoothed because its applied through the bicycle+rider's effective mass. For erg mode... personally I think I'd rather see the noise and smooth later, but I don't really care and I know people like to pretend power is smooth and accurate.

3) Assuming we must run a sim in the device driver, it would be better to plumb a reference to the bicyclesim down into the devicecontroller and then give it to the fortius to use for sim. This means all the simulation and constants would be the same, and we get a shared implementation. At this point altitude is missing from controller interface, so that'd be a good thing to add. With that change we will still see a speed difference because device's implemented (physical) momentum will be different.

mattipee:
@michelrdagenais I hope you can find time to follow here...


The most basic question I had is "is the speed scale" correct?

This is the constant (289.75) by which we divide the raw speed value from the Fortius to obtain speed in km/h (and therefore in deriving m/s), one of the parameters to determining power output.

The origin of this value appears to be https://github.com/totalreverse/ttyT1941/wiki:

> 'Load' is the target speed. load_speed = 'kph * 289.75'
> TTS4 calibrates with value '0x16a3' and says that this is 20 kph.
> 0x16a3 / 20 (kph) = 289.75.


I ran a test...

- Fortius motor spindle
  - 40.0 mm diameter, as best I can measure it
  - 40*pi ~= 125.6637 mm circumference.

- TACX trainer tyre
  - Standard blue TACX trainer tyre, 23-622(700x23c)
  - Single rotation results in ~16.533... turns of the motor spindle
  - calculated rolling circumference (unweighted) is 125.6637 mm * 16.533... ~= 2077.64 mm
  - there's error in that, granted...

- Constant, "known" speed
  - perhaps "assumed" speed is more accurate, given likely error in the above, and then also weighting the bike
  - pedalled with a metronome at 125 bpm, for a cadence of 62.5 rpm
  - gear ratio is 56:18 = 3.111...
  - (62.5 rpm * 60 min) * 3.111... * 2077.64 mm ~= 24.24 kph

- Speed values from the Fortius
  - 3m40s of samples
  - seemingly at 19.56 Hz
  - mean 7043.17, standard deviation of 47.03

- Calculated average speed
  - 7043.17 / 289.75 = 24.3 kph
  - an error of 0.25%


While this leaves me baffled as to the units of the raw speed values from the Fortius, I am confident that we accept 289.75 as the correct value for the speed scale.


The other value we have to accept is the constant (137) by which we divide the raw resistance value from the Fortius to obtain force in N, the other parameters to determining power output.

This is referenced again at https://github.com/totalreverse/ttyT1941/wiki:

> 'Load' is some kind of force.
> A force of 1 N ~= 137 'load units'


I will start by submitting a change to the PR to clarify units as variable name suffixes, and define all constants to ensure no magic values.

However we modify the driver, we should be clear what the unit is reporting and how to interpret it. There's no change numerically... we accept these constants as 289.75 and 137.

ericchristoffersen:
I found some tacx marketing literature that has graphs of various devices' braking limits:

https://www.qbp.com/diagrams/TechInfo/Tacx/MA9571.pdf

mattipee:
@ericchristoffersen 

> 1. Speed comparison: Is that device speed vs simulated speed?

The comparison I was making was the calculated device speed (ie raw speed / 289.75) against actual, real-world speed (cadence + gear + wheel circumference) - I think I've covered that now in the previous comment. :)


> Does no good to match sim speed if the trainer can't hold ... I don't think we should focus about sim and device speed matching.

I'm not trying to match any sim speed here... just read and report the speed of the trainer itself, and to reliably feed it back to the resistance calculations for ergo and slope. Within the Fortius driver.

Your PR is taking the reported power and feeding it into your sim to model intertia etc to derive the current simulated speed - which I like, and I think there is value in pursuing a way to match device speed with simulated speed given my observations and hacking around I think it might not be difficuly. But that's very much not what this PR is about. I'm just trying to baseline the Fortius driver and correct the maths on the master branch. To make slope mode work when the "Simulate Speed From Power" checkbox is unchecked.


> I think priority should be to investigate the low rpm power issue.

It's not my greatest priority, to be honest. I think the rest of this PR holds more value. The workaround that Wouter came up with is in... it simply limits any high resistance commanded at low speed, was derived empirically, and means there's no realism at low wheel speed. But it lets you get the speed up to where you'll likely be pedalling anyway. I do still see a little stuttering starting off, suggesting it's not perfect, but correcting the maths to implement ergo and slope within the Fortius driver is more important, in my opinion.


> 2. I really think you should disable the power smoothing from fortius.cpp, at least for testing the slope mode. That latency will hide the brake response latency. In slope mode power variance is smoothed because its applied through the bicycle+rider's effective mass. For erg mode... personally I think I'd rather see the noise and smooth later, but I don't really care and I know people like to pretend power is smooth and accurate.

Agreed... somewhat. The Fortius is rather good at providing samples that differ in value, and understandably. It's sampling periodically, at various points around the pedal stroke. The issue with ERG mode is that each cycle, it takes the current speed and the current resistance to report power - good, nicely fluctuating - and takes the current speed and the target power to set the resistance - perhaps not so good. The target power is obviously a fixed set point, but the current speed may just happen to correspond to a weak portion of the stroke... more significant at higher loads and lower cadence, if it thinks you're not trying hard enough and ups the resistance. I had issues with FortiusANT (see https://github.com/WouterJD/FortiusANT/issues/140, somewhat unconcluded)

We can remove all smoothing, and leave it bare, warts and all, and revisit if necessary. I think a smoothed wheel speed value to feed back into the resistance calculation in ERG mode may be an advantage. But let's start simple.


> 3. Assuming we must run a sim in the device driver, it would be better to plumb a reference to the bicyclesim down into the devicecontroller and then give it to the fortius to use for sim. This means all the simulation and constants would be the same, and we get a shared implementation.

The sim you're referring to is just maths right? A bunch of configured values, some formula and otherwise no state? I'm fine with that. Given slope, mass, wind resistance and speed and rolling resistance, and a constant for gravity, it's only half a dozen lines of code. I'm assuming the math of v*(Froll+Fair+Fslope) is pretty standard, and not in question. But I think the word "sim" is throwing me slightly... I'm not concerned here at all with feeding power into your model to derive a simulated speed... the FortiusController interface I thought already receives all the information it needs so it was a quick few lines to implement slope mode. But I'm all for shared code, so it there's an object on which I can call WattsForV() with the current trainer speed (ie raw_speed / 289.75, or in m/s if you prefer), that's perfectly fine. But an interface change to get the object reference passed down.

> At this point altitude is missing from controller interface, so that'd be a good thing to add. With that change we will still see a speed difference because device's implemented (physical) momentum will be different.

You've mentioned altitude a couple of times... whats the relevance? Is it not already taken into account in the value passed to setWindResistance() and therefore contribute to the calculation of the "Fair" component of the calc I'm doing already? Am I missing something? Does the constant of gravity change?? Can you explain what you mean by "device's implemented (physical) momentum will be different"? The last sentence of this paragraph is not a question. :)

ericchristoffersen:
> You've mentioned altitude a couple of times... whats the relevance? Is it not already taken into account in the value passed to setWindResistance() and therefore contribute to the calculation of the "Fair" component of the calc I'm doing already? Am I missing something? Does the constant of gravity change?? Can you explain what you mean by "device's implemented (physical) momentum will be different"? The last sentence of this paragraph is not a question. :)

I didn't realize you were comparing 'actual training bike wheel speed' with the speed reported by the trainer. I'm totally in favor of getting that math right. Of clarifying all the math. Good stuff that should never change once its correct. I do think its sort of irrelevant to the issue because the only speed the trainer needs to care about is its rpm, the tire contact diameter is the actual thing that is fixed.

What I keep going on about is that a good riding experience with a torque limited wheel-on trainer fundamentally means that trainer speed cannot match simulated speed. Adding the wouter workaround seems like its practically fine, but I think we should go further to encourage wheel speed to live in a practical zone where trainer can accept a casual out of saddle pedal stroke. That really is my only point. The physical limit means I will not be able to climb passes on a fortius with my mtb singlespeed because its gearing is too low and I would make too much torque for the trainer to hold. There is literally nothing that can be done to fix that because fortius can't hold my torque at my bikes max speed (12.5mph). A geared bike would be able to get the flywheel spinning faster and eat my power without slipping. This is a genuine physical limit that I think is more important than actual bicycle wheel speed.

Even if we did want wheel speed to match sim speed its not possible to actually get the sim and device to behave the same because the emulated inertia and device's (also emulated) inertia are different. We can adjust it using sim speed, but again its not sensible if this attempt pushes the wheel speed too low.

The device and rear bicycle wheel have inertia. The device is supposed to fake additional inertia to provide a 'realistic feel'. There's some function that describes the emulated inertia, it should not be speed sensitive but probably is: call it FortiusInertia(x) where x is speed. I'm just saying that FortiusInertia(x) won't match reality and won't match the sim, it might even be time dependent. Empirically that means the fortius will accelerate faster than the simulated mass. The difference between the inertia that fortius implements and the sim inertia will result in a difference between sim and training speed because the fortius will accelerate at a different rate than the sim. That means the fortius speed is uncoupled from sim speed, it will diverge and there is nothing to bring it back.

I only brought up altitude because its a value that is needed by the sim that isn't available in the current trainer api. Just one point, not a huge deal, it really just matters for descending because that is when air dominates.

The reason to inform the slope sim using bicycle reference instead of a busy set of trainer calls is that slope mode will automatically pick up any implementation extensions no matter what they are. I'm advocating that you literally run your own simstate against a reference to the current bicycle. Fortius needs its own simstate because it has its own speed.

The reason I think fortius must have its own speed is to keep it out of the low speed danger zone (torque limits at low speed.) Fortius will need a function to map from 'fortius device speed' to 'fortius sim speed', it might just be a simple offset so speed of zero is anything before 15-20kph, but we might want to have a linear remap where sim speed of 0-70kph maps to reasonable device speeds of 15-45kph.

Back to the comparison with resistance watts:

Device speed model is that rider steps hard, flywheel accelerates, driver sim raises resistance for faster flywheel, increased resistance allows rider to emit more watts, finally increased watts reach sim.

ResistanceWatts model: Brake resistance is in newtons. Rider steps hard. Increased wheel speed at x newtons should mean more watts reported to sim, sim speeds up and reports back a higher resistance to trainer. This runs a single sim and seems much simpler to me. I feel like I'm misunderstanding something basic.

Love these discussions. Sorry my forehead is so thick.

mattipee:
I think let's now keep this one simple. This PR implements behaviour equivalent to FortiusANT.

The summary is:
- accepted constants (137, 289.75) for speed/force calcs
- ergo mode calc been fixed
- slope mode calc has been re-implemented:
  - uses all available parameters on existing interface to calculate P(v)
  - relies on device's own virtual flywheel, however deficient
- calibration routine implemented
- simple resistance-limiting routine for low wheel speeds
- some code changes for clarity

Optional
- [remove power smoothing]

michelrdagenais:
> @michelrdagenais I hope you can find time to follow here...
> 
> The most basic question I had is "is the speed scale" correct?
> 
> This is the constant (289.75) by which we divide the raw speed value from the Fortius to obtain speed in km/h (and therefore in deriving m/s), one of the parameters to determining power output.

When I looked at the speed reported by Tacx software versus the values returned by the device, I obtained the following. This is quite close to 289.75.

 Bytes 33, 34 is the speed (little endian), 0x0800 : 5km/h, 0x0c00 : 10km/h, 0x1300 : 15km/h, 0x1600 : 20km/h, 0x1c00 : 25 km/h, 0x2200 : 30 km/h.

> > 'Load' is the target speed. load_speed = 'kph * 289.75'
> > TTS4 calibrates with value '0x16a3' and says that this is 20 kph.
> > 0x16a3 / 20 (kph) = 289.75.

Makes sense.
 
>   * 40.0 mm diameter, as best I can measure it
>   * 40*pi ~= 125.6637 mm circumference.
> * TACX trainer tyre

You should not need that. The roller speed is the tire speed. This it is the speed shown on your cyclometer, if your tire diameter is properly configured.

>   * (62.5 rpm * 60 min) * 3.111... * 2077.64 mm ~= 24.24 kph
> * Speed values from the Fortius
>   * 7043.17 / 289.75 = 24.3 kph
>   * an error of 0.25%
> 
> While this leaves me baffled as to the units of the raw speed values from the Fortius, I am confident that we accept 289.75 as the correct value for the speed scale.

Yes, totalreverse was also uncertain about the origin / units of this raw speed.

> The other value we have to accept is the constant (137) by which we divide the raw resistance value from the Fortius to obtain force in N, the other parameters to determining power output.

I have no better information. I am inclined to rely on their findings. This explains why the slope computation is incorrect currently.

> I will start by submitting a change to the PR to clarify units as variable name suffixes, and define all constants to ensure no magic values.

Indeed, clear code is easy to modify and improve!

mattipee:
> > While this leaves me baffled as to the units of the raw speed values from the Fortius, I am confident that we accept 289.75 as the correct value for the speed scale.
> 
> Yes, totalreverse was also uncertain about the origin / units of this raw speed.

Thanks, @michelrdagenais (cc @totalreverse)

I've found a numerical derivation of sorts...

- 0x16a3 = 5795
- 5795 / 1.609344 (km/mi) ~= 3600.846059
- 3600.846059 / 50 = 72.016921 m/s
- 3600.846 / 180 = 20.004700 kph

1.609344 * 50 * 3.6 = 289.681920 (~= 289.75)

src/Train/Fortius.cpp
```diff
@@ -46,24 +45,53 @@ const static uint8_t slope_command[12] = {
...
+};
 
+// From switchabl on FortAnt project, based on
+//  https://github.com/totalreverse/ttyT1941/wiki#newer-usb-1264-bytes-protocol
+//
+static const double s_ms_to_kph       = 3.6;
+static const double s_N_to_raw_force  = 137.0;
+
+// FIXME to have variables by which we can multiple raw values means that we use the inverse
+// However, double appears inaccurate enough that these small values introduce error (eg 20kph during calibration reports at 19.2kph or something)
```
ericchristoffersen, line 60: I think you need to root cause this precision issue. Some value is almost zero?

mattipee:
Superseded by new work on #3725

PR #3733, opened 2020-12-22
Expose "Device Speed" and "Simulated Speed" as data fields.

mattipee:
SimulatedSpeed value is always set regardless of whether sim is active or not... helps test things if I can see it either way.


amtriathlon:
It is fine you use this for testing purposes but to expose this to end users will likely be confounding factor, let’s try to keep it simple.

Issue #3319, opened 2020-01-27
Brake error on Tacx Flux S and 2


AnHaug:
With GC 3.5 and Tacx Flux S (newest firmware) the video display shows ("Bremsenfehler" - brake error). Workouts (non video) are not working meaning brake is not controlled by GC.

Platform: Windows 10

amtriathlon:
In the wiki FAQ there are some guidelines for troubleshooting and you can search the users forum where you will find mixed experiences and, eventually, ask for help.
This doesn’t seem to be a GC bug nor something we can give technical support here.

amtriathlon:
It seems to be a pattern with Tacx Flux using newer firmware, this is how it looks like according to the screen capture posted by another user in users forum:

![brake-fault](https://user-images.githubusercontent.com/1444784/74938546-f6f66c80-53cc-11ea-8705-8566f7803b41.jpg)

NB1: currently Trainer status is not displayed on notifications window, only on Video when Video Overlays are active, and this is available only in Windows, so it will be unnoticed on macOS/Linux.

NB2: since the Trainers reporting this error are working with other training applications I will reopen this as a feature request to support this device waiting for a developer with access to a Tacx Flux willing to work on this.

cyrusso:
Having the same issue with my new Tacx Flux S trainer, used to train on GC with my previous trainer Tacx Vortex Smart, but it had an issue and needed replacement. I've just set the new gear and was looking forward to run GC but it doesn't control the brake/resistance. I'm looking forward for device support on a new version of GC

amtriathlon:
Based in [this experience with other Tacx trainer](https://groups.google.com/d/msg/golden-cheetah-users/wFsMMiKYi5s/HX2LEKFCDQAJ) I would suggest to disable bluetooth on all devices (Notebook, Tablet, Cell Phone) around the trainer and try again.

vonmoock:
It is a problem with all the Flux modells S and 2. it has nothing to do with bluetooth decices. Wont work in a Garage where there is no other equipment than the trainer and a PC-Tower without bluethoot and with a Ant+dongle. The problem persitst up to the newest update from 12/14 (when I remember right the date). And it affects not onlythe rlv-vodeos but also the selfdesigned or not workouts of GC. I have no idea of coding, but if it helps sombody to betatest I am on your side with a Win 10 and Flux 2.

amtriathlon:
Thanks for the confirmation.
[...]

PR #3736, opened 2020-12-28
Add erg mode support for Tacx ANT over BTLE

amtriathlon:
This adds set target power support for Tacx ANT over BTLE protocol, intended for testing using TACX Flux S on https://github.com/GoldenCheetah/GoldenCheetah/issues/3319

### DIFF
--- a/src/Train/BT40Device.cpp
+++ b/src/Train/BT40Device.cpp
@@ -558,6 +558,9 @@ BT40Device::updateValue(const QLowEnergyCharacteristic &c, const QByteArray &val
             case INRIDE_STATE_SPINDOWN_ACTIVE:
                 notifyString = QString("SPINDOWN_ACTIVE-COAST TO STOP    : ");
                 break;
+            default:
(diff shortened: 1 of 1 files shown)


peret2000:
Maybe this a side effect of this build, coming from other changes cummulated to this version, or a problem with my installation, but, just in case:
With this build, I am neither able to synchronize with Strava (it says it lacks the token). not to link to Strava from scracth. I have tried also removing strava references from athele-private.ini, but the problem continues to happen.
Switching to a previous build (mine is Dec, 23rd) I can link the account and synchronize.

Can there be a format change in that ini file?

amtriathlon:
That is a known limitation of PR builds due to security reasons, this change is already on master, see https://groups.google.com/g/golden-cheetah-developers/c/5fZUO9DbJDY for details.

peret2000:
Understood! I read that thread of the forum but did not understand the implications. Sorry. Thanks!

Issue #3747, opened 2021-01-02
Build break in VideoWindow.cpp without GC_VIDEO_VLC


ChristophWurst:
While setting up my dev env to debug #3746 I was prompted with a compile error at https://github.com/GoldenCheetah/GoldenCheetah/blob/944f0ba73d1505b15a0a12fb0ca75f973359e94e/src/Train/VideoWindow.cpp#L460. Unlike the builds on Travis, I did not define `GC_VIDEO_VLC`, hence `m` is undeclared.

I could trace this regression back to the recent commit https://github.com/GoldenCheetah/GoldenCheetah/commit/5f68d2dce0c0e5e56c05574209d0bbd2fa077071.

@amtriathlon what shall be done when `GC_VIDEO_VLC` isn't defined? Should we always hide the live map widget in that case?

amtriathlon:
> Should we always hide the live map widget in that case?

Yes, overlay widgets are supported only with VLC

amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/d84d329a789ec8f43fec613799715bdccaccc718

PR #3748, opened 2021-01-02
Do not plot or show the live map without VLC

ChristophWurst:
`m` is only defined with VLC, hence this code causes a compile error when GC_VIDEO_VLC isn't defined.

Fixes https://github.com/GoldenCheetah/GoldenCheetah/issues/3747

@amtriathlon please let me know if this is the way you intended it.

### DIFF
--- a/src/Train/VideoWindow.cpp
+++ b/src/Train/VideoWindow.cpp
@@ -443,6 +443,7 @@ void VideoWindow::telemetryUpdate(RealtimeData rtd)
                 elevationMeterWidget->gradientValue = rtd.getSlope();
             }
         }
+#ifdef GC_VIDEO_VLC
         else if (p_meterWidget->Source() == QString("LiveMap"))
         {
             LiveMapWidget* liveMapWidget = dynamic_cast<LiveMapWidget*>(p_meterWidget);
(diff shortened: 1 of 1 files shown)


ericchristoffersen:
Current code shouldn't be testing for m, it wants to know if there is an active video. The livemap widget doesnt depend on 'm'.

I think that check for 'm' was added because there is some horrible race between video playback and html rendering.

None of these widgets make any sense unless there is an active video. Rather than more ifdefs could you instead add a new method to test that video is available to render onto? Put the ifdefs there:

bool VideoWindow::hasActiveVideo() const
{
#ifdef GC_VIDEO_VLC
    switch (libvlc_media_player_get_state(mp)) {
    case libvlc_Playing:
    case libvlc_Paused:
        return true;
    }
#endif
#ifdef GC_VIDEO_QT5
    return LoadedMedia == mp->mediaStatus();
#endif
    return false;
}

Then we can call it instead of testing 'm'.

ericchristoffersen:
Change #3764 takes care of this build issue.

ChristophWurst:
Sounds good. Then I'll close this PR :)

Issue #3749, opened 2021-01-03
Virtual Gearing For Sim Speed Mode


ericchristoffersen:
Implement what was discussed in this user group thread:

https://groups.google.com/g/golden-cheetah-users/c/G7yYcXf-UD4/m/4_E35N2fBAAJ

Currently, whether you're in gradient or ergo mode, the gc intensity adjustments rewrite the workout to be more difficult or easier.

If you click lower() enough in slope mode you can change a 1000m climb into a 500m climb.

As someone with only 3 functioning speeds on their trainer bicycle, I would much prefer if the raise/lower buttons just adjusted the intensity without actually changing the ride, so I don't need to actually shift. Ideal result for me would be that intensity buttons acted like virtual gears, so the feel of the slope would be less, but the climb itself would not change.

I have a friend who uses an old campy road bike on his trainer and its corncob ratio means he can't climb steep grades - he wants a way to adjust his gearing to something his bicycle won't do, and he also doesn't wish to dilute his climbs.

I can imagine a similar change being reasonable for ergo mode. You want to hit 400 watt intervals but aren't feeling it today, press lower() a few times and ergo mode will ask less of you, but the displayed workout won't change.

The case where this doesn't work perfectly is ergo mode for non-smart trainers. In that case I imagine wanting to see the actual power goals drop when I press lower, since in that case the target is all on me and not imposed by the trainer.

Should we add an option to have the intensity buttons act like virtual gears?

mattipee:
I have a lesser-option coded in https://github.com/mattipee/GoldenCheetah/tree/intensity, which is more of a first-step fix than a feature, and will:
- adjust grade by some factor, as course-rewriting did
- retain original course, so as to make sim-faithful progress along the original course

It's not virtual gearing, and only affect slope, not ergo.

I'd be Interested to get some feedback.

amtriathlon:
> I have a lesser-option coded in https://github.com/mattipee/GoldenCheetah/tree/intensity, which is more of a first-step fix than a feature, and will:
> 
> * adjust grade sent to trainer by some factor, as course-rewriting did
> * retain original course, so as to make sim-faithful progress along the original course
> 
> It's not virtual gearing

It looks similar to https://zwiftinsider.com/using-the-trainer-difficulty-setting-in-zwift/ but dynamically changeable, I like the idea since it is simpler to understand for the users and to verify it is working as intended, with simulate speed from power disabled it likely mimics actual behavior.

[...]

PR #3750, opened 2021-01-03
Virtual Gears

ericchristoffersen:
Issue #3749 

This change depends on change #3725.

### DIFF
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -77,6 +77,8 @@
 #include "TrainDB.h"
 #include "Library.h"
 
+double IntensifyNewtons(double N, double intensity);
+
 TrainSidebar::TrainSidebar(Context *context) : GcWindow(context), context(context),
     bicycle(context)
 {
@@ -86,6 +88,7 @@ TrainSidebar::TrainSidebar(Context *context) : GcWindow(context), context(contex
     setControls(c);
     autoConnect = false;
     useSimulatedSpeed = false;
+    intensityFactor = 1.;
 
     cl->setSpacing(0);
     cl->setContentsMargins(0,0,0,0);
@@ -1191,6 +1194,7 @@ void TrainSidebar::Start()       // when start button is pressed
         load = 100;
         slope = 0.0;
         resistanceNewtons = 0.0;
+        intensityFactor = 1.;
 
         // Reset Speed Simulation
         bicycle.clear();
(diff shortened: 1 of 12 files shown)


ericchristoffersen:
Rebased to 3725 latest.

mattipee:
Initial reaction is that TrainSidebar.cpp is getting populated with a lot of physics and tricksy maths. Is there sense in a refactor, or tend not to in order to keep risk/change down?

src/Train/TrainSidebar.cpp
```diff
@@ -2558,55 +2638,122 @@ void TrainSidebar::RewindLap()
...
+double IntensifyNewtons(double N, double intensity) {
+    return N * intensity;
+}
 
-        if (load <0) load = 0;
-        if (slope <-40) slope = -40;
+// When there is a divergence between simulated speed and device speed
+// adjust mass sent to device so it can emulate the simulated kinetic
+// energy.
+double TrainSidebar::getAdjustedInertialMass() const {
```
mattipee, line 2735: Do we need to account also for intensity factor? I haven't made the change and tried it, but discuss....?

ericchristoffersen: The connection with IntensityFactor is that when used in anger it, by design, causes a pretty massive divergence between sim and device speed. Is why I didn't notice the inertia problem until I started using the virtual gears. 1.1 or 0.9 is small potatoes and not so noticeable but with 5x difference in modelled speed the feel problem becomes obvious.

ericchristoffersen:
> Initial reaction is that TrainSidebar.cpp is getting populated with a lot of physics and tricksy maths. Is there sense in a refactor, or tend not to in order to keep risk/change down?

Definitely not as part of this change. I strongly prefer that a big reorganization is a functional nop.

I've been pushing for a long time to isolate training state from the trainsidebar. Is sort of absurd that a graphics element owns training state. I've been slowly isolating components from runtime state so that everything lives in trainsidebar, then eventually that state will go into its own object. Finally there will be a top level negotiator that trades data between devices, the training state and any 'future entities'.

There is a real need because some people want to add more complex controls and question is where and how that ui can work against current structure.

This is just my opinion. Plan needs to be discussed and agreed.

src/Train/TrainSidebar.cpp
```diff
@@ -1652,16 +1656,20 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
...
+                                                           // to within defined limits
+                    }
                 }
 
-                if (Devices[dev].type == DEV_FORTIUS || Devices[dev].type == DEV_IMAGIC) {
-	                rtData.setLoad(local.getLoad()); // and get load in case it was adjusted
-                    rtData.setSlope(local.getSlope()); // and get slope in case it was adjusted
-					// to within defined limits
-				}
+                if (!useSimulatedSpeed) {
```
ericchristoffersen, line 1666: Prior to this change, fortius device speed always overwrote sim speed, even if sim speed is enabled.

PR #3752, opened 2021-01-03
Auto Import Filtering / 30 day import / Only store in Import if importation successful / Cmd Line parameter processing

scr142:
Hi All,

   I’m a newbie to the project so please take the following change(s) as a suggestion, if I’ve not fully understand the project activity file workflow then I apologise, I’m happy to modify and get feedback if some of this work can be included….  

**Justification for change:**

I store my original fit/gpx files in a separate directory structure, and don't require the additional /Import directory storage GC provides, so I needed a way to disable this functionality to prevent unwanted file duplication.

**The change:**

This change adds an option in the Athlete Auto Import window to specify whether files from a particular Auto import directory are stored in /Imports or not. 

To make the change obvious when auto importing activities, I’ve added this information in read only form to the auto import dialog, and also when validating indicating to the user whether the file will be stored in /Imports or not. Once the “save” button is pressed, the additional field indicates whether it has been copied to imports successful or whether it is not copied as requested, or there is an error. 

The information on which of Auto Import directories to copy (or not) is stored in a new field in the AutoImport.XML file.
Looking through RideImportWizard.cpp I noted that there was the possibility that a file could be copied to /Imports, but fail validation and importation into GC which seemed wrong to me, so I have move the copy to /Imports functionality to only be invoked once a the importation is successful.

Additionally I have change the names of a couple of the status messages displayed in Status column of the Auto Import dialog as the two cases when an activity already exists in the /Activities directory aren’t really “Errors” for me, they are “Failed” importation issues, this leaves the “Error” to mean something fundamental has gone wrong.

To speed the Auto Importation process when the dialog is visible, I’ve added the option to only import the last 30 days, as I like to see the Auto Importation to check my recent activities are imported each time, but I don’t want to wait for 90 days of worth of duplicates to be checked.

I’ve simplified the calculation of the number of successfully imported files by using a counter, incremented only in the successful importation branch. This removes the dependency on searching through the Status messages after importation to look for the non-Error cases, IMHO not an ideal approach. 

And finally, I’ve re-enabled the Auto Import dialogues window close icon once the Finish button is available, as it seems logical that the user should be able to remove the window at this stage via “Finish” or “X” on the window banner. 

**Miscellaneous:**

In RideImportWizard.cpp & AlthletePages.cpp there is use of “magic” numbers to reference the field numbers, so I have created suitability named enums for the fields and use the enums to reference the fields, hopefully making the code easier to read:
RideImportWizard.cpp
enum WizardTable1 {
    DIRECTORY_RULE_COLUMN = 0,
    IMPORT_RULE_COLUMN,
    COPY_ON_IMPORT_RULE_COLUMN,
    DIRECTORY_STATUS_RULE_COLUMN,
    TOTAL_NUM_RULE_COLUMNS,
};
AlthletePages.cpp

enum WizardTable2 {
    DIRECTORY_AUTO_COLUMN = 0,
    IMPORT_AUTO_RULE_COLUMN,
    COPY_ON_IMPORT_AUTO_COLUMN,
    TOTAL_NUM_AUTO_COLUMNS,
};

Also in CMainWindow.cpp (line 1566) the RideImportWizard is create using a “new” call, then process, but the memory for RideImportWizard is never free’d and probably lost once out of scope, I have added "delete" to fix this, I also cahnged the pointer name from "import" to "importWiz" to prevent any confusion with language keywords.

**Help required:**

The format of the AutoImport.XML has been extended, does this need to be included in an Upgrade function?
Also adding the 30 days options to the importation has shifted the enum values for 90, 180 & 360 days, etc, this can easily be corrected by the user after an upgrade, so should this also be in an upgrade function?

ALSO

I have updated main.cpp handling of command line parameters so they take precedence over the many other layers of defaults, so now completely independent executable & athlete datasets can be maintained for stable and development builds. The chnages also remove a couple of missing error cases and hopefully IMHO improve the readability of the precedence of paths and athletes, given the path can be taken from the cmd line, local libary for USB, overridden by GC_HOME_FIR, old style libray path or new platform specific paths.....

This change can be independently taken from the changes above.


scr142:
I'm going to separate the changes in this pull request, so closing this request.

Issue #3756, opened 2021-01-04
Fix Video Hang By Synchronizing Widgets With Player State


ericchristoffersen:
I have been cursed with vlc hang-on-stop for the past several years. Usually my hang rate is 15-20%, but seems higher for difficult to decode videos like Silvretta. The hang I see is always waiting on vout in libvlc_media_player_stop.

Testing an unrelated change that caused a widget to spend much too long in paint, I found myself with a 95% hang rate with even the simplest videos, which led me to believe the root of the hang issue is widgets not honoring decoder stop.

A failure case is:

decoder under stress: lagging, seeking and trying to catch up
widget told to paint, qt grabs grabs a handle to what vlc calls vout (a resource handle?)
stop called, stop acts immediately, begins teardown
vlc_stop hangs waiting on vout, which is held by qt draw caused by widget paint.

This change adds a named state to videowindow and also a mutex for reading and writing that state. The important part of this change is to prevent widget calls from occurring after the process of 'stop' has begun.

With this change I have not been able to see a single hang after 50+ tries on 15 or so rlv, including these grand-slam hang favorites:
- stop right after start
- stop during seek
- stop after long normal playback

While making this change I took opportunity to remove some stuff that doesn't belong like the livemap management and this change also fixes the build break reported by #3747.

I have tested these change by trying to hang gc with 15 or so rlv files with vlc 3.0.9 and vlc3.0.10, but only on windows. Needs testing for other platforms.

ericchristoffersen:
https://groups.google.com/g/golden-cheetah-users/c/BxFF3clERXU

PR #3764, opened 2021-01-08
Fix VLC Hangs with Intentional State and Ordered Async Dispatch

ericchristoffersen:
Fix for #3756:

Two main parts to this change.
- a thread locked managed player state
- all calls to VLC are now made via an ordered Async dispatch.

Probably the ordered async dispatch is the more important of
the two as it moves calls to vlc stop() off the gui thread.

Tested by repetitively seeking and stopping on 50+ virtual rides on windows 10 with vlc 3.9.0 and 3.10.0.

Zero hangs in three days of trying.

### DIFF
--- a/src/Train/VideoWindow.cpp
+++ b/src/Train/VideoWindow.cpp
@@ -27,6 +27,14 @@
 #include "MeterWidget.h"
 #include "VideoLayoutParser.h"
 
+class Lock
+{
+    QRecursiveMutex& mutex;
+public:
+    Lock(QRecursiveMutex& m) : mutex(m) { mutex.lock(); }
+    ~Lock() { mutex.unlock(); }
+};
+
 VideoWindow::VideoWindow(Context *context)  :
     GcChartWindow(context), context(context), m_MediaChanged(false), layoutSelector(NULL)
 {
@@ -40,6 +48,8 @@ VideoWindow::VideoWindow(Context *context)  :
     videoSyncTimeAdjustFactor = 1.;
 
     curPosition = 1;
+
+    state = PlaybackState::None;
 
     init = true; // assume initialisation was ok ...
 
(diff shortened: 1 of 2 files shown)


ericchristoffersen:
This creates and destroys threads all the time which is rude. I am redoing with a threadpool.

ericchristoffersen:
Modified so instead of chaining futures it now runs everything on a traditional queue processed by a worker thread.

Sigfreid74:
I've tested this change the skipping fwd / bwd in rlv video is definitely more responsive. No hangs so far.

human705:
I've been testing this for about a week with no issues.

Sigfreid74:
I would love to see this improvement included soon as I'm having to use the current snapshot build which does not have this for the sake of Strava functionality and have gone back to having hanging VLC behaviour.

amtriathlon:
Since this is change affects Linux and macOS builds it needs to be tested on these platforms before merging. All testing reported seems to have been done on Windows only, AFAIK.

peret2000:
Ops! I did not realize those tests were expected...
So, I can tell my experience: I have been using that patch (merged into my own development fork, I have to say) and it seems to work nicely, on Ubuntu 20.10 (release and debug deliveries)
Maybe it was not the root cause of my problems, so I did not want to add 'noise': since december I was experiencing that videos froze, when synching with the ride, and updated frozen frames each few seconds. It was annoyning. I tried many changes: going back through thhe code, removing commits, etc, with no success.
Finally, merging with the branch of this pull request, the prooblem dissapeared, I have been using it for several weeks. So, although I was not convinced this was the solution, as I did not understand very well tha changes and the original problem, I would say that it has solved my issue; I am very happy.
And, anyway, it is not causing any trouble to the application; in my experience, it can be merged safely

human705:
Can you please share the instructions you followed to build  on Ubuntu 20?

peret2000:
I build on Ubuntu 16.04, and then deploy the binary another machine. I meant that the binary works fine on Ubuntu 20.10

amtriathlon:
> Can you please share the instructions you followed to build on Ubuntu 20?

[INSTALL-LINUX](https://github.com/GoldenCheetah/GoldenCheetah/blob/master/INSTALL-LINUX) is now updated to build in Ubuntu 18.04, it should help to build on 20.xx, please use developers forum for questions if you found problems.

mfkl:
FYI, LibVLC 4 introduces proper async stop, so this code may be removed in the future (when v4 goes stable).

PR #3766, opened 2021-01-09
Map chart changes

paulj49457:
Functional changes:
i) Added fix to display the complete shaded route, previously a section of route too short to an interval wasn't drawn, so now any route segment too short at the end of the activity is now included, although in a shorter interval element.
ii) New check box added to Map menu to allow the user to select the display of the shaded route (avg watts per 60s interval). This switches the displayed route between the complete activity (red line, aka "the old thick yellow line") to the segmented view of the shaded route, colours dependent on wattage.
iii) changed the complete activity line colour from the old thick yellow line to a narrower red line to improve visibility.

Non-functional changes:
iv) Removed the unused styleOptions functions and associated Q_PROPERTY.
v) Renamed the types that had the colour "Yellow" in their name, given that the colour isn't yellow anymore
iii) Added some enums for the possible map tile selections, rather than 0, 10, 20, etc, hopefully making the code more readable.


paulj49457:
I'll re-submit this without the unwanted files, sorry my bad config control, still learning Git.

PR #3767, opened 2021-01-11
Fixed command line parameter handling

paulj49457:
I have updated main.cpp command line parameter handling, so now these parameters take precedence over the many other layers of defaults, allowing completely independent executable & athlete datasets to be maintained for stable and development builds. The changes also remove a couple of missing error cases and hopefully simplify the precedence of library paths and athletes.

The library path is now determined in priority order by:
i) command line
ii) overridden by GC_HOME_DIR
iii) local library for USB
iv) old style library path
v) and finally new platform specific paths.....

The athlete is now determined in priority order by:
i) command line
ii) last athlete used in GC, assuming the athlete exists in the path determined above.
iii) otherwise GC will open with a menu for athlete selection

### DIFF
--- a/src/Core/main.cpp
+++ b/src/Core/main.cpp
@@ -254,7 +254,6 @@ main(int argc, char *argv[])
     bool server = false;
     nogui = false;
     bool help = false;
-    bool newgui = false;
 
     // honour command line switches
     QString arg;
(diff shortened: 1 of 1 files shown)


src/Core/main.cpp
```diff
@@ -535,67 +535,154 @@ main(int argc, char *argv[])
...
+            if (home.exists(PlatformlibraryPath)) {
+                home.cd(PlatformlibraryPath);
+                homeSetup = true;
+            } else { // things are getting bad, lets try and create the platform directory
+                if (!(homeSetup = home.mkpath(PlatformlibraryPath))) {
+                    // tell user why we aborted !
+                    QMessageBox::critical(NULL, "Exiting", QString("Cannot create library directory (%1)").arg(PlatformlibraryPath));
+                    exit(0);
                 }
-                home.cd(libraryPath);
```
amtriathlon, line 584: Have you tested this path?
It seems to me the working directory is not see accordingly when a new library is created.

paulj49457: Hi Ale,

     I did test numerous combinations, please put the PR on hold for the moment and I'll take a look at the issue.

Regards,
Paul.

paulj49457: Hi Ale,

    You were right the branch had been lost in my proposed change, it has now been added, great spot thank you.

    Unfortunately I have another issue with the change....   when the user provides a new Athlete Directory on the command line the GC_HOMEDIR should be updated by the call to        

            appsettings->initializeQSettingsGlobal(gcroot);

    but I'm find that "newFormat = true" (see code below) and so the call just returns without updating GC_HOMEDIR. This results in GC working correctly with the

new Athlete directory from the command line because gcroot is set, but the config dialog displays the old path because it uses GC_HOMEDIR which didn't get updated by the main function.

void  GSettings::initializeQSettingsGlobal(QString athletesRootDir) {

    if (!newFormat) return;

    if (global->isEmpty()) {

        global->append(new QSettings(athletesRootDir+"/"+settingFileNamesGlobal[GLOBAL_GENERAL],QSettings::IniFormat));
        global->append(new QSettings(athletesRootDir+"/"+settingFileNamesGlobal[GLOBAL_TRAINMODE],QSettings::IniFormat));

    }

Any help on resolving this issue would be great.

Thanks,

Paul

amtriathlon:
See https://groups.google.com/g/golden-cheetah-users/c/qzGSZrVV5Sg/m/iskESxS_BgAJ

paulj49457:
> The problem are settings, which are not included inside the library, the code supports 2 options: standard location and local library.
> The 2 parameters command line option should be deprecated, IMHO.
> See https://groups.google.com/g/golden-cheetah-users/c/qzGSZrVV5Sg/m/iskESxS_BgAJ

I use the two arg option so would like to see it retained, but I can see that improved clarity would help. The active path is displayed in the help/about menu but that is probably the last place I'd have looked. So to avoid confusion can I propose this information is included in the options menu, I can add a tooltip on the startup default path to explain that it can be overridden from the commandline, and is the case is the paths are different, the "active" field is read-only, so the new proposed option layout which I can create in a new PR would be:

image

paulj49457:
I'm going to close this PR, as it the existing two argument option provides the required functionality. I now use the following batch file to configure my Windows stable GC version (using GC provide python 3.7), and another similar script to configure my development GC environment which uses python 3.9, and has a different GC user data location. The Powershell scripting is required as both version share the GoldenCheetah.ini file.

```
REM - setup the python environment for the stable version of GC
set GC_INI_FILE=C:\Users\Paul\AppData\Roaming\goldencheetah.org\GoldenCheetah.ini
powershell -Command "(Get-Content %GC_INI_FILE%) -replace '^homedirectory=.*','homedirectory=C:\\Users\\Paul\\GoldenCheetah' | Out-File -encoding ASCII %GC_INI_FILE%"
powershell -Command "(Get-Content %GC_INI_FILE%) -replace '^embedPython=.*','embedPython=true' | Out-File -encoding ASCII %GC_INI_FILE%"
powershell -Command "(Get-Content %GC_INI_FILE%) -replace '^pythonhome=.*','pythonhome=C:/Program Files/Golden Cheetah' | Out-File -encoding ASCII %GC_INI_FILE%"

set STABLE="C:\Program Files\Golden Cheetah\GoldenCheetah.exe"
set DATAHOME="C:\Users\Paul\GoldenCheetah"
set ATHLETE=Paul

REM Need to use --no-angle to improve graphics performance with versions v3.6 and above
start "" %STABLE% %DATAHOME% %ATHLETE% --no-angle
```

But I think the proposed **Improve Clarity of GC Paths in Dialogs #3878 PR** would clarify what paths GC has selected and is using, and move their display to be logically located.

paulj49457:
I'm going to close this PR, as it the existing two argument option provides the required functionality. I now use the following batch file to configure my Windows stable GC version (using GC provide python 3.7), and another similar script to configure my development GC environment which uses python 3.9, and has a different GC user data location. The Powershell scripting is required as both version share the GoldenCheetah.ini file.

```
REM - setup the python environment for the stable version of GC
set GC_INI_FILE=C:\Users\Paul\AppData\Roaming\goldencheetah.org\GoldenCheetah.ini
powershell -Command "(Get-Content %GC_INI_FILE%) -replace '^homedirectory=.*','homedirectory=C:\\Users\\Paul\\GoldenCheetah' | Out-File -encoding ASCII %GC_INI_FILE%"
powershell -Command "(Get-Content %GC_INI_FILE%) -replace '^embedPython=.*','embedPython=true' | Out-File -encoding ASCII %GC_INI_FILE%"
powershell -Command "(Get-Content %GC_INI_FILE%) -replace '^pythonhome=.*','pythonhome=C:/Program Files/Golden Cheetah' | Out-File -encoding ASCII %GC_INI_FILE%"

set STABLE="C:\Program Files\Golden Cheetah\GoldenCheetah.exe"
set DATAHOME="C:\Users\Paul\GoldenCheetah"
set ATHLETE=Paul

REM Need to use --no-angle to improve graphics performance with versions v3.6 and above
start "" %STABLE% %DATAHOME% %ATHLETE% --no-angle
```

But I think the proposed **Improve Clarity of GC Paths in Dialogs #3878 PR** would clarify what paths GC has selected and is using, and move their display to be logically located.

Issue #3696, opened 2020-12-06
Deletion of workouts for auto-import


ChrisKehr:
I set up GC to **auto-import** from different locations (e.g. Dropbox folder, that contains Wahoo activities from riding outdoors or my local Zwift folder for indoor rides).

It works fine. However, if I choose to **delete a file** inside GC for some reason (e.g. I have recorded an indoor ride with Zwift as well as my Garmin watch for comparison of accuracy and only want to keep one of them), it **will reappear during the next start** of the program.

The only solution is to delete the files in their original directory. Which is not always intended...

I am not a programmer, but I can imagine a small data file that contains previously deleted files or at least their meta-data, and will exclude these files during the next import.

Best regards,
Chris

amtriathlon:
Every file you import is automatically copied to imports folder under your athlete’s directory, we could eventually to give the option to also delete the original file to avoid this behavior and to speed up auto-import (https://github.com/GoldenCheetah/GoldenCheetah/issues/1892), but to track deletions is not something we plan to do.

scr142:
The "Auto Import Filtering" #3770, change which I recently proposed has a field in the auto import window to display whether the file is going to be stored in /imports or not. Therefore I could make this field a ComboBox and allow a ignore option, which would then write the file name to a txt file stored in /Imports. Then everytime auto import runs it could parse this file and  set the ignore option automatically. Happy to work on this provided its would become part of the product.

ChrisKehr:
That would be great! That's exactly what I was thinking of!

scr142:
So attached is the proposed new look Auto Import window, with the function to select whether to store files in /import depending on their source directory (at the top of the window), then for information this info is displayed against each individual file, and this is the place where the user could toggle the choice from Import, no import or ignore.

I think what's required is some agreement from the more experienced GC developers that this is ok.

![Auto Import Menu](https://user-images.githubusercontent.com/46629337/104371311-978ad900-5517-11eb-8c79-b52c1346ab14.JPG)

paulj49457:
Hi Chris,

   I'm working on this change, but need to know whether you want the fullpath+filename or just file name stored?

For example the entries in the file would look like:

D:/Paul/Sport/Training Data/Running/20210108-160749-1-1328-ANTFS-4-0.FIT   or
20210108-160749-1-1328-ANTFS-4-0.FIT

Can you let me know what you had in mind ?

Thanks,
Paul.

[...]

Issue #621, opened 2013-05-30
Re-introduce overwrite files on import


liversedge:
That was temporarily deprecated in v3.

scr142:
I'm not sure which files are to overwritten in import (/import, /activity or /cache) and why this might be required, but as I've proposed some changes in the Auto Import dialog recently I'm happy to take a look at this, but will need some more specific information first.

amtriathlon:
Activities are identified by Date+Time, when the file you import collides the process is cancelled with an error message, before v3 the user had the option to overwrite the existing activity, IIRC.

paulj49457:
So the existing overwrite functionality appears to work for all imported files, so is there a requirement to have overwriting selectable for each imported directory, or even per file basis? It would be useful to know before I set off and implement something.

liversedge:
The overwrite on import flag means for all files. Would be good to keep a count and add it to the message at the end, e.g. 1000 files imported, 400 overwritten.

paulj49457:
Hi Mark,
Thank you for the guidance on file overwriting, I also see the commented out code in RideImportWizard.cpp, which is an area which I have worked on quite a lot recently due to #3770 Auto import filtering and file import exclusion. So I'll work on reinstating the optional overwrite capability for all activity files with conflicting times/filenames. I'll base these changes on those in #3770, as I have already a more expanded reporting of importation status (see pics in #3770) and I can extend this to include overwrites, also building on #3770 will simplify merging the changes back into the main build.
Regards,
Paul.

paulj49457:
This is now implemented in Auto import filtering, file import exclusion and file overwrite option #3770

PR #3770, opened 2021-01-12
Python processor import fix, file import exclusion and file overwrite option

paulj49457:
This PR contains the following fixes to issues that are all closely related, so they are combined in this single update, plus a number of tidy ups of the ImportWizard processing. The change for the Python processing was made simpler by the restructuring for the improved import reporting:

The fix to allow python processors to work on import (issue: Python API - activityMetrics and setTag/delTag/hasTag don't work in automatic DP #4095)

The capability to exclude files from importation has been requested (Issue: Deletion of workouts for auto-import #3696)

The capability to overwrite files  #621 "Re-introduce overwrite files on import" has been reinstated.

Improved import file reporting status for #621, #3696 & #4095

**Python processors to work on import change**

The Python processors failed to work correctly as the Python data processor passes the context and ride file to the Python environment, and they must be matching. The original import wizard code was written before the introduction of the python processors and the ride file was processed before the corresponding ride item was created. So although the raw data processing aspect of the python processor would work, any modification or tests of the metadata tags (using hasTag/setTag/delTag, etc) or processing that relied upon them would fail. The built-in core data processors are immune to this issue as they only work on the raw ride file data which only requires the ride file so they work correctly. The change has been to move the invocation of the "import" data processing after both the ride file and ride item have been created, and to ensure the ride item in the context matches. 

**The File Import Exclusion change:**

The exclusion of imported files is also now possible, as the import window validating process now checks each file to see whether it is in the exclusion list (./imports/GC_Auto_Import_Exclusion_List.txt) and marks imported files as excluded if they are in the list. You can manually edit the exclusion list and add the full path and filename on the next empty line for any files you wish to exclude.

Alternatively, an easier way to exclude activities is during import you can use the ComboBox _(double left click on the entry's import field, see picture below)_ to mark it as excluded while the Save Button is displayed. It will then be added to the exclusion list and not imported, this is simplest way to create the exclusion list and build up its entries.  If the exclusion list doesn't exist, GC will create the exclusion file in /imports and add any excluded files.

CaptureA

To remove an activity from the exclusion list you just need to manually delete it from the list, as adding this interactively would require structural change to the import processing, as when the Save Button is displayed GC has completed some of the processing. Therefore the ComboBox is disabled for all activities with an entry in the exclusion list.

Given the new import states, the auto import status when the Finish button is displayed now identifies the number of files successfully imported, those activities that already exist, those excluded and those that created an error, see improved reporting below.

**Reinstated the overwrite files option.**

 #621 "Re-introduce overwrite files on import" has now been implemented, and the successful import status messages now indicate whether the activity is new or an overwritten existing activity. Additionally the import overall status has been updated to reflect the number of files and their type of import processing.

_Import Complete: successful [ newfiles %1, overwritten %2 ]   unsuccessful [ existing skipped %3, excluded %4, errors %5 ]   total of %6 files._

CaptureC

**Improved imported file reporting status for #621, #3696 & #4095:**

Once the “save” button is pressed, the additional "Import" column indicates whether a file has been copied to imports successfully, is excluded from import, or it is not copied due an import processing error.

The Import Status column now includes improved reporting of processing to more clearly distinguish between typical failed processing cases and the real error cases:

**_Typical reports:_**
Failed - File added to the exclusion list in /imports  _(user had requested the activity be added to the exclusion list)_
Failed - File is already in the exclusion list in /imports _(File previously set as excluded from import)_
Failed - Activity already exists in /activities  _(typical case when file has previously been imported)_ 
Import successful - Overwritten in /activities _(only occurs if overwrite files has been selected)_
Import successful - Saved in /activities _(new activity is imported)_

**_Exceptional reports (error cases):_**
Failed - Activity with same start date/time exists in /activities _(usually due to a time zone issue)_
Error - Import of activity file failed
Error - .JSON creation failed
Error - %1 cannot be found in RideCache!
Error - Moving %1 to /activities
Error - %1 cannot open ride file to process!

**Corner Cases**

In RideImportWizard.cpp there was the possibility that a file could be copied to /Imports, but fail validation and importation into GC which seems wrong, so I have moved the copy to /Imports functionality so that it is only invoked once the importation process is successful.
 
Another corner case is that it is possible to import a new activity successfully, but there could already be a file in the imports area that has the same file name as the imported file, in this case to maintain the integrity between the /activities and files in /import directories the old imported file will be overwritten.

**Update to the Save import processing stage:**

Access to the new RideItem has been modified to avoid the implicit use of the last selected ride, making it more robust, and allowing the signalling of the only the last ride. The progress bar has been updated to take account of the split processing for duplicate, excluded and errored files, and those which can be successfully processed.

**Miscellaneous:**

And finally, the Auto Import dialogues window close icon once the Finish button is now available, as it seems logical that the user should be able to remove the window at this stage via “Finish” or “X” on the window banner.

Ensured /tmpActivities is tidied at the end of a completely error free importation. In all other cases of errored importation, or GC ungracefully exiting then the contents of /tmpActivities are retained for fault finding.

### DIFF
--- a/src/Core/RideCache.cpp
+++ b/src/Core/RideCache.cpp
@@ -256,14 +256,14 @@ RideCache::itemChanged()
 }
 
 // add a new ride
-void
+RideItem*
 RideCache::addRide(QString name, bool dosignal, bool select, bool useTempActivities, bool planned)
 {
     RideItem *prior = context->ride;
 
     // ignore malformed names
     QDateTime dt;
-    if (!RideFile::parseRideFileName(name, &dt)) return;
+    if (!RideFile::parseRideFileName(name, &dt)) return NULL;
 
     // new ride item
     RideItem *last;
@@ -316,6 +316,8 @@ RideCache::addRide(QString name, bool dosignal, bool select, bool useTempActivit
 
     // model estimates (lazy refresh)
     estimator->refresh();
+
+    return last;
 }
 
 void
--- a/src/Core/RideCache.h
+++ b/src/Core/RideCache.h
@@ -92,7 +92,7 @@ class RideCache : public QObject
 	    QVector<RideItem*>&rides() { return rides_; } 
 
         // add/remove a ride to the list
-        void addRide(QString name, bool dosignal, bool select, bool useTempActivities, bool planned);
+        RideItem* addRide(QString name, bool dosignal, bool select, bool useTempActivities, bool planned);
         void removeCurrentRide();
 
         // export metrics in CSV format
--- a/src/Gui/RideImportWizard.cpp
+++ b/src/Gui/RideImportWizard.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2006 Sean C. Rhea (srhea@srhea.net)
+ * Additions Copyright (c) 2021 Paul Johnson (paulj49457@gmail.com)
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the Free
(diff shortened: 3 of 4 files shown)


paulj49457:
This PR has been superseded by #4336

PR #3777, opened 2021-01-13
Ignore SetLoad in Null trainer

ericchristoffersen:
NullTrainer was honoring SetLoad which makes no sense and breaks testing. Null trainer should only generate load, not honor what is set.

### DIFF
--- a/src/Train/NullController.h
+++ b/src/Train/NullController.h
@@ -57,7 +57,8 @@ class NullController : public RealtimeController
         bool doesLoad() {  return false; }
 
         void setMode(int);
-        void setLoad(double watts) { load = watts; }
+        // Dont set load for null trainer: it generates fixed power regardless of load.
+        void setLoad(double) { /*load = watts*/; }
         void getRealtimeData(RealtimeData &rtData);
         void pushRealtimeData(RealtimeData &rtData);
 


amtriathlon:
I think this is intended since it allows to use the Robot device to simulate ERG mode, manual or using a workout, which is useful for demos and general understanding. This change breaks that functionality without any benefit I can see.

ericchristoffersen:
> I think this is intended since it allows to use the Robot device to simulate ERG mode, manual or using a workout, which is useful for demos and general understanding. This change breaks that functionality without any benefit I can see.

Null trainer is a strange cookie because it emulates both a device and rider.

The workout sets load to aspire to. The current null trainer behavior honors set load as a command to the rider. Current behavior doesn't let me model, graph and analyze the case where load doesn't match desired erg load.

amtriathlon:
> Current behavior doesn't let me model, graph and analyze the case where load doesn't match desired erg load.

There is a random variation added, may be the amplitude can be made larger. Other than that may be you can change when you are debugging without removing a useful functionality.

ericchristoffersen:
> > Current behavior doesn't let me model, graph and analyze the case where load doesn't match desired erg load.
> 
> There is a random variation added, perhaps the amplitude can be made larger (and initial load higher to be more useful in simulated slope mode). Other than that may be you can change when you are debugging without removing a useful functionality.

The change won't seem useful until trying to demo intensity adjustment.

amtriathlon:
Sorry but I don’t understand how this can be useful in a our current setup and it turns unusable a basic feature s.t. to use the Robot device in manual erg mode, besides workouts.

PR #3778, opened 2021-01-13
Trainer Work Part 2: add SetGradientWithSimState

ericchristoffersen:
Replaces setgradient with extended version that provides more information to device.
Fortius and Monark rely on the force info, speed is there so device controller can compute watts and also so device controller can attempt to synchronize sim and device speed.

### DIFF
--- a/src/Train/RealtimeController.h
+++ b/src/Train/RealtimeController.h
@@ -102,7 +102,12 @@ public:
 
     // only relevant for Computrainer like devices
     virtual void setLoad(double) { return; }
+private:
+    // This function exists to be overridden by existing drivers, but should
+    // not be called by simulation loop.
     virtual void setGradient(double) { return; }
+public:
+    virtual void setGradientWithSimState(double gradient, double, double) { setGradient(gradient); }
     virtual void setMode(int) { return; }
     virtual void setWindSpeed(double) { return; }
     virtual void setWeight(double) { return; }
--- a/src/Train/RealtimeData.cpp
+++ b/src/Train/RealtimeData.cpp
@@ -25,7 +25,7 @@ RealtimeData::RealtimeData()
 {
     name[0] = '\0';
     hr= watts= altWatts= speed= wheelRpm= load= slope= torque= 0.0;
-    cadence = distance = altDistance = virtualSpeed = wbal = 0.0;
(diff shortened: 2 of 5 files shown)


mattipee:
@ericchristoffersen 

With the "hook" now in Fortius for "targetNewtons" in slope mode, and calibration done, I'm slowly starting to think about this again. I'm also starting to ride again more. :)

Just a quick one at the moment - I'm confused why, since the first PR got merged, the diff for this PR is still showing me all the "if (status & RT_WORKOUT) ..." differences against GoldenCheetah:master... will rebasing or merging master into this branch fix that?

mattipee:
If we can implement virtual gears in terms of setGrade() and setWeight(), do we _need_ to also be able to supply resistanceNewtons through a SetGradientWithSimState()?

Excerpt from my email to you:
- "resistanceNewtons" over the RealtimeController interface felt like the perfect answer, when we were dealing with Fortius
- for a generic solution, we can't use that, giving rise to GradientForIntensity() over the existing setGradient()
- If GradientForIntensity() is mathematically sound, then perhaps we don't need the complication of "resistanceNewtons" on RealtimeController interface??
- Unless GradientForIntensity() is best-effort, and we'd prefer to use "resistanceNewtons" where-at-all-possible??

There's a discussion still to be had there.

But even if resistanceNewtons _is_ "better" for virtual gears (for those trainers that would be able to support it), it might be better to introduce the generic GradientForIntensity() for all trainers first, and later "optimise" by proposing resistanceNewtons, if it makes sense to do so. Basically, to "prefer no interface change" to begin with.

That could mean redoing PR 3 https://github.com/GoldenCheetah/GoldenCheetah/pull/3779 without first having introduced setGradientWithSimState(), combining at the same time with the fourth PR https://github.com/GoldenCheetah/GoldenCheetah/pull/3780 for intertia. I would also suggest a checkbox or drop-down to choose between gradient-scaling and virtual-gears.

But I think in fact virtual gears as a PR should end up being easier and having a smaller footprint if my PR https://github.com/GoldenCheetah/GoldenCheetah/pull/3809 (a fix for the existing implementation, independent of introducing a feature) was able to reach approval and go in first.

Thoughts?

PR #3779, opened 2021-01-14
Trainer Work Part 3: Virtual Gears

ericchristoffersen:


ericchristoffersen:
Issue #3749

Mickyduck55:
In the initial discussion (which I now cannot find) there was a small debate about ERG mode and if this should be affected by the virtual gearing.  I did a ERG workout today.  Using the up / down arrows on my head unit the decrease / increase was in increments of 10 % .  In virtual gearing the value decreases  / increases buy 10 % of the previous value so it looks like the virtual gearing is not invoked in ERG mode which is what was agreed.

src/Train/BicycleSim.cpp
```diff
@@ -356,26 +356,80 @@ struct MotionStatePair
...
+    //     - -----------------------------------
+    //                    2     3/2
+    //                  (m  + 1)
+
+    // Coefficient for velocity - dependent dynamic rolling resistance, here approximated with 0.1
+    static const double s_CrV = 0.1; 
+
+
+    // gradient to slope and clamp to -1 to 1.
+    double m = std::max<double>(-1, std::min<double>(1, simState.Slope() / 100));
```
michelrdagenais, line 405: The current slope is used to compute the fortified gradient to set
That fortified gradient is sent to the devices
Fortius and possibly other devices used to clip the received gradient to -5% - +20%
FortiusController does rtData.setSlope(myFortius->getGradient()); to report the clipped gradient
That fortified gradient/slope is thus used here for the next computation and gets again fortified for the next round
In manual slope mode, within a few seconds, if "fortify" is used, we get to maximum 40% slope

Some cleanup is required for that in GC. This patch is the straw that breaks this otherwise fragile construct. I believe that we need clearly identified and documented metrics such as, for example:

slope: the slope seen in the workout file
requestedGradient: the gradient command sent to the device
appliedGradient: the gradient that the device actually applied
...

ericchristoffersen: Part of this change is trainsidebar lines 1654-1655 which removes the reading of slope from device for both fortius and computrainer. I have a computrainer and can attest at how utterly useless the computrainer's returned device value is.

Suppose I left the device slope and load reads in, what exactly is trainsidebar supposed to do with the clipped values returned by device? Would you have trainsidebar compute a different altitude? As far as I can tell there's never been coherent handling of those returned values.

amtriathlon:
I am closing this since it will not be merged due to merge conflicts, lack of a clear description and inclusion of unrelated changes.

PR #3780, opened 2021-01-14
Trainer Work Part 4: Inertia compensate

ericchristoffersen:
Scale the weight reported to device running in gradient mode so that it can model correct kinetic energy. This is important when sim speed and trainer speed diverge, particularly when virtual gears are used.


amtriathlon:
I am closing this since it will not be merged due to merge conflicts and inclusion of unrelated changes.

PR #3782, opened 2021-01-14
Trainer Work Part 5: Fortius Support

ericchristoffersen:
Moved from #3725, stuck at end of chain of changes, instead of at the start.


amtriathlon:
I think the next step on Fortius support should be to add calibration, ideally without dependence on still unmerged PRs.

Issue #3737, opened 2020-12-28
Monark Trainers Should Use New ResistanceNewtons Value


ericchristoffersen:
I was browsing other trainers and see that Monark trainers expect a resistance set in KiloPonds:

https://en.wikipedia.org/wiki/Kilogram-force

The current code to implement monark gradient mode does:

void MonarkController::setGradient(double gradient)
{
    // Repurpose gradient as kp for Monarks
    m_monark->setKp(gradient);
}

After fortius change goes in every trainer will have access to NewtonsResistance. The setGradient method for monark should be swapped to:

void MonarkController::setGradientWithSimState(double, double resistanceNewtons, double)
{
    m_monark->setKp(resistanceNewtons * 9.80665);
}

amtriathlon:
Monark are not regular trainers but medical grade ergometers, for them gradient is used to control the force mode: https://github.com/GoldenCheetah/GoldenCheetah/pull/1908/commits/a765d99534a87a7e08b7df1efb9f22d9b08a8851
A quick search will show the original developer had access to this device and he has tested this module on his device and other models, it is not a bug due to lack of understanding what a kilopond is, IMHO.

ericchristoffersen:
> Monark are not regular trainers but medical grade ergometers, for them gradient is used to control the force mode: [a765d99](https://github.com/GoldenCheetah/GoldenCheetah/commit/a765d99534a87a7e08b7df1efb9f22d9b08a8851)
> A quick search will show the original developer had access to this device and he has tested this module on his device and other models, it is not a bug due to lack of understanding what a kilopond is, IMHO.

This is quite a surprising response to read.

I read the data brochures and also some research group's setup instructions. The monark Kp really is a kilopond and not a shorthand for gradient.

amtriathlon:
> I read the data brochures and also some research group's setup instructions. The monark Kp really is a kilopond and not a shorthand for gradient.

I know, and so do the original developer.

ericchristoffersen:
> > I read the data brochures and also some research group's setup instructions. The monark Kp really is a kilopond and not a shorthand for gradient.
> 
> I know, and so did the original developer.

Your responses are leaving me lost and my search foo must be weak. Can you please link to any discussion about this 'gradient' approach for monark?

The current implementation is expedient but cannot be correct.

amtriathlon:
[...]

PR #3783, opened 2021-01-14
Monark implement SetGradientWithSimState

ericchristoffersen:
Previous monark implementation used gradient for kiloponds. New implementation uses force directly.
Previous monark implementation invented watts using cadence when kp wasn't supported by device, now uses watts directly.

Erg mode is unchanged.

This change is branched off #3778.


ericchristoffersen:
Issue #3737

amtriathlon:
If it ain't broke, don't fix it.

PR #3785, opened 2021-01-15
Initial correction of force/speed/power calculations for Fortius

mattipee:
@amtriathlon @ericchristoffersen 

Please find here a minimal PR to correct the fundamentals of the calculations in the Fortius device, specifically around speed/force and power.

It is a foundation for other work, but stands alone. I currently have no intention to include further changes in this PR, subject to review.

I have ridden in both ERGO and slope modes, satisfactory to me (at higher wheel speeds, to be addressed in another PR).

Regards,
Matt.

### DIFF
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -582,10 +582,11 @@ int Fortius::sendRunCommand(int16_t pedalSensor)
 		//qDebug() << "send load " << load;
 		
         // Set trainer resistance to the force required to maintain requested power at current speed
-        const double  targetNewtons  = load / kph_to_ms(deviceSpeed);
-        const int16_t targetRawForce = clip_double_to_type<int16_t>(N_to_rawForce(targetNewtons));
+        const double  targetNewtons   = load / kph_to_ms(deviceSpeed);
+        const int16_t targetRawForce  = clip_double_to_type<int16_t>(N_to_rawForce(targetNewtons));
+        const int16_t limitedRawForce = rawForce_FortiusANT_AvoidCycleOfDeath(targetRawForce, deviceSpeed);
 
-        qToLittleEndian<int16_t>(targetRawForce, &ERGO_Command[4]);
+        qToLittleEndian<int16_t>(limitedRawForce, &ERGO_Command[4]);
         ERGO_Command[6] = pedalSensor;
         
         qToLittleEndian<int16_t>((int16_t)(130 * brakeCalibrationFactor + 1040), &ERGO_Command[10]);
@@ -597,13 +598,11 @@ int Fortius::sendRunCommand(int16_t pedalSensor)
         // Set trainer resistance to the force required to maintain current speed at current grade
         // The trainer's virtual flywheel itself will limit acceleration accordingly
 
-        // TODO.1 Fortius has issues with high force at low speed, to be addressed in subsequent PR
-        // TODO.2 expectation is that another PR will provide targetNewtons based upon intertial simulation
+        const double  targetNewtons   = NewtonsForV(kph_to_ms(deviceSpeed));
+        const int16_t targetRawForce  = clip_double_to_type<int16_t>(N_to_rawForce(targetNewtons));
+        const int16_t limitedRawForce = rawForce_FortiusANT_AvoidCycleOfDeath(targetRawForce, deviceSpeed);
 
-        const double  targetNewtons  = NewtonsForV(kph_to_ms(deviceSpeed));
-        const int16_t targetRawForce = clip_double_to_type<int16_t>(N_to_rawForce(targetNewtons));
-
-        qToLittleEndian<int16_t>(targetRawForce, &SLOPE_Command[4]);
+        qToLittleEndian<int16_t>(limitedRawForce, &SLOPE_Command[4]);
         SLOPE_Command[6] = pedalSensor;
         SLOPE_Command[9] = (unsigned int)weight;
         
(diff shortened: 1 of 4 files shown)


amtriathlon:
It looks good for me, if it is based on documentation (or code) from https://github.com/WouterJD/FortiusANT we should credit them.

mattipee:
Apologies for late change, @amtriathlon .

I noticed mistaken use of uint16_t. While there was no adverse behaviour, I have corrected it, and ensured values are clipped to within range.

src/Train/Fortius.cpp
```diff
@@ -646,3 +660,23 @@ bool Fortius::discover(QString)
...
+
+// Calculate the force in Newtons required to sustain speed at gradient
+// Units: force (N), gradient (%), speed (m/s), mass (kg)
+double Fortius::NewtonsForV(double gradient, double m, double v)
+{
+    // TODO.3 substitute values from trainer interface for some of the constants defined here
+    static const double g               = 9.81;
+    static const double Crr             = 0.004;
+    static const double CdA             = 0.51;
+    static const double v_wind          = 0.0;
```
michelrdagenais, line 672: All these values (WindSpeed, WindResistance, RollingResistance) are now set by the RealtimeController, because they were required for the Kickr Core. It is good to supply default values in member fields at construction time, and it would be preferable to use the values set by the RealtimeController interface.

src/Train/Fortius.h
```diff
@@ -164,6 +165,33 @@ class Fortius : public QThread
     // raw device utils
     int rawWrite(uint8_t *bytes, int size); // unix!!
     int rawRead(uint8_t *bytes, int size); // unix!!
+
+
+    // Unit conversion routines
+    static inline double kph_to_ms      (double kph) { return kph / 3.6; }
+    static inline double ms_to_kph      (double ms)  { return ms  * 3.6; }
```
michelrdagenais, line 172: There are about 25 places in the Train directory where this conversion is made. It would be nice to put it in a more reusable place.

michelrdagenais (commented):
The patch looks very good to me, with two minor comments

src/Train/Fortius.h
amtriathlon: This would be a good place: https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/Core/Units.*

src/Train/Fortius.cpp
mattipee: I'm equally happy to expand the scope of this of PR to wire these values through, or to do so in a subsequent PR. I had intentionally limited the scope here to as small a change as possible.

@amtriathlon do you have a preference?

amtriathlon: Since the calculation of force required to meet the gradient target is new, it is not a previously existing functionality, I agree with Michel it would be better to use existing parameters instead of fix them.

src/Train/Fortius.h
mattipee: Do I need to do anything here? Feels like a separate PR to refactor several places to a common constant.

src/Train/Fortius.cpp
mattipee: OK. I will add them.

src/Train/Fortius.h
amtriathlon: No, lets keep this self contained.

src/Train/Fortius.cpp
mattipee: There's a commit now to address this. There's a double mutex lock, but a simple function interface. Please comment.

mattipee: It is worth noting (without wishing to introduce much discussion here), that should Eric's proposed setGradientWithSimState() change be adopted in whatever form, then NewtonsForV() and the wind speed and resistance parameters I have just added could become redundant. Alternatively of course, a configuration option could be added to choose which calculation to use.

That said, the changes in this PR are self-consistent as a first step.

mattipee (commented):
By not addressing this "TODO.1" at all, the trainer will stutter badly when a high force is commanded at low wheel speed.

The correction applied in FortiusANT looks like this:

```
if self.SpeedKmh <= 10 and Resistance >= 6000:
            Resistance = int(1500 + self.SpeedKmh * 300)
```

While it is not perfect, I do not yet have a more perfect proposition.

It may be an idea to include this now. Or not.

Comments, please?

src/Train/Fortius.cpp
amtriathlon: Simulation is currently a user selectable option. The way I see this PR is achieving a similar behavior smart(er) trainers have in firmware and we use when simulation is disabled.

src/Train/Fortius.cpp
```diff
@@ -563,7 +594,16 @@ int Fortius::sendRunCommand(int16_t pedalSensor)
     }
     else if (mode == FT_SSMODE)
     {
-        qToLittleEndian<int16_t>((int16_t)(650 * gradient), &SLOPE_Command[4]);
+        // Set trainer resistance to the force required to maintain current speed at current grade
+        // The trainer's virtual flywheel itself will limit acceleration accordingly
+
+        // TODO.1 Fortius has issues with high force at low speed, to be addressed in subsequent PR
```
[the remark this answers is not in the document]
amtriathlon: Yes, I think it is a good idea.

src/Train/Fortius.cpp
michelrdagenais: > There's a commit now to address this. There's a double mutex lock, but a simple function interface. Please comment.
Either way is fine with me. I am not too worried about the performance cost of the lock. Un uncontended lock on Linux, because of the implementation with futex, stays in userspace and is a single atomic operation.

ericchristoffersen: > Simulation is currently a user selectable option. The way I see this PR is achieving a similar behavior smart(er) trainers have in firmware and we use when simulation is disabled.

The sim always runs, its use for progress is optional.

amtriathlon: I am talking about the user POV, when the check is disabled simulation does nothing useful for the user with any trainer, AFAIK.

ericchristoffersen: > I am talking about the user POV, when the check is disabled simulation does nothing useful for the user with any trainer, AFAIK.

We are discussing whether it makes sense to have fortius implement its own sim. Whether user enables sim for progress is besides the point, the sim value is currently always available in trainsidebar and available to use. We can go down the path of every device implementing its own sim or we can have 'sim-less' devices share the common value. Look at the monark pr for another case exactly like fortius.

amtriathlon: Honestly I have zero interest to start that discussion again, I understand your point from the other thread, and I don’t agree..

src/Train/Fortius.cpp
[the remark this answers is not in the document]
mattipee: There is a commit now to address this. The FortiusANT routine has been used and attributed. This addressed the "worst" of the stuttering, though as I said, it's not perfect.

I think this was indeed worth including in this first PR. I hope it can be improved later.

Any comments? Otherwise, feel free to resolve this conversation.

amtriathlon:
Thanks for your contribution Matt, do you like to give us your full name to be included in the contributors list?

mattipee:
You're welcome - and thanks to all involved leading up to this first one.

I may have typed the code, but the formulae contributed were not mine.

amtriathlon:
Perfect, then thank to all. Eric and Michel are already included as contributors in https://github.com/GoldenCheetah/GoldenCheetah/blob/a298715bb503884964e8030d8ac0824dc765d0e4/src/Gui/AboutDialog.cpp#L113, but you aren’t, yet.

PR #3790, opened 2021-01-19
Fortius calibration

mattipee:
@amtriathlon @ericchristoffersen @michelrdagenais

A round of changes to introduce Fortius calibration.

There are a few aspects of these changes where I'm not wholly convinced about my approach, and therefore I look forward to comments.

### DIFF
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -17,7 +17,7 @@
  */
 
 #include "Fortius.h"
-#include <Core/Settings.h>
+#include "Settings.h"
 
 //
 // Outbound control message has the format:
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -105,7 +105,7 @@ void Fortius::setMode(int mode)
     pvars.unlock();
 }
 
-// TODO description
+// Compensate for trainer friction (at 20 kph)
 void Fortius::setBrakeCalibrationForce(double force_N)
 {
     // save raw calibration value
@@ -579,6 +579,7 @@ void Fortius::run()
  * sendOpenCommand() - initialises training session
  * sendCloseCommand() - finalises training session
  * sendRunCommand(int) - update brake setpoint
+ * sendCalibrateCommand() - start calibration at 20 kph
  *
  * ---------------------------------------------------------------------- */
 int Fortius::sendOpenCommand()
@@ -596,6 +597,15 @@ int Fortius::sendCloseCommand()
     uint8_t close_command[] = {0x01,0x08,0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x52,0x10,0x04};
     
     int retCode = rawWrite(close_command, 12);
+	//qDebug() << "usb status " << retCode;
+	return retCode;
+}
+
+int Fortius::sendCalibrateCommand()
+{
+    uint8_t calibrate_command[] = {0x01,0x08,0x01,0x00,0xa3,0x16,0x00,0x00,0x03,0x00,0x00,0x00};
+    
+    int retCode = rawWrite(calibrate_command, 12);
 	//qDebug() << "usb status " << retCode;
 	return retCode;
 }
@@ -652,9 +662,7 @@ int Fortius::sendRunCommand(int16_t pedalSensor)
     }
     else if (mode == FT_CALIBRATE)
     {
-        // Not yet implemented, easy enough to start calibration but appears that the calibration factor needs
-        // to be calculated by observing the brake power and speed after calibration starts (i.e. it's not returned
-        // by the brake).
+        retCode = sendCalibrateCommand();
     }
 
 	//qDebug() << "usb status " << retCode;
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -67,11 +67,7 @@ Fortius::Fortius(QObject *parent) : QThread(parent)
     windSpeed = DEFAULT_WINDSPEED;
     rollingResistance = DEFAULT_Crr;
     windResistance = DEFAULT_CdA;
-
-    // Lead raw calibration value, and convert to N
-    const double raw_saved_calibration = appsettings->value(this, TRAIN_FORTIUSCALIBRATION, DEFAULT_CALIBRATION_FORCE_N).toInt();
-    brakeCalibrationForce_N = rawForce_to_N(raw_saved_calibration);
-
+    brakeCalibrationForce_N = appsettings->value(this, TRAIN_FORTIUSCALIBRATION, DEFAULT_CALIBRATION_FORCE_N).toDouble();
     brakeCalibrationFactor = DEFAULT_CALIBRATION_FACTOR;
     powerScaleFactor = DEFAULT_SCALING;
     deviceStatus=0;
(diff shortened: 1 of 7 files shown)


src/Train/Fortius.cpp
```diff
@@ -62,7 +67,12 @@ Fortius::Fortius(QObject *parent) : QThread(parent)
     windSpeed = DEFAULT_WINDSPEED;
     rollingResistance = DEFAULT_Crr;
     windResistance = DEFAULT_CdA;
-    brakeCalibrationFactor = DEFAULT_CALIBRATION;
+
+    // Lead raw calibration value, and convert to N
+    const double raw_saved_calibration = appsettings->value(this, TRAIN_FORTIUSCALIBRATION, DEFAULT_CALIBRATION_FORCE_N).toInt();
```
michelrdagenais, line 72: Do I understand correctly, TRAIN_FORTIUSCALIBRATION and raw_saved_calibration are raw forces units, as DEFAULT_CALIBRATION_FORCE_N is in Newtons?!

src/Train/FortiusController.cpp
```diff
@@ -179,3 +179,164 @@ FortiusController::setWindResistance(double windResistance)
...
+    {
+        // Waiting for user to kick pedal...
+        case CALIBRATION_STATE_REQUESTED:
+        {
+            int Buttons, Status, Steering;
+            double Power, Force_N, HeartRate, Cadence, SpeedKmh, Distance;
+            myFortius->getTelemetry(Power, Force_N, HeartRate, Cadence, SpeedKmh, Distance, Buttons, Steering, Status);
+
+            // Once we're up to speed, move on to next stage
+            if (SpeedKmh > 19.9) // up to speed
```
michelrdagenais, line 237: The 20kph in getCalibrationTargetSpeed could be defined (e.g. CALIBRATION_SPEED) and here you could have the following instead of a magical number like 19.9:

0.995 * CALIBRATION_SPEED // Within 0.005 of the target speed

or simply

0.995 * getCalibrationTargetSpeed() // Within 0.005 of the target speed

src/Train/FortiusController.h
```diff
@@ -25,6 +25,56 @@
 #ifndef _GC_FortiusController_h
 #define _GC_FortiusController_h 1
 
+template <size_t N>
+class NSampleSmoothing
```
michelrdagenais, line 29: If you search for rollling in the source tree, you will find about 20 places in GC where a rolling average is implemented. This class seems fairly generic. It could be placed in a core file and used initially by FortiusController. In parallel, I can raise an issue about starting to use it elsewhere.

src/Train/FortiusController.cpp
mattipee: Done, thanks.

src/Train/Fortius.cpp
mattipee: Silly oversight. Thanks... I've updated, re-review please?

src/Train/FortiusController.h
mattipee: I don't know if using a threshold on the stddev is actually the best way to do this. This class, where it is, is a bit of a thorn in this PR. The FortiusANT approach is to compare min/max of an array of rolling averages... using stddev was an attempt to improve on that slightly.

Agree that this code should be elsewhere. As an aside, I think ERGO mode should calculate trainer force based on a smoothed (rather than instantaneous) value for deviceSpeed. That's a tangent to this PR, but related, and worth some later thought.

src/Train/Fortius.cpp
```diff
@@ -18,6 +18,11 @@
 
 #include "Fortius.h"
 
+#include <Core/Settings.h>
+#define TRAIN_FORTIUSCALIBRATION "<global-trainmode>train/fortiuscalibration"
+// could (should?) be defined in Core/Settings.h
+// but they are not currently used anywhere other than Fortius.cpp
```
amtriathlon, line 24: Yes please, even when they are used only in one place we define them there to have a single registry for settings.

src/Train/FortiusController.h
Mickyduck55: Hi tried this today and a couple of points

Warmed up for 15 mins paused ride until wheel stopped, pressed play then calibrate, Kicked pedal Message came up stop pedalling but wheel did not turn.. went through process 60 seconds etc calibration returned as Zero... as no calibration took place.
Tried again all OK calibration of 52 @ 20kph 
Tried again and all Ok

So seems fine thanks

In real life riding this is clunky IF I am cycling with a video,  warm up 15 mins stop pedalling move mouse to calibrate click .... trainier stops dead then instantaneously starts calibrating... not a pleasant sensation.

I don't think the calibration is the problem its the train view buttons.

So if I stop peddling allow wheel to stop to avoid the clunk the data for my ride is all wrong because all the time I am stopping the wheel the dials are spinning, averages decreasing ......  If I press STOP my ride is over then I have to save, press play, calibrate then skip forward until I get to where I was before or even worse ride the first 15 mins again.
For this to work much more smoothly it would be really ideal IMO if the calibrate button was active when PAUSE was pressed, no saving ride skipping back to where you were or incorrect data because you are coasting to a stop before you can press calibrate
Susan has a i Magic the calibration for that would work perfectly with this existing scenario as it calibrates by cycling up to a speed then stop pedalling... so in the ride just click calibrate speed up to required KPH wait,  when complete resume cycling video in same place dials start spinning..... no issues..... Maybe it is just the Fortius... I guess this would need to go into the issues section rather than in @mattipee  PR?

mattipee: @Mickyduck55 - I did briefly look at the power/start/pause states over new year, and how they tie in with the Fortius commands. I agree in principle that Fortius calibration while paused would be more natural. Requires further thought, and a subsequent PR I think.

src/Train/Fortius.cpp
mattipee: Moved, also put it in [fortius] section alongside fortius/firmware, rather than under [train].

src/Train/FortiusController.h
Mickyduck55: Or if possible as this will not rely on changes to the train view controls?
Ideally warm up 15 mins
Click calibrate while riding,
when calibration pressed "Stop Pedalling" Message,
calibration process adjusts the speed to 20 kph up / down depending on speed at "click", Calibration takes place
if no speed then message to "pedal" displayed.

mattipee: @Mickyduck55
> Kicked pedal Message came up stop pedalling but wheel did not turn.. went through process 60 seconds etc calibration returned as Zero... as no calibration took place.

Odd... I've added low-wheel-speed as a termination condition for calibration so this shouldn't happen again.

Mickyduck55:
Thanks.. is rolling calibration a possibility or does the brake have to
stop before calibration can be done?

src/Train/Fortius.cpp
```diff
@@ -17,6 +17,7 @@
  */
 
 #include "Fortius.h"
+#include <Core/Settings.h>
```
amtriathlon, line 20: Please use #include "Settings.h", the search path will take care of the location, use the angle form only for standard includes.

michelrdagenais:
I tried it on my Fortius and it seems to work fine, saving the calibration from one execution of GC to the next.

src/Train/Fortius.cpp
mattipee: Done

mattipee:
@Mickyduck55 

> Thanks.. is rolling calibration a possibility or does the brake have to stop before calibration can be done?

Rolling calibration should be possible by matching (>20 kph) device speed in the calibration commands and gradually reducing to 20 kph, before commencing the sample capture. I got so far with the idea this evening, but don't intend to include it in this PR.


@michelrdagenais thank you for the test. :+1:

Mickyduck55:
Thanks Matt if it is possible that would be so much more user friendly ....

mattipee:
Thank you, all. :)

PR #3793, opened 2021-01-20
Zoomed Elevation widget

human705:
A widget that will show 300 meters of the ride at a time. The cyclist position is displayed at 100 meters as a red line. To the left of the red line, a bubble is displayed with the grade at the current cyclist position. The bubble changes size according to the grade and will touch the red line and the elevation line. The bubble also changes colors according to the grade. It transitions from black to white from -40 to 0, white to yellow from 0 to 3, yellow to orange from 3 to 8 and finally orange to red from 8 to 40. Here are a couple of screenshots with graphical meters and text meters.

image

image

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -195,6 +195,7 @@
 #define GC_DPRP_EQUIPWEIGHT             "<global-general>dataprocess/fixrunningpower/equipwheight"
 #define GC_DPDR_DRAFTM                  "<global-general>dataprocess/fixrunningpower/draftm"
 #define GC_DPFV_MA                      "<global-general>dataprocess/fixspeed/ma"
+#define GC_DPFE_AK                      "<global-general>dataprocess/fixelevation/ak"
 #define GC_CAD2SMO2                     "<global-general>dataprocess/fixmoxy/cad2smo2"
 #define GC_SPD2THB		            	"<global-general>dataprocess/fixmoxy/spd2thb"
 #define GC_DPFLS_PL                     "<global-general>dataprocess/fixlapswim/pool_length"
--- a/src/FileIO/FixElevation.cpp
+++ b/src/FileIO/FixElevation.cpp
@@ -43,31 +43,54 @@ struct elevationGPSPoint {
     double lat;
 };
 
-QStringList FetchElevationDataFromMapQuest(QString latLngCollection);
-
 // Config widget used by the Preferences/Options config panes
 class FixElevation;
 class FixElevationConfig : public DataProcessorConfig
 {
     Q_DECLARE_TR_FUNCTIONS(FixElevationConfig)
+
     friend class ::FixElevation;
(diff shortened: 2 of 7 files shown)


peret2000:
This is a terrific widget. It works fine. the code is using nice tools. Specially, the colour mapping template results in a goot tool to be reused in many other parts of the code, I guess, as the effect is impressive.
Congratulations, and thanks for the job.

I have been testing it and thinking of ways to make it even more useful (from my point of view).

I found that the zoomed graph can be enough to forecast the slope, but sometimes it lacks a numeric perception of the change. Viewing current slope in the bubble is good, but it is the 'present', something you are also viewing in meter widget, or in a meter field in training screen. I thought that it could be more interesting some kind of view of the inmediate future; that, combined with the graphical image could be powerful.
After several tries with different approachs (show instant slope 100 meters ahead, average slope for those next 100 meters, etc), I found a criteria that is representative of the future, and independent of the speed (distance perception is not the same when climbing at 10 kmh than descending at 50kmh):
The bubble can show the icrement/decrement of slope withing the next, say, 10 seconds: depending on current speed, that will be average slope for a different distance, compared with current instantaneous slope. Bubble colours should be adapted, of course, to a different range of values.
I know it is not the most consistent, as it is comparing an average slope with an instant slope, but it is the best approach I have come up with.
It also needs some brain training to well understand its behavior, but, to me, it is very useful.
Just for curiosity, [here](https://github.com/peret2000/GoldenCheetah/commit/4df80f6509766dc3384e20eac0052d18aba18d16) I have done changes to test it; not thought as a PR (it contains other small changes to adapt my tests, but to share the idea.

Thanks, Peter, for this relevant contribution, I missed this functionality from the first day I started using GC.

human705:
I have been thinking of ways to make this better and one of this is close to what I think you are describing. Rather than having the bubble, i was thinking of changing the color of sections of the polygon as it changes slope. Unfortunately, that may require to re-draw the polygon every time. In addition,  the slope itself may need to be "smoothed" to avoid a polygon with too many colors.

peret2000:
> I have been thinking of ways to make this better and one of this is close to what I think you are describing. Rather than having the bubble, i was thinking of changing the color of sections of the polygon as it changes slope. Unfortunately, that may require to re-draw the polygon every time. In addition, the slope itself may need to be "smoothed" to avoid a polygon with too many colors.

That sounds good. Regarding smoothing, it will depend on the circuit itself. I usually smoothen gpx files using GC functionality of editing GPS and altitude. Also, I guess the range of colours are not too wide, so if slope changes quickly but with small changes, colour may not change very often.

Regarding the redraw, I do not know where the problem is: I think you have to re-draw it continuously to update the profile as you are riding. So, painting with specific or fixed colours should not be an extra condition to redraw, should it?

amtriathlon:
Thank you!

PR #3732, opened 2020-12-22
LibUsb::OpenFortius - Remove call to usb_set_altinterface() 

mattipee:
Was known bad on OSX, proved an issue on LINUX (for me), and to be tested on WINDOWS following its removal.

I believe creating this PR will produce an artefact that is testable.

PR #3795, opened 2021-01-21
Fortius USB timing

mattipee:
@michelrdagenais @amtriathlon 

The Fortius will echo back the commanded resistance in bytes [40:41] of the frames returned.

I find that with the current timing (send + read + **sleep**), the echo'd resistance is 2 frames out of sync...

  sent **4383**, received 4722
  sent 4299, received 4538
  sent 4273, received **4383**
  sent **4216**, received 4299
  sent 4278, received 4273
  sent 4360, received **4216**

By moving the sleep as in this PR (send + **sleep** + read), the echo'd resistance becomes only 1 frame out of sync...

  sent **3601**, received 3516
  sent 3685, received **3601**
  sent 3744, received 3685
  sent **3710**, received 3744
  sent 3687, received **3710**
  sent 3592, received 3687

### DIFF
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -408,6 +408,10 @@ void Fortius::run()
         		return; // couldn't write to the device
 			}
 			
+            // The controller updates faster than the brake. Setting this to a low value (<50ms) increases the frequency of controller
+            // only packages (24byte). Tacx software uses 100ms.
+            msleep(50);
+
             int actualLength = readMessage();
 			if (actualLength < 0) {
 				qDebug() << "usb read error " << actualLength;
@@ -531,11 +535,6 @@ void Fortius::run()
                         
             timer.restart();
         }
-
-        
-        // The controller updates faster than the brake. Setting this to a low value (<50ms) increases the frequency of controller
-        // only packages (24byte). Tacx software uses 100ms.
-        msleep(50);
     }
 }
 


amtriathlon:
It looks like a reasonable change to me, and I would suggest to try 100ms delay as the comment says it does Tacx software, but I have no way to test this.

mattipee:
I tried this the other night... 100ms appeared to make no difference for me and the echo'd resistance was still one frame out-of-sync. The benefit of a tight loop is more responsive buttons on the head-unit.

I believe the head unit _should_ respond with short frames if you poll faster than the motor controller is able to provide content for the long frames, but I see a full frame even at 50ms - perhaps the current resistance and speed values are always unique, frame to frame. I will check.

@michelrdagenais have you any further thoughts on this? I also tried inserting a sleep+read after the initial start message, again no apparent change in behaviour.

amtriathlon:
Let me know what you like to do once both of you completed your tests, we also have https://github.com/GoldenCheetah/GoldenCheetah/pull/3732 pending I would like to close, may be that commit can be added here so they are tested together on Linux and we could ask for some testing on Windows now Fortius support is more complete.

michelrdagenais:
I will try a few things tonight. There are some things to look at. First, Fortius::run() does a SendOpenCommand, does not read the expected Version Message answer, and follows with a SendRunCommand. Then, LibUsb.cpp does buffering. Therefore, the first answer read by the loop may get the Version message, instead of correct power/speed measurements. However, since we read up to 64 bytes each time, while the messages are 48 bytes, these unread bytes should disappear fairly quickly and their effect is probably not noticeable. I will look at the effect of changing the delays, and the latency to receive the force echo.

mattipee:
@michelrdagenais 

I have found that simply performing readMessage() twice in the main loop appears to yield the behaviour I was looking for.

- the first response is always "old" wrt the echo'd resistance
- even increasing the sleep to 200ms makes no difference to that
- the second echo'd resistance appears to always be correct wrt the echo'd resistnace
- the reported speed/force values change between first/second responses, so "more recent"
- I still get full frames @ ~18Hz

If I also reduce the sleep to 25ms, I get ~32Hz with only _very occasional_ repetition of speed/resistance values. (I understand there's a limit to the frequency with which the motor controller can update the head unit over serial). But I am also curious why I only ever appear to see full frames, and never the shorter "controller only" 24 byte frames.

Interested to hear if you've made any progress yourself, and obviously your opinions on this. @switchabl - I'd really appreciate your input too, please :)


```
             // only packages (24byte). Tacx software uses 100ms.
-             msleep(50);
+             msleep(25);
 
+            readMessage(); // discard
             int actualLength = readMessage();
                        if (actualLength < 0) {
                                qDebug() << "usb read error " << actualLength;

```

michelrdagenais:
I tried quite a few things to understand how this is working. Every time you read from the Fortius you get an answer, even if you dont send a command. The head unit is returning its 24 bytes. It takes about 0.1ms for a single read but around 10ms for consecutive reads. When you send a run command, a longer answer will become available once. I was reading 20 times in a loop and the longer answer would come only once, at the 6th read, about 50ms after the run command was sent.

With a 50ms sleep between the run command and the readMessage, the correct long answer is obtained about 50% of the time. With a 60ms sleep, the long answer is obtained almost 100% of the time. I would suggest using a 60ms sleep. The whole cycle is not synchronized anyway. The whole loop with a 50ms sleep takes about 51ms anyway.

With a 50 or 60ms sleep between the run command and the readMessage, the force echo is delayed by one cycle. It seems better than putting the sleep at the end, as currently. However, in practice, it probably does not make much difference. Look at the following timeline. You get the same values about at the same time. Still, I believe that @mattipee is right and it is more logical to put the sleep(60) between the send run and the readMessage.

 0  send run
 1 readMessage (not ready)
 2 sleep 50
52 send run
53 readMessage (from send run in 0)
54 sleep 50
104 send run
105 readMessage (from send run in 52)
...

 0  send run
 1 sleep 50
51 readMessage (from send run in 0)
52 send run
53 sleep 50
103 readMessage (from send run in 52)
...

mattipee:
Hmmm... initial read of your description suggests my "25ms+double read" shouldn't be possible. Are you in a position to test my patch above?

Is readMessage()/rawRead() only ever returning either 24 or 48 for you? It returns 64 for me every time. I'm running Debian 10.7. Also, I have no idea if there's any difference between the T1932 and T1942 in its USB handling.

The practical difference moving the sleep is that the "send run" is calculating the values to send as a function of a more recent "readMessage".

OK, so default position will be to increase 50 to 60, pending any comment you have on my patch above?

Thanks for your time and effort!

michelrdagenais:
I did again some testing yesterday evening. My unit is always returning either 24 or 48 bytes. I an not getting the same results as you with your code. The force echo is still delayed, despite the extra readMessage. However, if I put 5 extra readMessage, sometimes the force echo is not delayed. The behavior seems more complicated than I tought. I will make some more tests.

I looked again at the https://github.com/totalreverse/ttyT1941/wiki documentation and they document that the T1932 head unit always returns 64 bytes. However, their study does not really cover the returned bytes:

T1941[24+23+1]: 03 13 02 00 00 00 00 00 00 00 f0 61 00 00 00 00 80 06 00 00 00
T1932[24+40]: TODO

I think that this additional tuning (echo / timing) can be delayed to another PR. We should already be able to store in the Fortius class the type of head unit that we have (different USB id). Moreover, I have a patch to also get, display and store the brake type and firmware version (to send when your work on the Fortius has stabilised). We will thus be able to introduce some type/version specific tweaks, since the different units behave differently in some ways.

mattipee:
Thanks, @michelrdagenais .

OK, I think the summary here is the following... do you agree?

- merge this PR now, as we agree on the principle of moving the sleep
- then next for you to please PR your version/type code
- leave to future PR any further timing improvements.

michelrdagenais:
Yes. This is a good plan. Moving the sleep is sensible with what we know, and this is What FortiusANT does, for instance send command to ask for version, sleep 100ms, read version reply. Once this is merged, I will update and submit my version reading patch.

Issue #3074, opened 2019-04-11
Activity list row height is not updated when data changes


batsonjay:
Present in 3.5DEV build 1903, MacOS (though the regression appeared in 1901).

The following is a regression. I can reproduce this reliably:
1) Import a new activity from file (in this case, Zwift data)
2) Select Activities -> Details -> Workout; enter Sport
3) Select Activities -> Details -> Notes; enter Note
4) Select Activities -> Summary

The newly-entered event details _do not_ immediately appear under the event in the Activities pane (Upper-left corner). This is a regression; previously a mere pane switch (to anything) updated the display. (In fact, it previously updated the event with the Workout info when you switched from the Workout to Notes tab.)

The newly entered details DO appear after exit & restart of application.  They also DO appear in other places where "Calendar Text" is used.

See this screencast to see this behavior: https://screencast-o-matic.com/watch/cqfXIhZDeF

If multiple rides are imported from file simultaneously, this problem only appears in the FIRST file to be imported (which tends to be the oldest).  Details entered in the 2nd, 3rd, ... imported files appear as expected.

amtriathlon:
The only scenario when I can reproduce this is: activity is the first in Activities list AND calendar text is empty so the additional line is not present and it is not created when calendar text becomes non-empty.
In my typical setup this didn't happened since I have other fields additional to Notes which are present at the time of activity creation, the second line is present and adding Notes behaves as expected.

liversedge:
Ah. So this is not a regression. Its about the height of the row not being recomputed when data changes. Will look at that.

batsonjay:
I just realized I mis-stated the behavior. When multiple rides are imported notes are not displayed only for the ride that is at the top of the list, which is the NEWEST ride, not the oldest (as I state in the initial bug report).

@Mark, this does indeed feel like a row height issue because of another behavior I have seen that I failed to put in the report. If the list is sorted inversely to what you see in that screencast (with recent events at the bottom), a newly added row at the bottom of the list is displayed just a little bit too low the scroll region in this pane, and I can only see the top few pixels of the characters in the row. Also, I cannot scroll down to see that item. Restarting GC makes that (new) bottom row visible.

So check that as well when you’re checking for row height at the top of the list.

MeisterP:
> The newly entered details DO appear after exit & restart of application.

For me, just changing the sort order of the columns makes the text appear.

sanfordhk:
[...]

PR #3798, opened 2021-01-24
Ensure that new activities imported into the RideNavigator are displayed expanded

paulj49457:
Currently when auto importing activities into GC they appear as a single line (date/time) when the other activities are expanded, this change ensures the Ride Navigator expands the new activities.

Note: There maybe a more targetted way to expand just the new field(s), but I spent time trying to use "setExpanded()" & "expand()" but failed, while the sledgehammer "expandAll()" approach works.

### DIFF
--- a/src/Gui/RideNavigator.cpp
+++ b/src/Gui/RideNavigator.cpp
@@ -966,6 +966,7 @@ RideNavigator::setRide(RideItem*rideItem)
                 tableView->scrollTo(tableView->model()->index(j,3,group), QAbstractItemView::PositionAtCenter);
 
                 currentItem = rideItem;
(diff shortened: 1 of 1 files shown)


amtriathlon:
It seems this should fix https://github.com/GoldenCheetah/GoldenCheetah/issues/3074, can you confirm this?

amtriathlon:
Lookin at https://groups.google.com/g/golden-cheetah-developers/c/ennvioNpQ3U, it seems better to read the documentation and try to understand how the program works before to propose changes.

paulj49457:
I agree with closure, as this would have expanded any groups, which unfortunately I wasn't aware of at the time.

PR #3800, opened 2021-01-26
RealtimeController::setWindResistance() should be passed CdARho, not just CdA

mattipee:
Source: ANT+ Fitness Equipment Device Profile, section 8.8.3:
  - Wind Resistance Coefficient - Product of Frontal Surface Area, Drag Coefficient **and Air Density** - 0.00 to 1.86 kg/m

I assume the same is true for the equivalent in BT.

Certainly the way in which Fortius::NewtonsForV() is implemented, this is now correct to the best of my knowledge.

### DIFF
--- a/src/Train/BicycleSim.h
+++ b/src/Train/BicycleSim.h
@@ -168,7 +168,7 @@ public:
     double EquivalentMassKG() const;         // Additional mass due to rotational inertia
     double KEMass() const;                   // Total effective kinetic mass
     double RollingResistance() const { return m_constants.m_crr; }
-    double WindResistance() const { return m_constants.m_Cd * m_constants.m_A; }
+    double WindResistance(double altitude) const { return m_constants.m_Cd * m_constants.m_A * AirDensity(altitude, m_constants.m_T); }
 
     const BicycleWheel &FrontWheel() const { return m_frontWheel; }
     const BicycleWheel &RearWheel() const { return m_rearWheel; }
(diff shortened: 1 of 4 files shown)


amtriathlon:
@michelrdagenais are you ok with this change, can you test it with your Kickr via BTLE?

michelrdagenais:
The BTLE FTMS and presumably the similar Wahoo interface expects a Wind Resistance value in Kg/m. In the formula for the force, 1/2 Cd A Rho v², Cd is dimensionless, A is in m² and Rho is kg/m³. Therefore, the Wind Resistance value expected should be Cd A Rho (kg/m³ x m² = kg/m), as implemented in this PR. I just finished a pull request to fix the Fortius interactions tonight. I will test with my Kickr Core and BTLE tomorrow to see if the updated value reaches the low level routine sending the Wind Resistance, and if this value is in the same range as the default values suggested by Wahoo.

michelrdagenais:
I ran this branch on the Wahoo Kickr and the value provided to the trainer is close to the default value (0.6), and varies with the altitude. It is about 0.6 at sea level and goes down with higher altitude. I am good with this PR.

When I ran it on a .pgmf workout and went fast forward to compare the values at different altitudes, the altitude did not update. For .pgmf files, the altitude is computed incrementally and it is not updated when seeking forward or backward. This is a bug unrelated to this PR. I will enter an issue for this.

mattipee:
Thank you, both.

PR #3801, opened 2021-01-27
Fix the request - reply interactions with the Fortius

michelrdagenais:
After a lot of experimentation, I now have something that seems to work well in order to get proper replies to commands sent to the Fortius. The open command gets the version correctly. The run command gets the proper force echo in the same request reply cycle. @mattipee please let me know if that works well for you (getting the correct version number and the force echo in the same cycle).

The open command is in fact to get a reply with the version. However, some already queued messages need to be skipped, before the well formed reply message is received. The open command is sent initially, and each time the device is disconnected and reconnected. The procedure to open the device and get the version was then put in a separate function.

When the run command is sent with a specified force, a reply comes back with the force echoed. However, here again, some messages may need to be skipped before the correct reply comes back.

### DIFF
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -447,6 +447,9 @@ void Fortius::run()
             // If you simply reissue the run command when 24 bytes are
             // received, the following 48 bytes reply will match the force echo
             // of the before last run command.
+            //
+            // On the T1932, replies are always 64 bytes. The best strategy seems to
+            // be simply skipping the first reply.
 
             msleep(50);
             for(int i = 0; i < 5; i++) {
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -69,6 +69,14 @@ Fortius::Fortius(QObject *parent) : QThread(parent)
     deviceStatus=0;
     this->parent = parent;
 
+    motorBrakeFirmwareVersion = 0;
+    motorBrakeRawSerialNumber = 0;
+    motorBrakeVersion = 0;
+    motorBrakeType = 0;
+    motorBrakeYear = 0;
+    motorBrakeSerialNumber = 0;
+    motorBrake = false;
+
     /* 12 byte control sequence, composed of 8 command packets
      * where the last packet sets the load. The first byte
      * is a CRC for the value being issued (e.g. Load in WATTS)
@@ -386,6 +394,7 @@ void Fortius::run()
     int curSteering;                    // Angle of steering controller
     // UNUSED int curStatus;
     uint8_t pedalSensor;                // 1 when using is cycling else 0, fed back to brake although appears unnecessary
+    int actualLength;
 
     // we need to average out power for the last second
     // since we get updates every 10ms (100hz)
@@ -408,15 +417,8 @@ void Fortius::run()
     pedalSensor = 0;
     pvars.unlock();
 
-
-    // open the device
-    if (openPort()) {
-        quit(2);
-        return; // open failed!
-    } else {
-        isDeviceOpen = true;
-        sendOpenCommand();
-    }
+    if(openDevice()) return;
+    isDeviceOpen = true;
 
     QTime timer;
     timer.start();
@@ -434,14 +436,28 @@ void Fortius::run()
         		return; // couldn't write to the device
 			}
 			
-            // The controller updates faster than the brake. Setting this to a low value (<50ms) increases the frequency of controller
-            // only packages (24byte). Tacx software uses 100ms.
-            msleep(50);
+            // On the T1942, the controller head returns 24 bytes for each read,
+            // except 48 bytes once as reply to a run command.
+            //
+            // It takes about 50ms for the 48 bytes reply to be ready.
+            // The force on the run command is echoed in the reply.
+            // You need to discard one or two 24 bytes replies, before getting
+            // the 48 bytes reply with matching force echo.
+            //
+            // If you simply reissue the run command when 24 bytes are
+            // received, the following 48 bytes reply will match the force echo
+            // of the before last run command.
 
-            int actualLength = readMessage();
-			if (actualLength < 0) {
-				qDebug() << "usb read error " << actualLength;
-			}
+            msleep(50);
+            for(int i = 0; i < 5; i++) {
+                actualLength = readMessage();
+                if (actualLength < 0) {
+	            qDebug() << "usb read error " << actualLength;
+	        }
+                // qDebug() << "Read reply " << QByteArray((const char *)buf, actualLength).toHex(':');
+                if(actualLength >= 48) break;
+            }
+
             if (actualLength >= 24) {
 
                 //----------------------------------------------------------------
@@ -553,16 +569,66 @@ void Fortius::run()
 
         } else if (!(curstatus&FT_PAUSED) && (curstatus&FT_RUNNING) && isDeviceOpen == false) {
 
-            if (openPort()) {
-                quit(2);
-                return; // open failed!
-            }
+            if(openDevice()) return;
             isDeviceOpen = true;        
-            sendOpenCommand();
                         
             timer.restart();
         }
     }
+}
+
+int Fortius::openDevice()
+{
+    const int SEND_RETRY = 3;
+    const int READ_REPLY_RETRY = 5;
+    int i = 0, j = 0;
+    int actualLength = 0;
+
+    // open the device
+    if (openPort()) {
+        quit(2);
+        return 1; // open failed!
+    }
+
+    // we need to flush queued packets, and retry a few times
+    actualLength = readMessage();  // skip
+    actualLength = readMessage();  // skip
+
+    for(i = 0; i < SEND_RETRY; i++) {
+        sendOpenCommand();
+        msleep(50);
+        for(j = 0; j < READ_REPLY_RETRY; j++) {
+            actualLength = readMessage();
+            // qDebug() << "Read version, " << QByteArray((const char *)buf, actualLength).toHex(':');
+
+            // Check that this is a proper version reply
+            if (actualLength >= 40 && buf[24] == 0x03 && buf[25] == 0x0c && buf[26] == 0 && buf[27] == 0) break;
+            if (actualLength < 0) {
+                qDebug() << "usb read error " << actualLength;
+            }
+        }
+        if(j < READ_REPLY_RETRY) break;
+    }
+
+    // Decode the proper version reply obtained
+    if (i < SEND_RETRY) {
+        motorBrakeFirmwareVersion = qFromLittleEndian<quint32>(&buf[28]);
+        motorBrakeRawSerialNumber = qFromLittleEndian<quint32>(&buf[32]);
+        motorBrakeVersion = qFromLittleEndian<quint16>(&buf[36]);
+        motorBrakeType = motorBrakeRawSerialNumber / 10000000;
+        // T1941 (230V, type 41) and T1946 (110V, type 46) are motor brake
+        motorBrake = motorBrakeType == 41 || motorBrakeType == 46;
+        motorBrakeYear = (motorBrakeRawSerialNumber % 10000000) / 100000;
+        motorBrakeSerialNumber = motorBrakeRawSerialNumber % 100000;
+
+        QString version = QString(tr("Motor Brake Unit Firmware=%1 Serial=%2\n    year=%3 type=T19%4\n    Version2=%5 MotorBrake=%6"))
+            .arg(QString::number(motorBrakeFirmwareVersion), QString::number(motorBrakeSerialNumber),
+            QString::number(motorBrakeYear + 2000), QString::number(motorBrakeType),
+            QString::number(motorBrakeVersion), QString::number(motorBrake));
+        qDebug() << version;
+        emit dynamic_cast<RealtimeController*>(parent)->setNotification(version, 0);
+    }
+    return 0;
 }
 
 /* ----------------------------------------------------------------------
@@ -580,6 +646,7 @@ int Fortius::sendOpenCommand()
     uint8_t open_command[] = {0x02,0x00,0x00,0x00};
     
     int retCode = rawWrite(open_command, 4);
+    // qDebug() << "Open command " << QByteArray((const char *)open_command, 4).toHex(':');
 	//qDebug() << "usb status " << retCode;
 	return retCode;
 }
@@ -629,6 +696,7 @@ int Fortius::sendRunCommand(int16_t pedalSensor)
         qToLittleEndian<int16_t>((int16_t)(130 * brakeCalibrationFactor + N_to_rawForce(brakeCalibrationForce_N)), &ERGO_Command[10]);
                 
         retCode = rawWrite(ERGO_Command, 12);
+        // qDebug() << "Ergo command " << QByteArray((const char *)ERGO_Command, 12).toHex(':');
     }
     else if (mode == FT_SSMODE)
     {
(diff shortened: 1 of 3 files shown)


mattipee:
Super, Michel. :+1: 

1. I can confirm the version number is reported correctly.

As an aside: I note also that when the motor controller power switch is off, there is no version message and the display shows "Connected". I wonder whether absence of an expected response might trigger a warning to the user, suggesting for example "No response from motor controller. Is it connected?". A thought, to discuss. It would be heavy handed to fail the connection outright.


2. The echo'd force is, for me, still one cycle out-of-sync. If I ignore the first response, I see the value I expect. I will confirm that the current force and speed values also update in that second response wrt the first, but that was my observation beforehand (50ms, double read, discard the first).

Perhaps, therefore, the "head-unit model split" could be minimally implemented by changing the termination condition for the retry loop as follows:

```

             for(int i = 0; i < 5; i++) {
                 actualLength = readMessage();
                 if (actualLength < 0) {
                     qDebug() << "usb read error " << actualLength;
                 }
                 // qDebug() << "Read reply " << QByteArray((const char *)buf, actualLength).toHex(':');
-                if(actualLength >= 48) break;
+                if(actualLength == 48 || (actualLength == 64 && i > 0)) break;
             }
```

There would be clearer ways to express this, but the above I believe is logically correct.


3. Is there anything significant to address with the bare sendOpenCommand() call in sendRunCommand() "mode == FT_IDLE"?

4. Trivial point - leading whitespace of the qDebug and closing brace above differs from the surrounding code.

Mickyduck55:
Hi I used this today for my cycle.. As @mattipee says the message is displayed on connection
IMG_9303

@michelrdagenais  @mattipee As you both on here some observations from today, I had two issues probably nothing to do with this PR

1.At a slope of over over 5% with low cadence the brake seemed to be right on the edge of stuttering i.e. a slight hesitation in te roll

2 When I hit the pedals hard the rear end totally let go ... no resistance at all, power up to 600+

The two screen grabs are from GC today
Screenshot 2021-01-27 111926
Screenshot 2021-01-27 112055

michelrdagenais:
> 1. I can confirm the version number is reported correctly.
Great!

> As an aside: I note also that when the motor controller power switch is off, there is no version message and the display shows "Connected". I wonder whether absence of an expected response might trigger a warning to the user, suggesting for example "No response from motor controller. Is it connected?". A thought, to discuss. It would be heavy handed to fail the connection outright.

Good point, it is trivial to send an informative message if no version number is obtained. I will add something. The "connected" message comes from somewhere else; at least it means that the head unit is connected.

>              for(int i = 0; i < 5; i++) {
>                  actualLength = readMessage();
>                  if (actualLength < 0) {
>                      qDebug() << "usb read error " << actualLength;
>                  }
>                  // qDebug() << "Read reply " << QByteArray((const char *)buf, actualLength).toHex(':');
> -                if(actualLength >= 48) break;
> +                if(actualLength == 48 || (actualLength == 64 && i > 0)) break;
>              }

Indeed, the current termination condition (actualLength >= 48) does not mean anything for the T1932, with all replies at 64 bytes. Yet, skipping the incorrect replies is key to getting the correct behavior. Is there a better condition than always taking the second reply? You could wait for the proper force echo, or some other criteria. Could you quote a few reply sequences (bad and good replies)? We could see which field could be used as indication (e.g. buf[24..27]). I will then update the code accordingly.

It would be useful to know if we have a T1932 unit. LibUsb does know but does not communicate that information. My idea is to add a field for the vendor and product id in the LibUsb class. It could then be accessed in Fortius with the following. What do you think?

this->usb2->idProduct == FORTIUSVR_PID

> 1. Is there anything significant to address with the bare sendOpenCommand() call in sendRunCommand() "mode == FT_IDLE"?

Humm, it is not there to "open" the device but only to keep sending something while it is idle. It is thus different from the other cases. The fact that they are sending commands without reading the replies probably explains why you need to skip a few old replies later. We could read and discard a reply or two if you think that could help.

> 2. Trivial point - leading whitespace of the qDebug and closing brace above differs from the surrounding code.

On what line, are you looking at 649? There are a few places in the code where the indentation uses tabs, instead of spaces, and then it depends on how you set your tabs. At line 649, there is a space, // qDebug, while on line 650 there is none, //qDebug, and it starts with a tab instead of 4 spaces.

My plan was to do a nice cleanup of the file eventually and add some comments (for instance about the request/reply format including the force echo field), in a separate PR that does not touch any code.

michelrdagenais:
Indeed, I really dont see how this PR could have an effect on what you report. Is it a noticeable and repeatable difference between this PR and the master branch?

> @michelrdagenais @mattipee As you both on here some observations from today, I had two issues probably nothing to do with this PR

Mickyduck55:
IDK its the first time I have ridden this ride and really was mainly for
information for @mattipee <tunabiscuits796@gmail.com> I will ride it again
tomorrow using the other version.  One other thing today again I guess is
unrelated... I decided to cal after 15 mins and pressed the button before
my rear wheel had stopped... it just kicked in and calibrated without the
hard stop then start... maybe I was lucky and had the speed just right.

mattipee:
@Mickyduck55 I'll reply to your points here quickly but for clarity, please let's keep the PR comments on topic. :)


> 1.At a slope of over over 5% with low cadence the brake seemed to be right on the edge of stuttering i.e. a slight hesitation in te roll

Yes... the force limiting routine in Fortius, now merged to master and therefore in this PR, is not perfect - I believe you are experiencing this. The threshold is 10 kph and the behaviour also feels to me as if "on the edge or stuttering". If you are otherwise running with Eric's virtualgears branch in order to maintain higher wheel speeds, please continue to do so for your main rides. :) An improvement is on my radar, but incomplete. I'll tag in you in a subsequent PR, when I get there.


> 2 When I hit the pedals hard the rear end totally let go ... no resistance at all, power up to 600+ Top picture

I will look into this separately. If you can reproduce, either raise a bug or contact me directly.


> I decided to cal after 15 mins and pressed the button before
> my rear wheel had stopped... it just kicked in and calibrated without the
> hard stop then start... maybe I was lucky and had the speed just right.

The fact that the wheel is spinning can provide enough force to the motor such that calibration will start. I have a tentative routine to enable running calibration, which I'll tag you on when I raise that PR.

src/Train/Fortius.cpp
```diff
@@ -434,14 +436,28 @@ void Fortius::run()
...
+            //
+            // If you simply reissue the run command when 24 bytes are
+            // received, the following 48 bytes reply will match the force echo
+            // of the before last run command.
+
             msleep(50);
+            for(int i = 0; i < 5; i++) {
+                actualLength = readMessage();
+                if (actualLength < 0) {
+	            qDebug() << "usb read error " << actualLength;
```
mattipee, line 455: lines 455 and 456 leading whitespace is mixed tabs/spaces. Absolutely, leave other whitespace in the file to a cleanup PR... just as this is a loop specifically added in this PR, trivial to change.

src/Train/Fortius.cpp
```diff
@@ -434,14 +436,28 @@ void Fortius::run()
...
+	        }
+                // qDebug() << "Read reply " << QByteArray((const char *)buf, actualLength).toHex(':');
+                if(actualLength >= 48) break;
```
mattipee, line 458: Comment here to keep subsequent replies together.

Initial proposal:

`if(actualLength == 48 || (actualLength == 64 && i > 0)) break;`

mattipee:
@michelrdagenais 

> It would be useful to know if we have a T1932 unit. LibUsb does know but does not communicate that information. My idea is to add a field for the vendor and product id in the LibUsb class. It could then be accessed in Fortius with the following. What do you think?
> 
> this->usb2->idProduct == FORTIUSVR_PID

I'm a **for** exposing vendor/product via LibUsb. Personally, I think it fits within the remit of this PR, as timing/logic is model-sensitive. On the otherhand, it should be a simple and non-contentious PR in it's own right, on which we could build. Perhaps a steer from @amtriathlon ?

src/Train/Fortius.cpp
mattipee: > Is there a better condition than always taking the second reply?

There is nothing simpler I can think of.


> You could wait for the proper force echo ...

I tried this, there are complications with either two sends issuing the same target resistance, or with dropped sends meaning you might be waiting for a response containing a value that will never come. The complexity of handling this in code is too great, compared with the simplicity and apparent suitability of always taking the second reply.


> or some other criteria. ... see which field could be used as indication ...

Worth a look, but "second reply" is simple and tests well, and would therefore be my fallback if nothing immediately stands out as "better".

src/Train/Fortius.cpp
```diff
@@ -646,7 +714,7 @@ int Fortius::sendRunCommand(int16_t pedalSensor)
         qToLittleEndian<int16_t>((int16_t)(130 * brakeCalibrationFactor + N_to_rawForce(brakeCalibrationForce_N)), &SLOPE_Command[10]);
         
         retCode = rawWrite(SLOPE_Command, 12);
-        // qDebug() << "Send Gradient " << gradient << ", Weight " << weight << ", Command " << QByteArray((const char *)SLOPE_Command, 12).toHex(':');
+        // qDebug() << "Slope Command " << QByteArray((const char *)SLOPE_Command, 12).toHex(':');
     }
     else if (mode == FT_IDLE)
     {
```
mattipee, line 720: FT_IDLE sendOpenCommand()... don't know if this comment will appear at the right place in source, as it's not a line that has change in your branch.

> Humm, it is not there to "open" the device but only to keep sending something while it is idle. It is thus different from the other cases. **The fact that they are sending commands without reading the replies probably explains why you need to skip a few old replies later.** We could read and discard a reply or two if you think that could help.

The comment in bold above is interesting. However, given the proposed change to the termination condition means that I always "read-twice-and-discard-the-first", it feels any ill effect of unread responses would be long gone, no?

src/Train/Fortius.cpp
michelrdagenais: Good point, for non changing force, checking the echo is pointless. There is nothing to check in bytes buf[24..27]? When reading the version, it is a good criteria. Otherwise, indeed let's keep this simple solution for now.

michelrdagenais:
> I'm a **for** exposing vendor/product via LibUsb. Personally, I think it fits within the remit of this PR, as timing/logic is model-sensitive. On the otherhand, it could be a simple and non-contentious PR in it's own right, on which we could build. Perhaps a steer from @amtriathlon ?

My question was more if there was some other better way to get that information already. Otherwise, I will push that separately, especially since we are not using yet.

src/Train/Fortius.cpp
michelrdagenais: Yes, since we are more robust now with "read twice" elsewhere, there is no problem leaving this as is. It would be more for consistency.

src/Train/Fortius.cpp
mattipee: I'll get back on the trainer later and provide some data.

mattipee: Hopefully this image is readable...

- commanded force [4:5] of sent command in green
- echoed force [40:41] where green is correct, red is "stale"
- speed [32:33] and current force [38:39] in yellow, always seen to change on the second reading.

image

michelrdagenais: Humm, interesting, thanks for letting us see the 64 bytes replies!

We cannot tell that the first reply is incorrect by itself. Is is identical to the previous reply which was good. We cannot either disqualify a reply on the basis that it is duplicated. We could reach a steady state. Possible alternatives:

- just keep the second reply as suggested, it is arbitrary and inelegant but we are dealing with less than perfect low-level hardware here. It is simple and seems to work;
- read/skip a message just before sending the run command, would that get rid of the duplicate on the T1932, it does not on the T1942.
- check if the distance is the same but the speed is not zero, this is a duplicate to discard. This is more elegant but does not cover the case where the speed is zero. It is probably not worth the complexity.

Any further thoughts? Otherwise, I will test/update the PR, to skip the first reply, this evening.

michelrdagenais:
It should be good now. The new message "Is the motor brake on?" is neat, I forget from time to time to turn it on.
@mattipee Is it good for you?

src/Train/Fortius.cpp
mattipee: Rather than screenshot, I'll paste a few lines below... you may need to copy-paste to editor to avoid line wrap...

Summary is that a pre-send read appears to do nothing to get rid of the duplicate. Always the second send is the one we want.

Distance... interesting, but I think semantic rules on what to accept or not do have some merit - whether distance, an expected value, etc - but I think skipping the first reply is the lighter touch and I'd certainly be happy with it.


```
Pre-send read    "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:3d:cd:02:00:96:0a:d8:6c:76:02:e5:08:c7:13:00:00:44:0d:02:1f:28:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
Send ERGO  "01:08:01:00:99:14:00:00:02:0a:6b:04"
Read reply   0   "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:3d:cd:02:00:96:0a:d8:6c:76:02:e5:08:c7:13:00:00:44:0d:02:1f:28:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
Read reply   1   "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:3f:cd:02:00:70:0a:88:6e:76:02:b8:09:99:14:00:00:44:0d:02:52:fe:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
Read reply   2   "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:3f:cd:02:00:70:0a:88:6e:76:02:b8:09:99:14:00:00:44:0d:02:52:fe:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
Pre-send read    "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:3f:cd:02:00:70:0a:88:6e:76:02:b8:09:99:14:00:00:44:0d:02:52:fe:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
Send ERGO  "01:08:01:00:e4:14:00:00:02:0a:6b:04"
Read reply   0   "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:3f:cd:02:00:70:0a:88:6e:76:02:b8:09:99:14:00:00:44:0d:02:52:fe:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
Read reply   1   "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:42:cd:02:00:76:0a:68:6e:76:02:e6:0a:e4:14:00:00:44:0d:02:5f:39:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
Read reply   2   "0b:d5:0b:00:00:02:00:00:09:00:00:00:00:00:00:00:00:00:0d:0a:00:00:00:00:03:13:02:00:42:cd:02:00:76:0a:68:6e:76:02:e6:0a:e4:14:00:00:44:0d:02:5f:39:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
```

mattipee:
> It should be good now. The new message "Is the motor brake on?" is neat, I forget from time to time to turn it on.
> @mattipee Is it good for you?

Yes - fetched, built, tested motor brake message, reconnect with motor on, rode a few pedal strokes with two qDebug() uncommented, two reads, second is the good one.

All good! :+1:

src/Train/Fortius.cpp
mattipee: > thanks for letting us see the 64 bytes replies!

As an aside, I did at one point wonder if it was a peculiarity of Linux and libusb... and I zero'd the buffer before read. Without repeating that again now, I think I can state that these are all genuine, full 64 (or at least 49-significant) byte responses.

PR #3804, opened 2021-01-27
Fortius "rolling" calibration

mattipee:
Following on from the introduction of Fortius calibration in #3790, this follow on PR contains a small tweak to gracefully decay wheel speed to the target calibration speed, in the event that calibration is commanded while user is riding in excess of that speed.

The existing implementation results in the motor inducing a very forceful and sudden wheel speed change.

### DIFF
--- a/src/Train/Fortius.cpp
+++ b/src/Train/Fortius.cpp
@@ -597,6 +597,37 @@ int Fortius::sendCalibrateCommand()
 {
     uint8_t calibrate_command[] = {0x01,0x08,0x01,0x00,0xa3,0x16,0x00,0x00,0x03,0x00,0x00,0x00};
     
+    static double excess_speed = 0;
+
+    // Detect initial condition... rider is already in excess of 20kph calibration speed (ie performing a rolling calibration)
+    // If so, we want to apply lower calibration speed gradually, otherwise the feel at the pedals is NOT GOOD
+
+    // Notes - if rider is below 20, or at 0, go immediately to calibration speed, as the bike's freewheel will allow this (unless they're on a fixie!!)
+    //       - use 21kph as excess threshold, don't want false-positives when in steady state, it's never EXACTLY 20 when calibration is running
+    //       - if user happens to be just below 21kph and initiates a rolling calibration, it may start with a little "bump", but it's REALLY not a big deal
+    //       - even using 21 as the excess threshold, this can trigger from a standing start as wheel speed can overshoot the target
+    //         (never mind becuase this logic will just decay it nice and smoothly, rather than it "snapping back" to 20)
+
+    if (deviceSpeed > 21 && excess_speed == 0)
+    {
+        // Record excess speed
+        excess_speed = deviceSpeed - 20;
+    }
+
+    // Each time period during calibration, if we have excess_speed to decay, decay it
+    // Note: ignore deviceSpeed, otherwise we can introduce oscillation
+    //       wheel speed responds very forcefully in calibration mode and can over/undershoot
+    if (excess_speed > 0)
+    {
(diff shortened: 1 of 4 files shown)


mattipee:
@mickyduck55 please might you give your opinions on this here, once artefact has built? It appears to work very smoothly for me.

Mickyduck55:
@mattipee <tunabiscuits796@gmail.com> delighted to..I will let you know
tomorrow

mattipee:
If speed is below 20kph, it will immediately target 20kph, as it would do from a standing start... the benefit of having a freewheel cassette is that wheel speed is free to increase.

The danger, in an extreme case of mistiming, I guess, is that immediately targeting 20kph when you're starting over speed, it could damage a knee. So this PR to gradually decay excess speed could, in fact, be described as a safety feature.

Aside: I suspect the case you reported of the wheel actually stopping when you clicked calibrate while pedalling may have been due to such an initial force applied to slow the wheel when it was already at high speed, that it resulted in the wheel speed undershooting 20 significantly and dropping to <8kph or so, around which I believe the motor will cut out if no force is being applied. Alternatively, if it wasn't to undershoot 20 quite as badly, then I often experience that it will pick back up again on its own, as you described earlier.

Does this all make sense?

Mickyduck55:
yes all clear... Gradually decay speed sounds good.. Personally I would
probably kick off calibration from near 20 k.. not very sensible to start
it if I'm at 45 kph on top gear going downhill or struggling up a slope at
10 kph.. I guess some people may do so i will try both scenarios tomorrow.
Thanks

mattipee:
:+1: I'm particularly interested if, at the point you click calibrate while pedalling in excess of 20kph, you experience any significant/sudden drop in speed/stop as you had reported much earlier. I do still need to trace whether any "idle" or such commands are being sent in-between your "grade" ride and the "calibration" mode that could interfere with the smooth transition to decay.

I can say however that this PR appeared to work very well for me.

Mickyduck55:
Ok got it and installed

Mickyduck55:
Thank you @mattipee 
99,9% perfect..
Tested from stationary.. As before
Tested from 45 kph.. speed slowed and stopped at 20 kph then calibrated as before
Tested from 15kph.. speed increased to 20 kph calibrated as normal.. I don't know is this was expected as the previous version did this yesterday so maybe I was lucky with being at a sensible speed.  Tried it again with the same result.
Now for the 0.01%... If the wheel is moving you still get the message to kick the pedal.. very fleetingly as the process detects spinning wheel and message goes... The kick pedal message not required if wheel spinning..
This is so much better when you are using a video to train or even using ERG a long flat 10 min 100W warmup.. press cal.. wait then get on with it all with one mouse click.

mattipee:
Thank you, @Mickyduck55.

I have addressed your 0.1% in the latest commit.

Mickyduck55:
Thanks @mattipee I will test tomorrow

Mickyduck55:
Tested this morning same scenarios as yesterday.. with same results ALL OK.  TBH message was gone so quickly I didn't even get time to read it.. so will assume it is displaying correct message .. didn't see Kick pedal except when stationary.  The connection to my Fortius dropped out at 10 min but it has done this once before and could not repeat it.

mattipee:
Between 0 and 20, the speed reaches 20 very quickly, so you won't see the "allow wheel speed to settle" message for long.
The higher above 20 you start, the longer you'll see it.

If you don't kick the pedal hard enough but just tickle it at a few kph, you'll see the "allow to settle" message until you stop pedalling, and it will revert to stationary and "give the pedal a kick".


Thanks for the test.

PR #3806, opened 2021-01-29
Get the Altitude for CRS files even when Fast Forwarding

### LINKED ISSUE #3805: Altitude not updated when seeking forward/backward in a .pgmf file
With some types of workout files like .pgmf, the altitude is computed incrementally from the slope in TrainSidebar.cpp. When seeking forward or backward (by 1km), this is not taken into account to update the altitude. It seems that the ErgFile structure now already computes the altitude from the slope at reading time. It should be possible to access the altitude with a new function AltitudeAt, modeled after ErgFileQueryAdapter::gradientAt?

michelrdagenais:
In the TrainSidebar main loop, the altitude is derived from the gradient for some CRS files. However, this computation does not take into account when the displayDistance is changed with fastForward or fastRewind. Thus, when advancing in this mode in a .pgmf file, the altitude would not be updated.

However, when CRS workout files are read, the altitude is already derived and precomputed from the gradient for all formats handled. A new function was added, altitudeAt, that accesses this value, avoiding to derive anew the altitude from the gradient in the TrainSidebar main loop, and getting the correct altitude value when FastForward.

Fixes #3805

### DIFF
--- a/src/Train/ErgFile.cpp
+++ b/src/Train/ErgFile.cpp
@@ -1949,6 +1949,26 @@ ErgFileQueryAdapter::gradientAt(double x, int& lapnum) const
     return gradient;
 }
 
+// Altitude
+double
+ErgFileQueryAdapter::altitudeAt(double x, int& lapnum) const
+{
+    // Establish index bracket for query.
+    if (!updateQueryStateFromDistance(x, lapnum))
+        return -1000;
+
+    assert(hasGradient()); // altitudeAt should not be called if ergfile has no gradient and altitude.
+    if (!hasGradient())
+        return -1000;
+
+    ErgFilePoint p1 = Points().at(qs.leftPoint);
+    ErgFilePoint p2 = Points().at(qs.rightPoint);
+    double altitude = p1.y;
+    if(p1.x != p2.x) altitude += (p2.y - p1.y) * (x - p1.x) / (p2.x - p1.x);
+
+    return altitude;
+}
+
 // Returns true if a location is determined, otherwise returns false.
 // This is the const stateless edition(tm), which does not rely on location
 // state. This method is used to make queries independant of the main ergfile
--- a/src/Train/ErgFile.cpp
+++ b/src/Train/ErgFile.cpp
@@ -1901,9 +1901,10 @@ ErgFileQueryAdapter::wattsAt(double x, int& lapnum) const
     if (!updateQueryStateFromDistance(x, lapnum))
         return -100;
 
-    assert(hasWatts()); // wattsAt should not be called unless ergfile has watts.
-    if (!hasWatts())
+    if (!hasWatts()) {
+        qDebug() << "wattsAt should not be called unless ergfile has watts";
         return -100;
+    }
 
     // two different points in time but the same watts
     // at both, it doesn't really matter which value
@@ -1940,9 +1941,10 @@ ErgFileQueryAdapter::gradientAt(double x, int& lapnum) const
     if (!updateQueryStateFromDistance(x, lapnum))
         return -100;
 
-    assert(hasGradient()); // gradientAt should not be called if ergfile has no gradient.
-    if (!hasGradient())
+    if (!hasGradient()) {
+        qDebug() << "gradientAt should not be called if ergfile has no gradient";
         return -100;
+    }
 
     double gradient = Points().at(qs.leftPoint).val;
 
@@ -1957,9 +1959,10 @@ ErgFileQueryAdapter::altitudeAt(double x, int& lapnum) const
     if (!updateQueryStateFromDistance(x, lapnum))
         return -1000;
 
-    assert(hasGradient()); // altitudeAt should not be called if ergfile has no gradient and altitude.
-    if (!hasGradient())
+    if (!hasGradient()) {
+        qDebug() << "altitudeAt should not be called if ergfile has no gradient";
         return -1000;
+    }
 
     ErgFilePoint p1 = Points().at(qs.leftPoint);
     ErgFilePoint p2 = Points().at(qs.rightPoint);
@@ -1979,9 +1982,10 @@ bool ErgFileQueryAdapter::locationAt(double meters, int& lapnum, geolocation& ge
     if (!updateQueryStateFromDistance(meters, lapnum))
         return false;
 
-    assert(hasGradient()); // locationAt should not be called if ergfile has no gradient.
-    if (!hasGradient())
+    if (!hasGradient()) {
+        qDebug() << "locationAt should not be called if ergfile has no gradient";
         return false;
+    }
 
     // At this point leftpoint and rightpoint bracket the query distance. Three cases:
     // Bracket Covered: If query bracket compatible with the current interpolation bracket then simply interpolate
--- a/src/Train/ErgFile.h
+++ b/src/Train/ErgFile.h
@@ -130,7 +130,7 @@ class ErgFile
         double Cp;
         int format;             // ERG, CRS, MRC, ERG2 currently supported
 
-        bool hasGradient() const { return CRS == format; }
+        bool hasGradient() const { return CRS == format; } // Has Gradient and Altitude
         bool hasWatts()    const { return ERG == format || MRC == format; }
 
 private:
@@ -247,6 +247,7 @@ public:
     // State queries (maintain mutable state.)
     double wattsAt(double msec, int& lapnum) const;
     double gradientAt(double meters, int& lapnum) const;
+    double altitudeAt(double meters, int& lapnum) const;
     bool   locationAt(double meters, int& lapnum, geolocation& geoLoc, double& slope100) const;
 };
 
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -1780,12 +1780,8 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                         }
 
                         if (!fAltitudeSet) {
-                            // For classic rlv with no location data:
-                            // Estimate vertical change based upon time passed and slope.
-                            // Note this isn't exactly right but is very close - we should use the previous slope for the time passed.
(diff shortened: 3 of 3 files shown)


michelrdagenais:
I checked and gradientAt is not as bad as I thought. It keeps the last left / right point from the last query and does not perform a new linear search each time. Furthermore, the .at() call for the Points() QList is documented in Qt as being a very fast, constant time operation. They must be using some sort of arrays underneath this "list" interface. 

I thus implemented altitudeAt separately, to stay consistent with the existing interface. If you prefer a "more efficient" combined function, to retrieve the gradient and altitude at the same time, I can easily update this pull request.

mattipee:
Code review from me - minimal footprint, no further comments. :+1:

src/Train/ErgFile.cpp
```diff
@@ -1949,6 +1949,26 @@ ErgFileQueryAdapter::gradientAt(double x, int& lapnum) const
...
 
+// Altitude
+double
+ErgFileQueryAdapter::altitudeAt(double x, int& lapnum) const
+{
+    // Establish index bracket for query.
+    if (!updateQueryStateFromDistance(x, lapnum))
+        return -1000;
+
+    assert(hasGradient()); // altitudeAt should not be called if ergfile has no gradient and altitude.
```
amtriathlon, line 1960: Our builds don't disable assert, so it is better to use it only in unavoidable situations since it leads to a crash, for this case likely a message to qDebug() may be enough

michelrdagenais: Replaced all 4 identical asserts by qDebug(). Indeed, it is not a serious problem and it would be a shame to stop the program.

ericchristoffersen:
This change seems strange to me. What exactly is the case that this 'fixes'? Issue seems to suggest pgmf.

A crs is a simple strict mode slope workout so altitude should reflect the actual amount climbed. If you skip ahead you skipped the climbing (or descending) and I would not expect your altitude to change.

A pgmf is a virtual ride workout represented as CRS in ergfile. Perhaps the better fix would be to

     StrictGradient = false;

in the pgmf reader? That would cause altitude to be honored and yield smooth gradient and not break desired crs behavior.

michelrdagenais:
> A crs is a simple strict mode slope workout so altitude should reflect the actual amount climbed. If you skip ahead you skipped the climbing (or descending) and I would not expect your altitude to change.

It depends on the user expectations. This way, the same altitude is shown at the same distance, even if you skip some part. It matches the model of running along a course. If you want the meters climbed, then indeed you may not want that, but this is not "altitude" anyway. As for insuring that you have the correct altitude for the air density computation, this is a relatively small effect, unless you really climb a lot. 

> A pgmf is a virtual ride workout represented as CRS in ergfile. Perhaps the better fix would be to
> 
> ```
>  StrictGradient = false;
> ```
> 

My understanding was that this is not desirable, since the gradient is already available natively in this file, and thus it should not be "filtered down", as you need to do with a gpx file where the z coordinate jumps up and down with noise.

ericchristoffersen:
> > A crs is a simple strict mode slope workout so altitude should reflect the actual amount climbed. If you skip ahead you skipped the climbing (or descending) and I would not expect your altitude to change.
> 
> It depends on the user expectations. This way, the same altitude is shown at the same distance, even if you skip some part. It matches the model of running along a course. If you want the meters climbed, then indeed you may not want that, but this is not "altitude" anyway. As for insuring that you have the correct altitude for the air density computation, this is a relatively small effect, unless you really climb a lot.

Uh... I get the feeling you aren't actually using any of this. Because crs is slope only the altitude starts at zero. There is no semblance of realism. I think of crs as a sort of formal 'workout' file, the slope version of erg, it doesn't work well at all as a simulation workout but that is ok because that isn't what it is.

> > A pgmf is a virtual ride workout represented as CRS in ergfile. Perhaps the better fix would be to
> > ```
> >  StrictGradient = false;
> > ```
> 
> My understanding was that this is not desirable, since the gradient is already available natively in this file, and thus it should not be "filtered down", as you need to do with a gpx file where the z coordinate jumps up and down with noise.

Have you tried it? The non-strict mode is not filtering anything, it performs a smooth spline interpolation between altitudes, then computes the instant gradient at each point. If you had the top of a pass with 100m of 5%, then 100m of -5%, the strict mode would show 5%, then -5%. The continuous (non-strict) mode would show a continuous gradient from 5 to -5 (exactly the shape of the curve is determined by surrounding points. Non-strict mode to me seems like the correct approach for pgmf since they are actually used to model real rides - I don't think anyone wants strict gradient in sim-movie mode.

Did you guys verify altitude still works in manual slope mode?

Anyway. Good luck.

amtriathlon:
I think Eric have a point, to accurately record the meters climbed seems more relevant than to preserve the altitude-location relationship for practical purposes. I overlooked this issue.

michelrdagenais:
Training videos come with different types of "slope" files, .pgmf, .gpx, .tts... Sometimes only one type is available and other times all three types are (e.g. videos from http://rlv-alucard.de/rlvs.htm). Without this patch, the .pgmf file displayed the altitude computed incrementally, but not accounting for skipped sections, while the .pgx file for the same video displayed the altitude, accounting for the skipped sections. This does not appear very coherent.

If you want to have the elevation gain, to know how much you have "worked", this is a very valid metric, available on many cycling head units. This would be nice (and easy) to add. The previous behavior of GC, for some "slope" file types, was the "incremental altitude" which does not appear to be a very interesting metric but could be added as well.

ericchristoffersen:
> Training videos come with different types of "slope" files, .pgmf, .gpx, .tts... Sometimes only one type is available and other times all three types are (e.g. videos from http://rlv-alucard.de/rlvs.htm). Without this patch, the .pgmf file displayed the altitude computed incrementally, but not accounting for skipped sections, while the .pgx file for the same video displayed the altitude, accounting for the skipped sections. This does not appear very coherent.

Gradient mode Pgmf is a tacx format that was shipped with their virtual rides. Like crs the altitude is stuck starting at 0 (unless there's a hidden basis in the format that weve not found yet) I agree its a good thing to have altitude work correctly during skip for pgmf, which is maybe the point of the change.

Unfortunately the change also touched crs support and that change to crs feels broken to me. I suspect this also broke altitude in manual slope mode. Have you tried these?

> If you want to have the elevation gain, to know how much you have "worked", this is a very valid metric, available on many cycling head units. This would be nice (and easy) to add. The previous behavior of GC, for some "slope" file types, was the "incremental altitude" which does not appear to be a very interesting metric but could be added as well.

In crs workout mode I personally will skip back in order to get more climbing in. Blame me for running the wrong workout but its what I do. I just don't think it makes sense to talk about altitude for a crs because its all about gradient.

Pgmf altitude is computed from gradient and distance during pgmf load. It really feels to me like correct approach for pgmf is to have it use non-strict slope mode which will provide more realistic gradient for virtual ride. Have you tried this? I haven't used a pgmf in forever because my rides have tts. Its important to check your pgmf files and see if non-strict actually works well.

Is important to check that it works. Tacx files have surprises, for example the tts files contain altitude data but it is often crazy and unusable, instead we derive our altitude from gradient during tts load (using an altitude as a basis.)

If you think the non-strict pgmf behavior is acceptable then that change will also fix the altitude seek problem.

amtriathlon:
> Without this patch, the .pgmf file displayed the altitude computed incrementally, but not accounting for skipped sections, while the .pgx file for the same video displayed the altitude, accounting for the skipped sections. This does not appear very coherent.

I agree we should have a coherent recording strategy: the altitude corresponding to the point in the workout (explicit or implicit) or the one accrued by the incremental climbing and descending.

The first makes existing Elevation Gain metric well behaved for the whole activity, the second only piecewise between FFW/REW events and it is similar to what happens in the real work if the user pauses recording in one place and restart in a different one, only difference is the lack of a time gap since our "teleportation" is instantaneous.

> If you want to have the elevation gain, to know how much you have "worked", this is a very valid metric, available on many cycling head units. This would be nice (and easy) to add. The previous behavior of GC, for some "slope" file types, was the "incremental altitude" which does not appear to be a very interesting metric but could be added as well.

I am talking about pos-ride analysis where the metrics are computed from the altitude/elevation series.

michelrdagenais:
> I agree we should have a coherent recording strategy: the altitude corresponding to the point in the workout (explicit or implicit) or the one accrued by the incremental climbing and descending.

We could have both metrics available. Total ascent and total descent would also be interesting metrics. The accrued altitude can be derived from the total ascent - total descent.

> I am talking about pos-ride analysis where the metrics are computed from the altitude/elevation series.

I see, I have not used much the post-ride analysis.

amtriathlon:
> We could have both metrics available. Total ascent and total descent would also be interesting metrics. The accrued altitude can be derived from the total ascent - total descent.

To add yet another RealTimeData field doesn't change what we log as the Altitude field in the generated activity, which is my concern.

Anyway, it is likely reasonable to log the altitude corresponding to the location in the ErgFile, but when there is no ErgFile (aka Manual Slope mode) we should go back to the previous behavior since currently no altitude is recorded in this mode.

Additionally we should document this, specially since the behavior of FWD/RWD on post-ride Elevation Gain may be not intuitive at first.

michelrdagenais:
> > Anyway, it is likely reasonable to log the altitude corresponding to the location in the ErgFile, but when there is no ErgFile (aka Manual Slope mode) we should go back to the previous behavior since currently no altitude is recorded in this mode.
> 
> I was quite careful to check the code paths when I modified the code. It looked like this code was only called for CRS files. Looking at the code again, it appears that manual slope mode was not using the incremental altitude computation, because if(ergfile) was false in that case. If I understand correctly, this would be new behavior for the manual slope mode, to account for the "accrued altitude". I agree that it makes sense though.

This new behavior, computing the "accrued altitude" from the slope, was added for the Manual Slope Mode in PR #3847

> > Additionally we should document this, specially since the behavior of FWD/RWD on post-ride Elevation Gain may be not intuitive at first.
> 
> I can document that in the Train View, in the workout section and in the workout control (where you command the FWD/RWD). Any place else?

This was documented in the wiki at both these two places.

PR #3808, opened 2021-01-30
Don't rewrite gradient course files, allow sim to see original gradient

### LINKED ISSUE #3807: Intensity adjustement should preserve original slope profile for "sim speed" to work correctly
With intensity adjustment for slope workouts as it stands, after course rewriting, the sim only sees the rewritten grade. If you tell the trainer "do 5% instead of 10%", then the sim still needs to know the original 10% in order to make faithful progress along the original course.

The PR I intend to raise against this I hope will also be a useful precursor to #3749.

See also comment https://github.com/GoldenCheetah/GoldenCheetah/issues/3749#issuecomment-770067824
See also https://groups.google.com/g/golden-cheetah-users/c/G7yYcXf-UD4/m/6zGhwuSoAAAJ

mattipee:
First commit for PR to fix #3807

Note: there are a couple of things that might be good to include, but I'd like the initial commit to be reviewed first, and subsequent changes either agreed in-scope, or for another PR.

- user message in Higher()/Lower() to reflect change in intensity
- additional metrics in rtData for completeness:
  - slopeToTrainer, slopeFromTrainer
  - deviceSpeed, simulatedSpeed

### DIFF
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -349,7 +349,7 @@ TrainSidebar::TrainSidebar(Context *context) : GcWindow(context), context(contex
     wbalr = wbal = 0;
     load_msecs = total_msecs = lap_msecs = 0;
     displayWorkoutDistance = displayDistance = displayPower = displayHeartRate =
-    displaySpeed = displayCadence = slope = load = 0;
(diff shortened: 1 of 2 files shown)


mattipee:
I made a commit mistake.

PR #3809, opened 2021-01-30
Don't rewrite gradient course files, allow sim to see original gradient 

### LINKED ISSUE #3807: Intensity adjustement should preserve original slope profile for "sim speed" to work correctly

mattipee:
PR to fix #3807

There are essentially three main points:
 - slope course is NOT rewritten
 - sim sees original course gradient
 - gradient-scaling is done using slopeToTrainer, instead of course-rewriting

Note: there are a couple of related things that might be good to include, but I'd like the initial commit to be reviewed first, and any subsequent changes either agreed in-scope, or for another PR.

- user message in Higher()/Lower() to reflect change in intensity
  - improve usability
- additional metrics in rtData for completeness:
  - slopeToTrainer, slopeFromTrainer
    - by removing the local.getSlope() in guiUpdate(), we lose slopeFromTrainer, where trainer may "clip" the requested grade
    - it may or may not be useful to have that
  - deviceSpeed, simulatedSpeed
    - currently, displaySpeed is used to store either deviceSpeed, or simulatedSpeed if sim speed is enabled
    - there is value in differentiating the two, probably in addition to how displaySpeed is used currently.

### DIFF
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -1693,6 +1693,12 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                 }
             }
 
+            // In order to differentiate trainer speed from simulated speed it would be ideal
+            // to have trainers setTrainerSpeed() directly, instead of setSpeed() as current
+            // Here, we cross populate value, and call setSpeed() later with the correct value
+            // to keep the "Speed" dataseries in the train GUI showing correctly
+            rtData.setTrainerSpeed(rtData.getSpeed());
+
             // If simulated speed is *not* checked then you get speed reported by
             // trainer which in ergo mode will be dictated by your gear and cadence,
             // and in slope mode is whatever the trainer happens to implement.
@@ -1859,11 +1865,14 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                 rtData.setLapMsecs(lap_elapsed_msec);
             }
 
+            // Set correct value for "Speed" dataseries in GUI, also affects dispalySpeed below
+            rtData.setSpeed(useSimulatedSpeed ? rtData.getSimulatedSpeed() : rtData.getTrainerSpeed());
+
             // local stuff ...
             displayPower = rtData.getWatts();
             displayCadence = rtData.getCadence();
             displayHeartRate = rtData.getHr();
-            displaySpeed = useSimulatedSpeed ? rtData.getSimulatedSpeed() : rtData.getSpeed();
+            displaySpeed = rtData.getSpeed();
             displayLRBalance = rtData.getLRBalance();
             displayLTE = rtData.getLTE();
             displayRTE = rtData.getRTE();
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -349,7 +349,7 @@ TrainSidebar::TrainSidebar(Context *context) : GcWindow(context), context(contex
     wbalr = wbal = 0;
     load_msecs = total_msecs = lap_msecs = 0;
     displayWorkoutDistance = displayDistance = displayPower = displayHeartRate =
-    displaySpeed = displayCadence = slope = load = 0;
+    displaySpeed = displayCadence = slope = slopeToTrainer = load = 0;
     displaySMO2 = displayTHB = displayO2HB = displayHHB = 0;
     displayLRBalance = displayLTE = displayRTE = displayLPS = displayRPS = 0;
     displayLatitude = displayLongitude = displayAltitude = 0.0;
@@ -1189,7 +1189,7 @@ void TrainSidebar::Start()       // when start button is pressed
 
         // START!
         load = 100;
-        slope = 0.0;
+        slope = slopeToTrainer = 0.0;
 
         // Reset Speed Simulation
         bicycle.clear();
@@ -1361,7 +1361,7 @@ void TrainSidebar::Stop(int deviceStatus)        // when stop button is pressed
     //}
 
     load = 0;
-    slope = 0.0;
+    slope = slopeToTrainer = 0.0;
 
     QDateTime now = QDateTime::currentDateTime();
 
@@ -1633,7 +1633,7 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
             return;
         } else {
             rtData.setLoad(load); // always set load..
-            rtData.setSlope(slope); // always set load..
+            rtData.setSlope(slope); // always set (original) slope..
             rtData.setAltitude(displayAltitude); // always set display altitude
 
             double distanceTick = 0;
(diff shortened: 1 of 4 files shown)


src/Train/TrainSidebar.cpp
```diff
@@ -1784,7 +1776,11 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                             displayAltitude = ergFileQueryAdapter.altitudeAt(displayWorkoutDistance * 1000, displayWorkoutLap);
                         }
 
-                        rtData.setSlope(slope);
+                        // Reading a new slope from ergFile with gradient means we may need to adjust slopeToTrainer, if intensity is not 100%
+                        // This (possibly modified) value is used to populate the GUI
```
mattipee, line 1780: ...is **also** used...

src/Train/TrainSidebar.cpp
```diff
@@ -1647,17 +1647,8 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                 // get spinscan data from a computrainer?
                 if (Devices[dev].type == DEV_CT) {
                     memcpy((uint8_t*)rtData.spinScan, (uint8_t*)local.spinScan, 24);
-                    rtData.setLoad(local.getLoad()); // and get load in case it was adjusted
```
mattipee, line 1650: Perhaps leave load alone for now... removal here and below for DEV_FORTIUS||DEV_IMAGIC may have been heavy-handed.

src/Train/TrainSidebar.cpp
```diff
@@ -1784,7 +1776,11 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                             displayAltitude = ergFileQueryAdapter.altitudeAt(displayWorkoutDistance * 1000, displayWorkoutLap);
                         }
 
-                        rtData.setSlope(slope);
+                        // Reading a new slope from ergFile with gradient means we may need to adjust slopeToTrainer, if intensity is not 100%
+                        // This (possibly modified) value is used to populate the GUI
+                        slopeToTrainer = slope * lastAppliedIntensity/100.;
+                        rtData.setSlope(slopeToTrainer);
+
```
mattipee, line 1783: If "sim speed" is not enabled, then perhaps we should scale altitude to what has "actually" been achieved... rather than what the file said originally... the use-case I'm thinking of is wanting to go and do a pretty video climb, without sim-speed, at reduced intensity, and have the altitude/gain/loss reflect what's actually been ridden. Obviously submitting a reduced-intensity ride without sim-speed enabled to Strava or something is going to give artificial times on virtual segments, but we could at least be calculating altitude faithful to the reduced gradient, and relative to the altitude at the start of the course.

src/Train/TrainSidebar.cpp
```diff
@@ -2685,6 +2681,14 @@ void TrainSidebar::adjustIntensity(int value)
     ErgFile* ergFile = const_cast<ErgFile*>(ergFileQueryAdapter.getErgFile());
     if (!ergFile) return; // no workout selected
 
+    if (ergFile->hasGradient()) {
+        // Don't rewrite gradient courses, changing lastAppliedIntensity is enough
+        lastAppliedIntensity = value;
+        slopeToTrainer = slope * lastAppliedIntensity/100.; // also updated in guiUpdate()
```
mattipee, line 2687: this particular line may be redundant

Mickyduck55:
@mattipee 
Something really BROKEN here
connection to Fortrius drops out shortly after starting.. First time I got the message telling me brake version etc
Started to cycle within 1k like the bad old days no connection video still moving but cad HR power etc frozen.
shut down and Re started Message "is brake on"... yes it was

Re booted PC started Ok but connection lost again very soon into ride

Loaded my "other" version all ok so not my hardware.. or does not seem to be

mattipee:
> @mattipee
> Something really BROKEN here

That doesn't sound good. It works for me, and I can't imagine any change in this PR is the cause.

Let's diagnose without adding to the conversation on this PR. If you can, try the main branch artefact, and I'll catch up with you directly. :+1:

Mickyduck55:
Sorry where can I get the main branch artefact

src/Train/TrainSidebar.cpp
michelrdagenais: We had problems in the past with that when the telemetry load was used to compute the intensified gradient. Using the same variable for dual purposes is bad. Ultimately, we may need to have separate telemetry variables for simulated slope (specified in the gpx file), requested slope (affected by virtual gearing) and device slope (what the device or driver finally did, because of limitations). For load I guess that this would give load (specified in the Erg file) and device load (changed because of limitations). 

I am all for removing the current code where we set the load for the device and the device sets a different value back.

src/Train/TrainSidebar.cpp
michelrdagenais: I would again suggest to have both values available in the Telemetry, (simulated) slope and requested slope (or slopeToTrainer). Then, different parts of the program can decide which value to use.

mattipee: I've added slopeToTrainer to RealtimeData. I'll hold off exposing the dataseries just for the moment. The default behaviour at the moment is to maintain the requested slope as the value presented in "Slope" dataseries, as it was when rewriting.

src/Train/TrainSidebar.cpp
```diff
@@ -1864,13 +1866,14 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
                 rtData.setLapMsecs(lap_elapsed_msec);
             }
 
+            // Set correct value for "Speed" dataseries in GUI, also affects dispalySpeed below
+            rtData.setSpeed(useSimulatedSpeed ? rtData.getSimulatedSpeed() : rtData.getTrainerSpeed());
```
mattipee, line 1870: Similarly, trainerSpeed and simulatedSpeed are now in RealtimeData, though the "Speed" dataseries is still the only one exposed to the GUI, and is now explicitly set according to bool useSimulatedSpeed, to maintain behaviour that the single "Speed" series box in the gui will show the correct value, whether useSimulatedSpeed is true or false.

mattipee:
It already feels to me like there are a couple of orthogonal/independent changes wrapped up in here which could be split out two or three very simple PRs. In turn, these would make this PR more concise and easier to review.

Certainly these two:
- establishing trainerSpeed/simulatedSpeed in RealtimeData and populating them appropriately, maintaining the "Speed" dataseries behaviour so as not to impact GUI
- removal of the load/grade "read back" from trainer's, where we are overwriting values by doing so

Would this be preferable, to separate some smaller PRs?

mattipee:
>1: If you are patching for a bug or adding a feature in existing code
>
>Focus on a single bug or feature at a time and do not submit style formatting or refactoring changes in the original patch, please.
>
>Just submit the changes that fix the bug or tweak the code with a clear and, ideally concise, description. This is so those who review the code can see what is being done easily, and it allows each change to be reverted if it needs to.

Instinct suggests that if I myself think that this PR now contains too many moving parts (see previous comment), then others will too. Being fairly new to GC, could I get a quick steer from you please, @amtriathlon ?


Also have you any comment, either here or falling back to either #3807 or the google group as appropriate, on the overall goal here and the way in which I'm approaching it:
  - a clean fix to intensity modification without course rewriting
  - avoiding setGradeWithSimState()
  - manoeuvring the code towards a minimal PR for Eric's virtual gears

I get the impression @ericchristoffersen has his focus beyond this work now (or is there still overlap, Eric?), but I'm happy to commit to carry the work thought, if there's the support for it. I am convinced enough to put in the effort.

amtriathlon:
> Instinct suggests that if I myself think that this PR now contains too many moving parts (see previous comment), then others will too. Being fairly new to GC, could I get a quick steer from you please, @amtriathlon ?

I have a similar feeling, I think it would be better to focus on the primary issue you want to solve, without adding loosely related features or references to not merged PRs.

I will add more comments to #3807

Issue #3812, opened 2021-01-31
Tacx Fortius Disconnects


Mickyduck55:
During testing of recent Artifacts which are addressing issues with the Fortius not working in slope mode I have come across an issue where the trainer disconnects from GC.  I am working my way back through builds so I can try to find the point at which this broke.
I am using Windows 10 on an i7 HP EliteBook
GC is in train mode using video chart.
I had the issue using The Argus tour initially but for ease of repeatability I am now using Flat course (TACX Amsterdam city trip).  
ci 3627 disconnects almost immediately
ci 3625 disconnect at 17 mins and 10 mins so varies
I will add to this as I work my way backwards.

At the suggestion of @mattipee  I jumped back to ci3604.. same scenario as above 31 mins NO PROBLEMS

mattipee:
Thanks @Mickyduck55 

@michelrdagenais , tagged you for second opinion, please.

Firstly, here is a list of the relevant builds:

![image](https://user-images.githubusercontent.com/15227658/106389375-dc00ea80-63da-11eb-9784-702ad7183186.png)

The top one is the current PR which Micky says fails immediately - I can't see in the code anything that would cause it. A second opinion there would be appreciated.

My hunch says if we have introduced an issue causing instability over time, then it's likely one of:
- ci.3609 usb_set_altinterface()
- ci.3610 moving the msleep(50)
- ci.3620 request/reply changes

The fact that ci.3604 appears to test ok, supports this.

Unfortunately, the time-to-failure for Micky seems long... at least it makes repeat testing a physical process.

Micky - if ci.3609 fails for you, that could support ci.3609 "altinterface" being an issue - I'll prepare a branch to reintroduce the call on Windows only, and can turn that into a PR if your testing supports that.

Otherwise, thank you for all the testing, particularly the effort in fetching and reinstalling builds to diagnose.

Mickyduck55:
Thanks for the suggestions @mattipee and @michelrdagenais, I will get on ci 3609 tomorrow

mattipee:
The other key bit of information here is that the artefact produced from #3725 (I think) appears to work without disconnecting.

Mickyduck55:
That is correct as far as I can tell having used this Artefact frequently
in the last few weeks.

Before yesterday I had two disconnects one was after over an hour and @Michel
Dagenais <michelrdagenais@gmail.com> suggested that it may be static
buildup and described a solution by grounding the trainer, I have not done
[...]

PR #3813, opened 2021-01-31
Reinstate Fortius usb_set_altinterface call (on WINDOWS only)

mattipee:
A PR to test whether reintroduction resolves https://github.com/GoldenCheetah/GoldenCheetah/issues/3812


mattipee:
@michelrdagenais wrote: "It could be interesting to add a qDebug there to see if that code (using the alt interface) is called frequently. That would confirm that removing it would be problematic."

TODO

amtriathlon:
This can be closed or still can be useful?

michelrdagenais:
There is no clear indication that this is the problem. I cannot test on Windows. Not useful for me.

PR #3814, opened 2021-02-01
Power Spike Filtering

paulj49457:
Looking at the power spike filtering I found that it wouldn't remove negative spikes (dropouts) from the file and so I have now made this optionality selectable in the Fix Power Spikes menu. Also while making this improvement I also noticed that the "Variance (%)" field isn't actually a percentage, its a wattage value above which data values are considered anomalous to their neighbours relative to a 30 second moving average and therefore are considered a spike to be fixed. So I have updated the explanatory text in the dialog to correct the variance wattage value description. And finally, I've provided the option to adjust the the length of the 30 second window averaging from 5secs to 60seconds, so you can now experiment to see the window's affect on power spike fixing.

Capture2

The window length and dropout functions are also available for auto imported files:

Capture3

Selecting and saving the window length & dropouts options modifies the behaviour of the power spikes and power dropout detection in the data window:

Capture1

As a way of example, my original file looked like:

GC1

Using the unmodified Power Spike functionality with Absolute watts set to zero, and wattage "%" at 10, it removes some of the small positive spikes:

GC1-5

So after this change, using the same values and dropout correction enabled, I now get:

GC2

### DIFF
--- a/src/Charts/RideEditor.cpp
+++ b/src/Charts/RideEditor.cpp
@@ -605,21 +605,30 @@ AnomalyDialog::check()
         // get spike config
         double max = appsettings->value(this, GC_DPFS_MAX, "200").toDouble();
         double variance = appsettings->value(this, GC_DPFS_VARIANCE, "20").toDouble();
+        int avgWindowTime = appsettings->value(this, GC_DPFS_WINDOWTIME, "30").toDouble();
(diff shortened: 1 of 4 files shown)


paulj49457:
I have better solution to improving the spike filtering based upon median filtering, I'll submit a new PR with the proposed solution.

PR #3815, opened 2021-02-01
A test branch to diagnose the problem of lost connection on Windows

michelrdagenais:
This is not for merging, just for @Mickyduck55 to test, @mattipee to comment, in order to diagnose #3812 

Debug statements are enabled for all USB communication errors from the Fortius. Then, USB open and write commands are retried. With this, we should see messages for all USB errors. If the errors are minor and intermittent, a simple retry may save the day. By running this test branch, we should get a much clearer picture of what the problem is, hopefully.


mattipee:
Thank you @michelrdagenais .

I can now test on Windows, also. So far, no issues in the tests that I have done, but I will install this artefact and continue.

michelrdagenais:
This branch now contains the simspeed code, PR 3809, and debug statements. @Mickyduck55 please test and report what you get. It may solve the problems, since it retries operations that would otherwise have caused the Fortius code to quit. However, by seeing where errors are reported, and possibly recovered by retrying, we should get a better understanding of the problem.

Mickyduck55:
@michelrdagenais thank you have I installed ready to test

Mickyduck55:
@michelrdagenais @mattipee 
no sweating today fell over twice within 1 km and once within 2.
Error state from used view.. video continues to play but all telemetry stuck.
After the first error I was able to just stop and restart using the tool bar in GC
After second and third error had to physically unplug USB.
1.txt is error to first disconnect
3.txt is all three in the same file
Hope this helps
I will say it again Thank you for all the effort
[3.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/5918017/3.txt)
[1.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/5918019/1.txt)

michelrdagenais:
Humm, the connection was lost after about 3 minutes and then all the retries simply failed. From the message, the windows driver or the device gets in an incorrect state. This is consistent with the fact that you have to restart everything. Since you have this problem with several versions, but much more often with this version (simspeed), this is a property that we can use. I see two possibilities:

- The commands sent or the timing is different with this version. We could extend the debugging info will all commands sent and replies read, with timestamp. Then, we could compare the good and bad runs, for instance the commands sent just before the lost connection.
- This specific version has things placed differently in memory and simply sensitizes a software or hardware bug/fault. Running a memory check application on that PC or running the same GC version on a different computer, but with the same Fortius trainer, would help rule out that case.

You wrote that you had another PC, did you try this GC version with the different PC?

Mickyduck55:
Hi @michelrdagenais 
just finished setting up and testing the other PC so fresh install, same version, 1.txt same USB cable then 2&3 different USB cable
3 failures within a minute
[2.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/5919550/2.txt)
[3.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/5919551/3.txt)
[1.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/5919552/1.txt)

Mickyduck55:
The only other variable I could try is using Susan's head unit... she has an iMagic so no idea if that would work but head unit is the same

michelrdagenais:
What version are you using? It looks like the latest merge, adding simspeed, did not build. The error appears to be in an imported package, not in the GC code. Where are you taking these "artefact" builds? I never installed such builds since I compile the different branches myself.

The next step would be to produce two versions, one without simspeed but with all the debugging (all messages sent and received by USB), and the other with simspeed and all the debugging. If the first works fairly well and the second fails quickly, we will try to understand the difference between the two.

Mickyduck55:
@Michel Dagenais <michelrdagenais@gmail.com>

https://ci.appveyor.com/project/Joern-R/goldencheetah-knhd8/builds/37580792/artifacts

Mickyduck55:
@michelrdagenais @mattipee 
If the you keep building I will keep testing  There is something properly and reproducibly not correct.. I have no idea what it is but recent tests to failure require very little effort from me
Thanks

mattipee:
> What version are you using? It looks like the latest merge, adding simspeed, did not build. The error appears to be in an imported package, not in the GC code. Where are you taking these "artefact" builds? I never installed such builds since I compile the different branches myself.

@michelrdagenais

The AppVeyor (Windows) build did build ok shortly after you committed.

Your link is to **ci.3637** which is correct, @Mickyduck55 

The Travis CI (Linux/Mac) build I don't think did first time, but looks like build completed within the last hour.

See the links under "All checks have passed" at the bottom of this PR and click "Show all checks", or scroll to the bottom of https://github.com/GoldenCheetah/GoldenCheetah/ for similar links. Artefacts themselves are available through there.

Mickyduck55:
@mattipee <tunabiscuits796@gmail.com>   TFFT...

Mickyduck55:
3638 building.. something to do tomorrow, thanks

michelrdagenais:
Actually, there will be 2 builds. The current one with the master branch plus full debugging of the Fortius connection (all read and writes). The upcoming one will have the intensity3 branch merged. The current hypothesis is that without intensity3 there should be little or no problem. This will provide us with a baseline for comparison. Then, with intensity3, there should be problems within a few minutes. We will see the commands sent (and their timing) up to when the problem occurs. You will have to train twice, you will get in excellent shape soon if this continues!

> 3638 building.. something to do tomorrow, thanks

PR #3818, opened 2021-02-05
trainer capabilities request fix

### LINKED ISSUE #3784: Brake fault with Tacx Flow Smart connected via FE-C/ANT+
Similar to Issue #3319 but hopefully I can add some further info. Attached is screenshot showing the Ride output of a 30 minute run using a Windows laptop, Windows 10, GC build 4001 developer build, ANT+ connection with a Tacx Flow Smart trainer, the run was carried out using full screen video and Zwerschstubental.gpx, .avi and . rlv downloaded from http://rlv-alucard.de/.
Everything works well except 'Brake fault' kept coming up on screen at which point the resistance of the trainer falls and power etc. drops with it. Every few seconds the fault would disappear and the trainer resistance would rise together with power for a few seconds then repeat the sequence. The effect can be clearly seen in the Ride output screenshot.

Screenshot (1)

The entire hardware setup works fine with Rouvy so it seems to me that the issue lies within the GC software and the way it is sending the resistance requirement to the trainer via ANT+. I shall attempt to investigate further but welcome any help that others can give.

IainMacdonald:
I believe that this fixes #3784.

Issue I had was I updated the firmware on my Tacx S Smart (to 3.3.40/1.1.6) and the brake/resistance stopped working. Spent a while tracking ant+ messages and the request for trainer capabilities never generated a reply. 

Found this code: https://github.com/alex-hhh/TrainerControl/blob/2c908746e1c193a1e121f8c84d5c0266f9bd5cd1/src/AntStick.h#L232 which sets the 'transmission count to 4' so tried it in GC and it works. I've been using it for several days now.

Tested on linux (Ubuntu 20.04.2 LTS with QT 5.15.2).

### DIFF
--- a/src/ANT/ANTMessage.cpp
+++ b/src/ANT/ANTMessage.cpp
@@ -1141,7 +1141,7 @@ ANTMessage ANTMessage::fecRequestCapabilities(const uint8_t channel)
                       FITNESS_EQUIPMENT_REQUEST_DATA_PAGE,        // data page request
                       0xFF, 0xFF,  // reserved
                       0xFF, 0xFF,  // descriptors
-                      0x01,        // requested transmission response (1 time only)
+                      0x04,        // requested transmission response
                       FITNESS_EQUIPMENT_TRAINER_CAPABILITIES_PAGE,        // requested page
                       0x01);       // request data page
 }


amtriathlon:
Good finding, I will merge this, thanks!

PR #3820, opened 2021-02-07
Expand new activity lines on auto import & dynamic resizing of Calendar text within Ride Navigator 

paulj49457:
This change addresses a number of issues with the activity presentation and updates within the Ride Navigator. I did think of trying to separate them into more than one pull request, but they are all intertwined functions.

**#3074 - Activity list row height is not updated when data changes:**
This issue is within the batch importation (RideImportWixard.cpp), if you import too many files and only a few are new files, then the tableWidget->rowCount() is always going to be greater than 20, see code below from RideImportWixard.cpp:

 _// to the "clean" activities folder
 context->athlete->addRide(QFileInfo(tmpActivitiesFulltarget).fileName(),
                                 tableWidget->rowCount() < 20 ? true : false, // don't signal if mass importing
                                 true, true);                    // file is available only in /tmpActivities, so use this one please_

Therefore the signalling code below in RideCache.cpp ever gets executed, and there is never a ridesAdded signal emitted:

_if (dosignal) context->notifyRideAdded(last); // here so emitted BEFORE rideSelected is emitted!_

Additionally Ride Navigator only listens for rideSelected, and not rideAdded, so it never received rideAdded, so the table was never refreshed to recalculate & update the table list entries.

**Note**: The above changes to RideImportWixard.cpp in this pull request should be considered a patch, as it just ensures the first 20 processed updates cause a rideAdded signal to be emitted, but ideally it would be better for only the last of the valid updates in a mass importation to raise a rideAdded and rideSelected signal. 
**_This change has now been implemented in RideImportWixard.cpp by pull request #3770 Auto import filtering and file import exclusion, by slightly restructuring phase 3 of the importation processing, ensuring rideAdded & rideSelected is only notified for the last successful activity._** 

**Addition of rideAdded signal to RideNavigator**

Adding rideAdded signal listening to RideNavigator.ccp ensures the table is refreshed to recalculate & update the table list entries, therefore the height and positioning of the activities and calendar text is now correctly displayed.

**Dynamic resizing of calendar text** 
This change provides dynamic resizing of the calendar text from a single line up to three lines, depending on the calendar field contents and the width of the Ride Navigator field. Dragging the separator between the Ride Navigator and the main window allows the calendar text to occupy a single line if only short content calendar items are selected for display.

### DIFF
--- a/src/Gui/RideImportWizard.cpp
+++ b/src/Gui/RideImportWizard.cpp
@@ -1000,7 +1000,7 @@ RideImportWizard::abortClicked()
     }
 
     QChar zero = QLatin1Char ( '0' );
-
+    int notifiableRides = 0;
 
     // Saving now - process the files one-by-one
     for (int i=0; i< filenames.count(); i++) {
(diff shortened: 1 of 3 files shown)


amtriathlon:
> This change addresses a number of issues with the activity presentation and updates within the Ride Navigator. I did think of trying to separate them into more than one pull request, but they are all intertwined functions.
> 
> **#3074 - Activity list row height is not updated when data changes:**

Issue #3074 is independent of activity import, see https://github.com/GoldenCheetah/GoldenCheetah/issues/3074#issuecomment-768671201

You are welcome to propose a solution for that issue, provided it effectively fixes it (I rejected 2 previous proposal which don't fix the issue), and it is not mixed with unrelated changes or unnecessary reformats or refactors.

PR #3823, opened 2021-02-13
Workout: LiveMapWidget in video layout starts with the zoom specified in video-layout.xml

peret2000:
I noticed zoom configured in xml file is ignored, and code is always starting with a value of 16 for thee zoom of thee live map widget.
This change uses <Zoom> field of the xml; it was already being parsed but not used, so using it is simple and straightforward


amtriathlon:
Good catch, thanks!

PR #3832, opened 2021-02-26
Update FitRideFile.cpp

dconka:
Add missing SIGMA SPORT Products with consistant format.

### DIFF
--- a/src/FileIO/FitRideFile.cpp
+++ b/src/FileIO/FitRideFile.cpp
@@ -503,7 +503,7 @@ struct FitFileReaderState
                 case 44: return "SIGMA ROX 12.0";
                 case 45: return "SIGMA iD.FREE";
                 case 46: return "SIGMA iD.TRI";
-                default: return "SIGMA SPORT Device";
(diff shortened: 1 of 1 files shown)


src/FileIO/FitRideFile.cpp
```diff
@@ -493,10 +493,17 @@ struct FitFileReaderState
...
-                default: return "Sigmasport ROX";
+                case 15: return "SIGMA ROX 10.0";
+                case 18: return "SIGMA ROX 7.0";
+                case 41: return "SIGMA ROX 11.0";
+                case 42: return "SIGMA iD.RUN";
+                case 43: return "SIGMA iD.RUN HR";
+                case 44: return "SIGMA ROX 12.0";
+                case 45: return "SIGMA iD.FREE";
+                case 46: return "SIGMA iD.TRI";
+                default: return "SIGMA SPORT Device";
```
amtriathlon, line 506: “Device” word is redundant, it would be better to include the product Id in default case, like Stages et al.

amtriathlon:
Thank you!

PR #3842, opened 2021-03-12
Use QtBluetooth enums from their namespace

andreasbuhr:
For increased type safety, some enums have been changed to
scoped enums in Qt 6.2, see
https://codereview.qt-project.org/c/qt/qtconnectivity/+/337069
https://codereview.qt-project.org/c/qt/qtconnectivity/+/336678
This patch adapts GoldenCheetah to this change.
Since C++11, enums inject their symbols in both their own
and their parent namespace, so this patch can be merged right
now.


andreasbuhr:
Hi, could we merge this?

andreasbuhr:
This could also be labeled as Qt6 Refactor

amtriathlon:
Thank you!

PR #3848, opened 2021-03-23
New check box added to Map menu to display shaded route (60sec power intervals)

paulj49457:
i) New check box added to Map menu to allow the user to select the display of the shaded route (avg watts per 60s interval). This switches the displayed route between the complete activity ("the red line with thick yellow border") to the segmented view of the shaded route, colours dependent on wattage.

This uncovered another minor display issue with the segmented view:

ii) Fixed the display of the shaded route (when selected) on the map, as previously a section of route too short to be a 60 second interval wasn't drawn, so now any route segment too short at the end of the activity is now included, although in a shorter interval element with appropriate calculation of power in the smaller time segment. Note: The thicker yellow border has been disabled in this mode as the mixing of the various power "colours" and opacity the line segments aren't always visible.

Note: Apologies that there are a few unintended/unwanted whitespace changes, these are not picked up by my VisualStudio SourceTree tools, but once uploaded to GitHub these are detected, these are due to extracting the changes from an earlier rejected pull request.

### DIFF
--- a/src/Charts/RideMapWindow.cpp
+++ b/src/Charts/RideMapWindow.cpp
@@ -471,7 +471,7 @@ void RideMapWindow::createHtml()
             "routeYellow.on('mousemove', function(event) { webBridge.hoverPath(event.latlng.lat, event.latlng.lng); });\n"
 
             "}\n").arg(styleoptions == "" ? "#FFFF00" : GColor(CPLOTMARKER).name())
-                  .arg(styleoptions == "" ? 0.4 : 1.0);
+            .arg(styleoptions == "" ? 0.4 : 1.0);
     }
     else if (mapCombo->currentIndex() == GOOGLE) {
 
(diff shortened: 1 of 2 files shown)


amtriathlon:
You can see the changes in your fork before to make a Pull Request.
OTOH git is very powerful: you can use `git diff` to see the changes before commit, `git show` after commit, `git commit --amend` to fix the las commit (or `git reset` to "uncommit") and `git rebase -i` to squash several commits in one.

paulj49457:
Thank you, I'll was using compare changes with Visual Studio, so not sure whether I have an ignore whitespace setting. I was surprised that a number of whitespace only changes made it through, I go back and sort them out and re-push the files to update the pull request.

amtriathlon:
There are still unrelated changes in this PR like member variable renaming, comments removed and other unnecessary editions. I think contribution guidelines are clear:
```
Focus on a single bug or feature at a time, and do not submit style formatting or refactoring changes in the original patch.

Just submit the changes that fix the bug or tweak the code with a clear and concise description.
```
Besides, since this is a cosmetic feature nobody else has requested, you should check default behavior is unchanged.

PR #3851, opened 2021-03-26
New Map Chart Menu options 

paulj49457:
### Justification

This change was created out of a discussion in the developers group:

https://groups.google.com/g/golden-cheetah-developers/c/_I8YbKlja44/m/bByJyQAHBQAJ

Please see  text:

_Ale Martinez  Jan 8, 2021, 4:04:16 PM  to golden-cheetah-developers
Histogram and CP charts has Power Shading/Shade Zones as a chart option, it seems reasonable to make that optional in Map chart too.
The clean up also seems a good idea to me._
 
 
 
### The Change:

Note: The OptionsStyle clean up mentioned in the discussion thread is in a separate PR.

I have extended the change in the dev discussion to provide the following Map Chart options for the user:

_Note: Default GC behaviour remains unchanged and that is to display:_

i) The route using the Power Shading Zones in 60sec power intervals (when power data is available), otherwise the route is displayed as a red line.
ii) A background to the route as thick yellow line.

This change allows the user to independently via the new chart checkboxes to:

i) ignore any Power Shading Zones (if power data is available) and always draw the red line **(Hide Shaded Zones)**
i) remove the route background yellow line **(Hide Yellow Line)**

Capture

While testing the above changes it uncovered a minor display issue with the segmented view:

Run Power Map Gap

Currently a section of route which is too short to be a 60 second interval wasn't drawn, so this is now fixed and any route segment too short at the end of the activity is now included, although in a shorter interval element with a power calculation based upon the smaller time segment.

### DIFF
--- a/src/Charts/RideMapWindow.cpp
+++ b/src/Charts/RideMapWindow.cpp
@@ -69,12 +69,18 @@ RideMapWindow::RideMapWindow(Context *context, int mapType) : GcChartWindow(cont
 
     showMarkersCk = new QCheckBox();
     showFullPlotCk = new QCheckBox();
+    hideShadedZonesCk = new QCheckBox();
+    hideShadedZonesCk->setChecked(false);
+    hideYellowLineCk = new QCheckBox();
+    hideYellowLineCk->setChecked(false);
     showInt = new QCheckBox();
     showInt->setChecked(true);
 
     commonLayout->addRow(new QLabel(tr("Map")), mapCombo);
     commonLayout->addRow(new QLabel(tr("Show Markers")), showMarkersCk);
     commonLayout->addRow(new QLabel(tr("Show Full Plot")), showFullPlotCk);
+    commonLayout->addRow(new QLabel(tr("Hide Shaded Zones")), hideShadedZonesCk);
+    commonLayout->addRow(new QLabel(tr("Hide Yellow Line")), hideYellowLineCk);
     commonLayout->addRow(new QLabel(tr("Show Intervals Overlay")), showInt);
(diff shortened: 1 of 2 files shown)


amtriathlon:
> Note: The OptionsStyle clean up mentioned in the discussion thread is in a separate PR.

I merged that PR so there is a merge conflict now in this one which needs to be fixed, the way to avoid that is to put non-functional changes in the same PR but in a separate commit so they can be evaluated independently and, eventually cherry-picked.

> I have extended the change in the dev discussion to provide the following Map Chart options for the user:
> 
> i) The option to display the Power Shading Zones in 60sec power intervals (when power data is available) or red line route (the default)

As I commented before, I think it is a bad idea to change default settings unless there is a very compelling reason, which doesn't seem to be this case.

paulj49457:
**I merged that PR so there is a merge conflict now in this one which needs to be fixed, the way to avoid that is to put non-functional changes in the same PR but in a separate commit so they can be evaluated independently and, eventually cherry-picked. git rebase -i is a useful tool.**

I have updated the PR and description and removed any merge conflicts, what remains is the minimal functional change.

**As I commented before, I think it is a bad idea to change default settings unless there is a very compelling reason, which doesn't seem to be this case. The reason for this is to prevent current users from wondering "Why is the path no longer colored by power?"**

I intended to follow the guidance provided, but I made a mistake.

I have now updated the PR and the default GC behaviour is retained, and I've updated the dialog labels and code to make it clearer that the options modify the default when selected.

amtriathlon:
Thank you Paul for your contributions, I will add you to contributors list.

PR #3852, opened 2021-03-26
Compiler warning in MetricSelect.cpp

paulj49457:
Fixes:

Gui\MetricSelect.cpp(67): warning C4554:  '&': check operator precedence for possible error; use parentheses to clarify


amtriathlon:
It seems to be harmless at this moment, but it looks like an error, so better lets fix it. Thanks!

PR #3853, opened 2021-03-28
New Median Power Spike filtering dialog

paulj49457:
### Story & Justification:
I've been getting more than my fair share of power spikes when using GC in training mode with my Kickr Core. So using the power spike filtering within GC to correct my files I found that it wasn't effective on spikes of more than a single data point, as the algorithm uses the average of the two adjacent neighbours to fix the erroneous data point. When the spike is a number of points wide this just reduces the spike and subsequent attempts with the filter never completely removes the spike.

The issue of spikes was also discussed here:
https://groups.google.com/g/golden-cheetah-users/c/_bxIzhFPVWA/m/Q5PizsveBAAJ

### Proposal:
Therefore I have created a new dialog "Fix Power Spikes Median" which as the name suggests uses a median window method on the data samples as the median filter is robust to local outliers, and using a suitable window will remove multi sample spikes completely, whilst preserving sharp edges/changes in power output, see:

https://www.gnu.org/software/gsl/doc/html/filter.html?highlight=median#standard-median-filter

The reason for proposing a separate new dialog is that it retains the existing GC functionality, so minimising impact, and will provide the opportunity for the two approaches to be compared.

### Change

So a demonstration of the median filtering, starting with the typical case:

Normal Spike case

after running the median filter with the default values:

Normal Spike case - Median Fixed

So now a demonstration of the median filtering on my worst case (and its bad), and no I cannot suddenly generate 2KW+

Big Spike case

after running the median filter with the default values:

Big Spike case - Median Fixed

The Edit menu now has an additional entry "Fix Power Spikes Median...":

Median Dialog

And the import processing has an additional entry "Fix Power Spikes Median":

Import

The new dialog looks like:

Median

The dialog takes the following parameters:

> 

`Median` Window Size` - this defines the number of neighbouring points used to determine a median value; the window size is always odd to ensure we have a central median value.

`Variance` Watts` - Determines the threshold beyond which a data point will be fixed, if the difference between the data point value and the median value exceeds this parameter.

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -181,6 +181,8 @@
 #define GC_DPFG_STOP                    "<global-general>dataprocess/fixgaps/stop"
 #define GC_DPFS_MAX                     "<global-general>dataprocess/fixspikes/max"
 #define GC_DPFS_VARIANCE                "<global-general>dataprocess/fixspikes/variance"
+#define GC_DPFSM_MED_WIN                "<global-general>dataprocess/fixspikesmed/medwindow"
+#define GC_DPFSM_MED_VARIANCE           "<global-general>dataprocess/fixspikesmed/medvariance"
 #define GC_DPTA                         "<global-general>dataprocess/torqueadjust/adjustment"
 #define GC_DPPA                         "<global-general>dataprocess/poweradjust/adjustment"
 #define GC_DPPA_ABS                     "<global-general>dataprocess/poweradjust/adjustment_abs"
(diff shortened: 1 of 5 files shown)


amtriathlon:
If you have lot of spikes in Train mode it is likely due to a configuration or interference problem, it is better to fix the cause than to consequences, see https://github.com/GoldenCheetah/GoldenCheetah/wiki/ANT--and-BTLE-Troubleshooting#train-mode-fe-c-speed-and-power-spikes
Since there is already a Fix Spikes DP you could add an alternative algorithm or use a https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Special-Topics_Custom-Data-Processors-in-Python

PR #3855, opened 2021-03-31
Facility to trace all the USB communications

michelrdagenais:
A new logging category is added gc.usb, inactive by default, to log all the USB transfers between GC and the USB trainers. It can be activated by changing the logging filter with the --debug-rules option.

### DIFF
--- a/src/Core/main.cpp
+++ b/src/Core/main.cpp
@@ -243,8 +243,11 @@ main(int argc, char *argv[])
     bool debug = false;
     QString debugFormat = QString("[%{time h:mm:ss.zzz}] %{type}: %{message}");
 #endif
+
+    // By default, print messages from all categories, but not those from
+    // specialised categories like qt.* or gc.* which can be quite verbose
+    QString debugRules = QString("*.debug=true;gc.*.debug=false;qt.*.debug=false");
     QString debugFile = NULL;
-    QString debugRules = QString("*.debug=true;qt.*.debug=false");
 
     bool server = false;
     nogui = false;
--- a/src/Train/LibUsb.cpp
+++ b/src/Train/LibUsb.cpp
@@ -28,6 +28,9 @@
 #include "Settings.h"
 #include "Context.h"
 
+Q_DECLARE_LOGGING_CATEGORY(gcUsb)
+Q_LOGGING_CATEGORY(gcUsb, "gc.usb")
+
 #ifdef LIBUSB_V_1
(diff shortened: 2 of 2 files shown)


amtriathlon:
Hi Michel, I really like this feature and I think we could use to replace the several conditionally compiled “printd” we have to allow selective debug enabling at runtime.
My concern is performance penalty when debug is not enabled, do you have an evaluation?

michelrdagenais:
I believe that the logging in Qt was built for that purpose, low overhead when not enabled, and thus to be widely used in code. It is certainly used in that way within the Qt library. If you enable the qt.*.debug logging categories, you will be swamped with messages. There are thus numerous non enabled qCDebug statements all throughout the Qt code.

The code generated for a qCDebug(gcUsb)  statement is something like:

If (gcUsb().isDebugEnabled()) { ... do all the formatting and printing ... }

It could have been even lower overhead as a test on a static boolean, defined for the category. I suspect that they wanted to have an accessor method with locking. I wrote a simple test, adding a loop with qCDebug statements:

            qDebug() << "First debug statement";
            qDebug() << "Second debug statement";
            for(int i = 0; i < 1000000; i++) {
                qCDebug(gcTest) << "This is a test";
            }
            qDebug() << "Third debug statement";

The result is the following:

[23:49:57.715] debug: First debug statement (Train/TrainSidebar.cpp:1645)
[23:49:57.715] debug: Second debug statement (Train/TrainSidebar.cpp:1646)
[23:49:57.765] debug: Third debug statement (Train/TrainSidebar.cpp:1650)

The difference between the first and second statement is just the active debug statement, is smaller than 1ms and not measureable. The difference between the second and third statement is 1M non active debug statements and this takes 50ms, or 50ns per inactive debug statement.

IMO, this is certainly negligible for LibUsb when you have at most one event each several ms. If there is a situation in another module where the frequency is such that 50ns is too much, it would be possible to define a boolean variable initialized from the status of the logging category. A test on a simple boolean variable would then be even much lower overhead.

There is some memory overhead in having all those debug statements, but this would only be a problem for embedded systems with extremely limited memory. In addition, it is possible to define QT_NO_DEBUG_OUTPUT at compilation time, to get rid of all the debug statements.

michelrdagenais:
The overhead reported was without optimization. With -O3 like the release builds, the time is rather 6 to 8ms for 1M inactive statements, or about 7ns per inactive qCDebug statement.

PR #3860, opened 2021-04-04
Read EPOC from Garmin FIT file

axl13:
**Exercise Load**
The load of an exercise session is a numeric score indicating the degree of its impact on your body. It’s based on estimated EPOC. Garmin measures load for individual exercise sessions and also compiles load measurements for calculating weekly Training Load and Training Status.

NOTE: This is undocumented, tested on my Garmin generated files, could work for other's devices licensing [Firstbeat metrics](https://www.firstbeat.com/en/science-and-physiology/epoc-and-training-effect/), i.e Polar.


amtriathlon:
It doesn’t work with FIT files generated by Garmin 935 in my tests.
Can you attach a sample file?

axl13:
Thank you for looking in to this. Here's my files. Do you have the EPOC displayed in the "Training Effect" on Garmin Connect?
(also please check the under the "Extra" tab in the GC "Details"

Thanks!

[6569491078.zip](https://github.com/GoldenCheetah/GoldenCheetah/files/6283263/6569491078.zip)
Screen Shot 2021-04-09 at 6 58 06 AM
Screen Shot 2021-04-09 at 7 01 09 AM

amtriathlon:
> Thank you for looking in to this. Here's my files. Do you have the EPOC displayed in the "Training Effect" on Garmin Connect?

No, that's the problem.

It looks like this is a feature of new Garmin devices, probably it is also available in 945, but not in 935.

I will merge this, but it would be interesting to document it in the wiki, ideally with a list of supported devices.

Thanks for your contribution!

PR #3867, opened 2021-04-11
Add Zwift intervals to Erg intervals + Add intervals to FIT export

comediebenji:
Hey, I just add the possibility to have Zwo training sessions beeing automatically divided into intervals (laps) in order to be able to analyze the phases of the training afterwards. (and in the future to add cadence target for each interval: i'm working on that, pull request soon normally).

The intervals created during workout are now integrated into the FIT export (which I personally use). In addition, I added the possibility of having intervals of a duration of 1s in the RideDB.json and in the summary (just to be synchronous with the JSON files of the saved activities).

### DIFF
--- a/src/Core/RideItem.cpp
+++ b/src/Core/RideItem.cpp
@@ -927,8 +927,10 @@ RideItem::updateIntervals()
            (interval->start <= begin->secs && (interval->stop+f->recIntSecs()) >= end->secs))))
              continue;
 
-        // skip empty backward intervals
-        if (interval->start >= interval->stop) continue;
+        // skip empty backward intervals --> No because FIT intervals lenght (in sec) is "interval->stop - interval->start + 1" and start/stops are defined in secs: so what append if start = 3s and stop=3.8s ?
+        // So skip empty backward intervals only if start==stop==0;
+        if (interval->start == 0 || interval->stop == 0) continue;
+        qDebug() << "RideItem: " + QString::number(interval->start);
 
         // create a new interval item
         const int seq = count; // if passed directly, it could be incremented BEFORE being evaluated for the sequence arg as arg eval order is undefined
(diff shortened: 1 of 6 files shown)


amtriathlon:
First, thanks for your contribution. Second, some comments:
1) This seems 2 completely independent features, please make 2 separate pull request
2) At first sight creation of laps from zwo files seems to be ok, you can check in qwkcode and erg export.
3) For FitRideFile to access intervals you can make it a friend class like in: https://github.com/GoldenCheetah/GoldenCheetah/blob/e9b192d84a2461e01558bfbbd54122789bb533f3/src/FileIO/RideFile.h#L183, but don't make any other change to RideFile class.
4) GoldenCheetah intervals can overlap, this needs to be tested on fit export.

comediebenji:
Hello,
1. Okay, I'll do it like this. Sorry, I'm going to do two pull requests: i'm just a newcomer on Github.
2. It seems that if we modify the ZWO workout with qwkcode and export it to ERG, it keeps the original laps but starts with lap "0" whereas with Zwo workout, Laps start from "1"; after the workout with the ERG file, 1 lap is missing, and their duration does not match as well (see attached pictures)
pb
3. Okay, will change that
4. How is it possible (to reproduce that phenomemon)? I tested with a workout and add during it a manual Lap, and it's Okay with the FIT export.

amtriathlon:
Hi, for 1) you can see https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/proposing-changes-to-your-work-with-pull-requests

WRT to 2) I don't know, but once it is on a separate PR we can test and compare, ideally the behavior should match.

For 4) you can add interval manually in Activity chart. Since the export FIT feature is a general one, it needs to behave well with free laps, not only with sequential/contiguous like the ones generated by Train or Devices. I am not saying it doesn't work, only it is something to test.

amtriathlon:
>I'm going to do two pull requests

Then let’s close this one.

Please check https://github.com/GoldenCheetah/GoldenCheetah/wiki/Guidelines-for-submitting-a-patch

PR #3872, opened 2021-04-18
Correcting bug with quote character in workout Textevents and ZWO export

comediebenji:
When adding a Zwo workout, modify it (or not) and save it with .zwo format, if some Textevents have "some words" inside quotes, they were not replaced with XML comptible characters "\&quot\;".

### DIFF
--- a/src/Train/ErgFile.cpp
+++ b/src/Train/ErgFile.cpp
@@ -1520,14 +1520,11 @@ ErgFile::save(QStringList &errors)
                                   << "PowerHigh=\"" <<sections[i].end/CP << "\"";
                 if (Texts.count() > 0) {
                     out << ">\n";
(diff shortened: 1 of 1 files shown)


amtriathlon:
Good catch, but since this is a common problem with strings included in xml files, it would be better to use the available utility function for that task: https://github.com/GoldenCheetah/GoldenCheetah/blob/696a04e8a0d2175d4bb5f6de8c628026c7049924/src/Core/Utils.cpp#L144

comediebenji:
Ok, I will do the change

amtriathlon:
And the reverse function needs to be used on import

comediebenji:
It seems to be already the case, with the Qxml-InputSource-SimpleReader

Issue #3760, opened 2021-01-06
Device metadata is not updated on change, activity save or application restart


tjeerd-tx:
On GC Windows when changing metadata trough the details tab it is only updated after the activity was saved and the metadata field was changed again. But than it is changed to the previous value. This way for instance the column in Activities or GC.activityMetrics() always show the previous value that was entered for a metadata field.

![image](https://user-images.githubusercontent.com/61319032/103800478-c4894880-504c-11eb-9c94-c524e964e10e.png)

liversedge:
This is an historic artefact where the devicetype was a first class field and changing it would have cause a significant issue.
Now it is no longer so crucial. We should either prohibit editing it (make the Device field read only) or update the code to apply changes to the devicetype field.

This only applies to the Device field btw.

amtriathlon:
> This is an historic artefact where the devicetype was a first class field and changing it would have cause a significant issue.
> Now it is no longer so crucial. We should either prohibit editing it (make the Device field read only) or update the code to apply changes to the devicetype field.
> 
> This only applies to the Device field btw.

I like the alternative to make Device field to behave like a standard field, this would allow to remove other limitations s.t. the use of set for bulk update in https://github.com/GoldenCheetah/GoldenCheetah/blob/55e9cdb444035647a2623a23a5562b41a40afe22/src/Core/DataFilter.cpp#L2491

Looking at RideFile, [deviceType_](https://github.com/GoldenCheetah/GoldenCheetah/blob/55e9cdb444035647a2623a23a5562b41a40afe22/src/FileIO/RideFile.h#L430) is a private member, so we could remove it and change [deviceType/setDeviceType](https://github.com/GoldenCheetah/GoldenCheetah/blob/55e9cdb444035647a2623a23a5562b41a40afe22/src/FileIO/RideFile.h#L312) to access/set "Device" field with "unknown" default.

PR #3881, opened 2021-04-28
Fixes: "Device" tag not tracking changes to the DeviceType field

paulj49457:
This fix only applies to updating the "Device" tag field and it ensures the "Device" tag tracks the DeviceType field fixing #3760

The "device" which recorded the activity appears in two separate locations within the activity json file, and different windows/fields in GC display either one or the other.

"DEVICETYPE":"Garmin 1234 ",
"TAGS":{
        .................
       "Device":"Garmin 1234 ",
        .................


amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/9ff6d6ab9742f3bcc8b482e6fe25fcc8f043ef74

PR #3887, opened 2021-05-11
Additional Median Option in "Fix Power Spikes" Data Processor

paulj49457:
### Justification:
I've been getting power spikes when using GC in training mode with my Kickr Core, and I have tried to investigate this issue but without success so far. So using the power spike filtering within GC to correct my files I found that it wasn't effective on spikes of more than a single data point, as the algorithm uses the average of the two adjacent neighbours to fix the erroneous data point. When the spike is a number of points wide this just reduces the spike and subsequent attempts with the filter never completely removes the spike.

### Proposal:
Therefore I have added another algorithm option in "Fix Power Spikes" which uses a median window approach on the data samples as this type of filtering is robust to local outliers, and using a suitable window size will remove multi sample spikes and dropouts completely, whilst preserving sharp edges/changes in power output, see:

https://www.gnu.org/software/gsl/doc/html/filter.html?highlight=median#standard-median-filter

### Change

So a demonstration of the median filtering, starting with my unprocessed file

Capture1

Using the existing smoothing/interpolating filtering in "Fix Power Spikes" 

Capture2

Now using the new the median filter on the unprocessed file with the default values:

Capture2

The updated  "Fix Power Spikes " dialog is shown below and contains the description of the new algorithm and parameters.

Capture4

And the import processing for the "Fix Power Spikes " row includes the new options:

Capture4

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -181,6 +181,8 @@
 #define GC_DPFG_STOP                    "<global-general>dataprocess/fixgaps/stop"
 #define GC_DPFS_MAX                     "<global-general>dataprocess/fixspikes/max"
 #define GC_DPFS_VARIANCE                "<global-general>dataprocess/fixspikes/variance"
+#define GC_DPFS_MEDWINSIZ               "<global-general>dataprocess/fixspikes/medwinsiz"
+#define GC_DPFS_MEDALGO                 "<global-general>dataprocess/fixspikes/medalgo"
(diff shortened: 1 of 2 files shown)


amtriathlon:
Thanks for your contribution Paul, I will merge it.
If you want to update the corresponding wiki entry, it is this one: https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Special-Topics_Activity-Processing#edit-tool-fix-power-spikes

PR #3890, opened 2021-05-13
RideNavigator - Use uniform row height

### LINKED ISSUE #3074: Activity list row height is not updated when data changes

amtriathlon:
Variable row height, depending on activity calendar text being empty or not,
provokes refresh issues when calendar text changes from empty to not, or
viceversa.
Instead to try to solve this issue with specific code, I think it is
simpler, and more regular from UX point of view, to have a uniform row height.
After this change row height depends on metadata config only:
1) a new RideMetadata method is introduced (hasCalendarText) to check if
   Calendar Text can be non empty, i.e. if some field has diary checked.
2) RideNavigator uses this method when config changes to set hasCalendarText
   member used to determine row height for all activities.

Fixes #3074

### DIFF
--- a/src/Gui/RideNavigator.cpp
+++ b/src/Gui/RideNavigator.cpp
@@ -49,6 +49,7 @@ RideNavigator::RideNavigator(Context *context, bool mainwindow) : GcChartWindow(
     fontHeight = QFontMetrics(QFont()).height();
     reverseColor = GlobalContext::context()->colorEngine->reverseColor;
     currentItem = NULL;
+    hasCalendarText = false;
 
     init = false;
 
@@ -168,6 +169,7 @@ RideNavigator::configChanged(qint32 state)
 {
     fontHeight = QFontMetrics(QFont()).height();
     reverseColor = GlobalContext::context()->colorEngine->reverseColor;
+    hasCalendarText = GlobalContext::context()->rideMetadata->hasCalendarText();
 
     // hide ride list scroll bar ?
 #ifndef Q_OS_MAC
@@ -1074,7 +1076,7 @@ QSize NavigatorCellDelegate::sizeHint(const QStyleOptionViewItem & /*option*/, c
     QSize s;
 
     if (rideNavigator->groupByModel->mapToSource(rideNavigator->sortModel->mapToSource(index)) != QModelIndex() &&
-        rideNavigator->groupByModel->data(rideNavigator->sortModel->mapToSource(index), Qt::UserRole).toString() != "") {
+        rideNavigator->hasCalendarText) {
         s.setHeight((rideNavigator->fontHeight+2) * 4);
     } else s.setHeight(rideNavigator->fontHeight + 2);
     return s;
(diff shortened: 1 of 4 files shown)


paulj49457:
Does this address the issue of mass importation when RideNavigator isn't notified of new activities when more than twenty are imported at the same (see issue 2 in PR #3884) ? Maybe there is another mechanism which this fix picks uses to refresh the table in this case? If so, I'd be interested to understand how it works, otherwise this looks a simple solution to the problem(s).

amtriathlon:
> Does this address the issue of mass importation when RideNavigator isn't notified of new activities when more than twenty are imported

Yes, but more testing is welcome. You can try in your build using “command line instructions” below, or install the binary generated by AppVeyor.

Additionally it should solve a 3rd case: when the change in calendar text is provoked by set/unset filter formula to other activities different from the currently selected one.

paulj49457:
> > Does this address the issue of mass importation when RideNavigator isn't notified of new activities when more than twenty are imported

Your proposed change looks good and works fine, but fixing the display to four lines per activity, one for main info and two for calendar fields and one as a space could look better IMHO, a it looks like:

standard

So can I suggest we modify the space to 1/2 a line, it's just three value changes, which gives:

3 5 height

Just looks a bit more compact, if you like this let me know and I'll put the three value changes in a PR.

paulj49457:
A PR is overkill, here are the numbers:

1080 -         s.setHeight((rideNavigator->fontHeight+2) * **3);**

1217 --             QRect high(myOption.rect.x()+myOption.rect.width() - (7*dpiXFactor), myOption.rect.y(), (7*dpiXFactor), (rideNavigator->fontHeight+2) * **3);**

1222 -             myOption.rect.setHeight(rideNavigator->fontHeight * **2.5);** //was 36

liversedge:
I do not like compact setups, also the aesthetic is very much according to your metadata setup. I would leave it as it is.

amtriathlon:
I will merge it then, thanks for the feedback.

PR #3901, opened 2021-06-16
Bug Fixes for live map base url and remove erroneous debug message

human705:
Changed the default  base url to use the thunderforest tile server rather than the OSM one. Let me know if this string should be included in the athlete-preferences.ini file so it will not require code changes if the url changes yet again. 

Added the osmURL string to be valid and not throw a debug error message.


amtriathlon:
Hi Peter, https://tile.thunderforest.com/cycle/{z}/{x}/{y}.png?apikey=<YOUR_API_KEY> doesn't work so it is not a good default, without the apikey parameter at least it shows the map with the apikey required message.

PR #3902, opened 2021-06-16
ChartBar - Add menu to select charts without scrolling

amtriathlon:
Enabled when scrolling is enabled, allows faster selection of
invisible charts without scrolling.
This is an experimental feature, let's see how it works.

ChartBarSelectionList

Ideally it should be complemented with ChartBar autoscrolling, which would also be useful for other cases s.t. Add/Download/Library when you need to scroll to access the chart menu of the active chart.

### DIFF
--- a/src/Charts/ChartBar.cpp
+++ b/src/Charts/ChartBar.cpp
@@ -33,6 +33,7 @@ ChartBar::ChartBar(Context *context) : QWidget(context->mainWindow), context(con
     // left / right scroller icon
     static QIcon leftIcon = iconFromPNG(":images/mac/left.png");
     static QIcon rightIcon = iconFromPNG(":images/mac/right.png");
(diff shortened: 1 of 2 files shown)


liversedge:
I hate it. We need a way of collecting charts into groups, I can look at this.

liversedge:
I think we could easily add a drop down next to the search bar which allows you to add and select a "view", which would effectively select a group of charts and reduce the amount of scrolling (could also have default charts for run, swim, bike grouped this way)

Ultimately I'd like to use the overview tile layout to collect charts together, but this will need more work and has a performance impact as all charts get updated (see WKO5 for this, its horrible).

amtriathlon:
Hi Mark, I like that, it is a more comprehensive solution, I will close this.

I think we could add the view identifier (Trends/Analysis) to chart group layout files to allow the same chart group name be used in both views.

PR #3912, opened 2021-07-16
first steps to port to 5.15

andreasbuhr:
Hi, I am working on a port to Qt 5.15.
This is still work in progress.

### DIFF
--- a/src/Cloud/Dropbox.cpp
+++ b/src/Cloud/Dropbox.cpp
@@ -103,7 +103,7 @@ bool Dropbox::createFolder(QString path)
     request.setRawHeader("Content-Type", "application/json");
 
     QByteArray data;
-    data.append(QString("{ \"path\": \"%1\", \"autorename\": false }").arg(path));
+    data.append(QString("{ \"path\": \"%1\", \"autorename\": false }").arg(path).toLatin1());
     QNetworkReply *reply = nam->post(request, data);
 
     // blocking request
@@ -146,11 +146,11 @@ Dropbox::readdir(QString path, QStringList &errors)
         QByteArray data;
 
         if (firstRequest) {
-            data.append(QString("{ \"path\": \"%1\", \"recursive\": false ,\"include_deleted\": false }").arg(path));
+            data.append(QString("{ \"path\": \"%1\", \"recursive\": false ,\"include_deleted\": false }").arg(path).toLatin1());
             firstRequest = false;
         } else {
             request.setUrl(QUrl("https://api.dropboxapi.com/2/files/list_folder/continue"));
-            data.append(QString("{ \"cursor\": \"%1\" }").arg(cursor));
+            data.append(QString("{ \"cursor\": \"%1\" }").arg(cursor).toLatin1());
         }
         QNetworkReply *reply = nam->post(request, data);
 
--- a/src/Cloud/TodaysPlan.cpp
+++ b/src/Cloud/TodaysPlan.cpp
@@ -235,7 +235,7 @@ TodaysPlan::readdir(QString path, QStringList &errors, QDateTime from, QDateTime
             // Prepare the Search Payload for First Call to Search
             QString userId = getSetting(GC_TODAYSPLAN_ATHLETE_ID, "").toString();
             // application/json
-            QByteArray jsonString;
+            QString jsonString;
             jsonString += "{\"criteria\": {";
             if (userId.length()>0)
                 jsonString += "\"user\": "+ QString("%1").arg(userId) +", ";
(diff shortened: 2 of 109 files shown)


liversedge:
I don't understand as we already build just fine on QT 5.15 ?

liversedge:
version

andreasbuhr:
Hi,

yes, this is preparation for a Qt 6.2 port. Most things which are removed in Qt 6 are deprecated in Qt 5.15. So the low-pain porting procedure for Qt 6.2 is:

1. Port to 5.15
2. Fix all deprecation warnings.
3. Port to 6.2

This patch (which is still work in progress) is step 2 in the list above.

liversedge:
Ah, that makes more sense. Although it will need a LOT of testing (!)
Thanks
Mark

andreasbuhr:
I see.

A lot of changes are trivial, but some are not. Especially, it is now not possible any more to cast from QString to QByteArray. It is now enforced to specify the encoding, so one has to use .toUtf8() or .toLatin1(). In some cases I am unsure what the correct encoding is. Maybe you can help me with that? I'll add some comments in the code in this pull request later.

src/Cloud/Dropbox.cpp
```diff
@@ -103,7 +103,7 @@ bool Dropbox::createFolder(QString path)
     request.setRawHeader("Content-Type", "application/json");
 
     QByteArray data;
-    data.append(QString("{ \"path\": \"%1\", \"autorename\": false }").arg(path));
+    data.append(QString("{ \"path\": \"%1\", \"autorename\": false }").arg(path).toLatin1());
```
andreasbuhr, line 106: Here I am unsure what the correct encoding is. In what encoding does the Dropbox API expect its arguments?

src/Cloud/TodaysPlan.cpp
```diff
@@ -248,13 +248,15 @@ TodaysPlan::readdir(QString path, QStringList &errors, QDateTime from, QDateTime
             jsonString += "\"opts\": 1 "; // without fields description
             jsonString += "}";
 
-            printd("request: %s\n", jsonString.toStdString().c_str());
+            QByteArray jsonStringInUTF8 = jsonString.toUtf8();
 
-            QByteArray jsonStringDataSize = QByteArray::number(jsonString.size());
+            printd("request: %s\n", jsonStringInUTF8.toStdString().c_str());
+
+            QByteArray jsonStringDataSize = QByteArray::number(jsonStringInUTF8.size());
```
andreasbuhr, line 255: Here I just guessed that UTF-8 is the right encoding.

src/Core/Measures.cpp
```diff
@@ -42,7 +42,7 @@ Measure::getFingerprint() const
     for (int i = 0; i<MAX_MEASURES; i++) x += 1000.0 * values[i];
 
     QByteArray ba = QByteArray::number(x);
-    ba.append(comment);
+    ba.append(comment.toUtf8());
```
andreasbuhr, line 45: How are these checksums used? Would it be a problem if they changed with Qt 6.2? I.e. are they used e.g. in files, when storing results?

src/Train/TodaysPlanWorkoutDownload.cpp
```diff
@@ -379,11 +379,13 @@ TodaysPlanWorkoutDownload::getFileList(QString &error, QDateTime from, QDateTime
             jsonString += "\"opts\": 1 ";
             jsonString += "}";
 
-            QByteArray jsonStringDataSize = QByteArray::number(jsonString.size());
+            QByteArray jsonStringAsUTF8 = jsonString.toUtf8();
```
andreasbuhr, line 382: Here again I just guessed that UTF-8 is the correct encoding. Is it correct?

andreasbuhr (commented):
Here I highlighted some code paths where I am unsure what the correct encoding is when converting from QString to QByteArray. Maybe you can help?

liversedge:
this will take a lot of time to validate- my priority is getting 3.6 ready for release, so will look at this after that.
Thanks
Mark

src/Core/Measures.cpp
amtriathlon: They are used to detect changes and trigger recomputation of dependent metrics, if they change due to implementation details, they will force an otherwise not required recomputation on the first run.

andreasbuhr:
I fear this pull request will get too big and too hard to review.
I'll close it and create new pull requests with only one topic in each. Like for example only QString::null -> QString() replacement.
Of course these can stay open while you are preparing for the 3.6 release.

PR #3916, opened 2021-07-19
Replace QString::null by QString()

andreasbuhr:
QString::null was deprecated. Using the default constructor
is the way to go now.

### DIFF
--- a/src/Charts/AllPlotSlopeCurve.h
+++ b/src/Charts/AllPlotSlopeCurve.h
@@ -40,7 +40,7 @@ class AllPlotSlopeCurve: public QwtPlotCurve
 
 public:
 
-    explicit AllPlotSlopeCurve( const QString &title = QString::null );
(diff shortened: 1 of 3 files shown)


andreasbuhr:
I omitted the usages of QString::null in the "qwt" and "contrib" subdirectories, as I assume those will be removed when updating the dependencies.

amtriathlon:
> I omitted the usages of QString::null in the "qwt" and "contrib" subdirectories, as I assume those will be removed when updating the dependencies.

For qwt we had to update our version to build with Qt 5.15 (https://github.com/GoldenCheetah/GoldenCheetah/commit/e1f9005cf5ebafbed6be4b0cae4bee59112ea780) and likely we would need to do the same for Qt 6, I am not sure about other dependencies in contrib.

amtriathlon:
Thank you!

PR #3917, opened 2021-07-19
Replace QString::(Skip|Keep)EmptyParts by Qt::(Skip|Keep)EmptyParts

andreasbuhr:
QString::SkipEmptyParts and QString::KeepEmptyParts are deprecated,
they have been moved to the Qt:: namespace. This patch adapts
the code accordingly.


amtriathlon:
Thank you!

PR #3918, opened 2021-07-19
replace qrand() by QRandomGenerator::global()->generate()

andreasbuhr:
qrand() was deprecated. Instead, generate() is called on
the global QRandomGenerator singleton.


amtriathlon:
Thank you!

PR #3919, opened 2021-07-19
Use hex and endl from Qt namespace

andreasbuhr:
Using hex and endl from QTextStream's namespace is deprecated.
Using those from Qt:: namespace instead.


amtriathlon:
Thank you!

PR #3920, opened 2021-07-19
Replace QLists toSet() method by QSet constructor

andreasbuhr:
The toSet() method was deprecated. This patch changes its usages
to use the QSet constructor.


amtriathlon:
Thank you!

PR #3922, opened 2021-07-19
Replace QDate::shortDayName by QLocale::dayName

andreasbuhr:
QDate::shortDayName was deprecated. This patch replaces all usages
by QLocale::dayName.

### DIFF
--- a/src/Charts/LTMPlot.h
+++ b/src/Charts/LTMPlot.h
@@ -237,7 +237,7 @@ class LTMScaleDraw: public QwtScaleDraw
                 int year=group/12;
                 int month=group%12;
                 if (!month) { year--; month=12; }
-                label = QString("%1\n%2").arg(QDate::shortMonthName(month)).arg(year);
+                label = QString("%1\n%2").arg(QLocale::system().monthName(month, QLocale::ShortFormat)).arg(year);
             }
             break;
 
--- a/src/Core/GcCalendarModel.h
+++ b/src/Core/GcCalendarModel.h
@@ -369,7 +369,8 @@ public:
             QDate today = date(proxyIndex);
             if (today.month() != month &&
                 (today.addDays(1).month() == month || today.addDays(-1).month() == month))
-                return QString("%1 %2").arg(today.day()).arg(QDate::shortMonthName(today.month()));
+                return QString("%1 %2").arg(today.day()).arg(
+                        QLocale::system().monthName(today.month(), QLocale::ShortFormat));
             else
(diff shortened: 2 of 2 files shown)


src/Charts/LTMPlot.h
```diff
@@ -237,7 +237,7 @@ class LTMScaleDraw: public QwtScaleDraw
                 int year=group/12;
                 int month=group%12;
                 if (!month) { year--; month=12; }
-                label = QString("%1\n%2").arg(QDate::shortMonthName(month)).arg(year);
+                label = QString("%1\n%2").arg(QLocale::system().monthName(month, QLocale::ShortFormat)).arg(year);
```
ediosyncratic, line 240: Your replacement here is faithfully correct to what was in 5.15 and before: QDate always did use QLocale::system() for these names.

Note that this QString().arg() construct (is slightly more efficient than, but) would be more idiomatically expressed by:
`QLocale::system().toString(QDate(year, month, 1), u"MMM\nyyyy")`

This will, incidentally, ensure a four-digit year, even if the date being considered is more than 1021 years ago.
If that isn't desired, ```u"MMM\ny"``` will do the trick.

In any case, using QLocale::system() probably isn't an ideal choice.
The following is advice for whoever maintains GoldenCheetah !
It is probably better to use QLocale() rather than QLocale::system(); the former is configurable by calling QLocale::setDefault(); it defaults to the latter if that isn't called.
Alternatively, if this code is within some form of GUI element (widget, window, etc.) you may find it has a locale() method, that it might be more appropriate to use, although that would be a change from the prior behaviour.
That would let the user configure the locale by using the normal facilities of Qt GUI elements.

src/Core/GcCalendarModel.h
```diff
@@ -369,7 +369,8 @@ public slots:
             QDate today = date(proxyIndex);
             if (today.month() != month &&
                 (today.addDays(1).month() == month || today.addDays(-1).month() == month))
-                return QString("%1 %2").arg(today.day()).arg(QDate::shortMonthName(today.month()));
+                return QString("%1 %2").arg(today.day()).arg(
+                        QLocale::system().monthName(today.month(), QLocale::ShortFormat));
```
ediosyncratic, line 373: The more idiomatic form of this would be: `QLocale::system().toString(today, u"d MMM");`
If you want a two-digit day-number (with leading zero for days < 10), you can use `dd` in place of `d`.

amtriathlon:
I will merge the original version due to being simpler, thank you!

PR #3923, opened 2021-07-19
Replace QList::swap by QList::swapItemsAt

andreasbuhr:
QList::swap was renamed to QList::swapItemsAt. This patch adapts
the source code accordingly.


amtriathlon:
Thank you!

PR #3924, opened 2021-07-19
Replace QTime by QElapsedTimer to measure timings

andreasbuhr:
Usage of QTime to measure timings is deprecated. This patch
replaces all usages by QElapsedTimer.


andreasbuhr:
This requires careful testing whether TrainSidebar still works as expected.

PR #3925, opened 2021-07-19
Initialize QFlags by default constructor

andreasbuhr:
Initializing QFlags by "0" is deprecated. This patch changes all
code places which does this to use the default constructor.


amtriathlon:
Thank you!

PR #3926, opened 2021-07-19
Replace qVariantFromValue by QVariant::fromValue

andreasbuhr:
The function qVariantFromValue is deprecated and was replaced
by QVariant::fromValue. This patch adapts the code accordingly.


amtriathlon:
Thank you!

PR #3927, opened 2021-07-19
Change QDate to QDateTime conversions, specifying time

andreasbuhr:
The QDateTime constructor taking a QDate as argument is deprecated.
QDate::startOfDay() should be used instead when converting a QDate
to a QDateTime. This makes clear what time is used in this conversion.
This patch adapts the code to this.


amtriathlon:
Thank you!

PR #3931, opened 2021-07-20
Use default constructor for QFlags in QWT

andreasbuhr:
Conversion from int is deprecated.

Only to be merged if qwt is not updated.


amtriathlon:
I am closing this because we replaced qwt in https://github.com/GoldenCheetah/GoldenCheetah/commit/49cf6340a4d2e218a60e5a80ef97ce4cba0f6894

PR #3932, opened 2021-07-20
Specify encoding when communicating with TodaysPlan

andreasbuhr:
Conversion from QString to QByteArray is deprecated in Qt 6.
It will be required to specify the encoding. This patch
adapts the interaction with TodaysPlans API to specify the
encoding as UTF8.
This should be no behavior change, as encoding in UTF8 was the
default in the now deprecated methods, see
https://doc.qt.io/qt-5/qbytearray-obsolete.html#operator-2b-eq-3
.


amtriathlon:
Thank you!

PR #3933, opened 2021-07-20
Specify encoding when communicating with DropBox

andreasbuhr:
QByteArray::append(QString) was deprecated. It is now required
to specify an encoding. This patch adds the corresponding calls
to the communication with DropBox.
This should not be a behavior change, as UTF8 was the default
when converting from QString to QByteArray.


amtriathlon:
Thank you!

PR #3936, opened 2021-07-20
Replace qStableSort by std::stable_sort

andreasbuhr:
qStableSort is deprecated. This patch replaces all usages by
std::stable_sort.


amtriathlon:
Thank you!

PR #3937, opened 2021-07-20
Add assignment operators to some classes

andreasbuhr:
The implicitly defined assignment operator for classes having custom
defined copy constructor is deprecated. This patch adds explicit
assignment operators.

This is not required for porting to Qt6.

### DIFF
--- a/src/FileIO/RideFile.h
+++ b/src/FileIO/RideFile.h
@@ -562,13 +562,17 @@ public:
             string[i]="";
         }
     }
-    XDataPoint (const XDataPoint &other) {
+    XDataPoint(const XDataPoint &other) {
+        *this = other;
+    }
+    XDataPoint& operator=(const XDataPoint &other) {
         this->secs=other.secs;
         this->km=other.km;
         for(int i=0; i<XDATA_MAXVALUES; i++) {
             this->number[i]= other.number[i];
             this->string[i]= other.string[i];
         }
+        return *this;
     }
 
     double secs, km;
@@ -579,7 +583,8 @@ public:
 class XDataSeries {
 public:
     XDataSeries() {}
-    XDataSeries(XDataSeries &other) {
+    XDataSeries(const XDataSeries& other) { *this = other; }
+    XDataSeries& operator=(const XDataSeries &other) {
         name = other.name;
         valuename = other.valuename;
         unitname = other.unitname;
@@ -589,6 +594,7 @@ public:
         foreach (XDataPoint *p, other.datapoints) {
             datapoints.push_back(new XDataPoint(*p));
         }
+        return *this;
     }
 
     ~XDataSeries() { foreach(XDataPoint *p, datapoints) delete p; }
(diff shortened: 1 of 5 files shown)


amtriathlon:
I reverted the merge due to crashes on metric refresh after changing CP.

src/FileIO/RideFile.h
```diff
@@ -589,6 +594,7 @@ class XDataSeries {
         foreach (XDataPoint *p, other.datapoints) {
             datapoints.push_back(new XDataPoint(*p));
         }
```
andreasbuhr, line 596: This foreach loop is wrong for an assignment operator. It only adds new points, but the old points have to be deleted first (if present).
We'd need something like the following before:
```
foreach(XDataPoint *p, datapoints) delete p;
datapoints.clear();
```

andreasbuhr (commented):
I reviewed the code again and found one problem. Could the problem I found in the XDataSeries assignment operator be the reason for the crashes you saw?

src/FileIO/RideFile.h
liversedge: This is a constructor so old data series should not exist?

andreasbuhr: Yes, in the constructor, it is correct. But this MR moved the code to the assignment operator and then implemented the copy-constructor in terms of the assignment operator. And in the assignment operator, this previous delete/clear is necessary.

liversedge: Ah ok, yes I misunderstood your comment- that should fix the error in the commit

amtriathlon:
> I reviewed the code again and found one problem. Could the problem I found in the XDataSeries assignment operator be the reason for the crashes you saw?

Yes, and that seems to make sense but, after applying the proposed fix (https://github.com/amtriathlon/GoldenCheetah/commit/6610dd6f16a9ac6b421a3994c5a053a2e9f29f67) now I am having crashes much more random and hard to debug, even at startup, so I don't understand what's happening.

andreasbuhr:
Sorry, I am out of ideas as well.

amtriathlon:
Thank you, my suspicion is the problem may not necessarily be caused by this change, it may be triggering a hidden issue which currently doesn’t provoke a crash but perhaps a memory leak.

thejockl:
One source of such a crash is in JsonRideFile.y (after having the additional delete-patch from @andreasbuhr applied to RideFile.h):
When creating a new XDataSeries, the old object is copied from jc->xdataseries to the object add, including the datapoints and its contained pointers without creating new instances of these XDataPoints. Afterwards  a new instance of XDataSeries is assigned to jc->xdataseries, causing a deletion of all datapoints. Now add contains an array of already deleted datapoints.

Changing the code to
```
xdata_series: '{' xdata_items '}'              { XDataSeries *add = new XDataSeries(jc->xdataseries);
                                                 jc->JsonRide->addXData(add->name, add);

                                                 // clear for next one
                                                 jc->xdataseries = XDataSeries();
                                               }
```
fixed it for me.

I will continue testing to see if I can find more such cases.

amtriathlon:
Great, that fixed the crashes I was experiencing!

Looking at other uses of XDataSeries I found a memory leak in: https://github.com/GoldenCheetah/GoldenCheetah/blob/81fc763a4f8428e70ade980a22067806274f1685/src/Gui/SplitActivityWizard.cpp#L801-L802

I have the intention to apply the fixed PR plus these changes tomorrow, provided no other cases are detected.

thejockl:
No more crashes on my machine, also valgrind did not give any further related alerts. For valgrinds other findings (mostly uninitialized but used variables), I will create a separate PR

amtriathlon:
Fixed version was merged by https://github.com/GoldenCheetah/GoldenCheetah/commit/35433b4f6594f2334520d64c2c08ba56c65b1937, thank you!

PR #3940, opened 2021-07-21
Use QPainter::setTransform instead of QPainter::setMatrix

andreasbuhr:
QPainter::setMatrix is deprecated. This patch replaces
usage of setMatrix by usage of setTransform.


amtriathlon:
Thank you!

PR #3941, opened 2021-07-21
Replace QColor::dark by QColor::darker and light by lighter

andreasbuhr:
QColor::dark and QColor::light are deprecated. This patch adapts
the code to use QColor::darker and QColor::lighter.


amtriathlon:
Thank you!

PR #3942, opened 2021-07-21
Usage of QDesktopWidget is deprecated. Adapt code

andreasbuhr:
This patch adapts the code to not use QDesktopWidget but to
use QGuiApplication::screens instead. It defaults to screen number 0.


liversedge:
this assumes we are on screen 0?
what if we are on another screen?

andreasbuhr:
Hmm, maybe we should use QGuiApplication::primaryScreen() to replicate the old behavior. Needs more investigation.

andreasbuhr:
I changed to code from screens()[0] to primaryScreen().

PR #3951, opened 2021-07-22
Replace QPalette::Background by QPalette::Window

andreasbuhr:
Both QPalette::Background and QPalette::Window have the value 10,
there is no difference.
See
https://doc.qt.io/qt-5/qpalette.html#ColorRole-enum


amtriathlon:
Thank you!

PR #3953, opened 2021-07-22
Qt6: specifying utf8 for QTextStream is the default

andreasbuhr:
This must not be merged before porting to Qt6.
QTextStream defaults to codecForLocale in Qt5, see
https://doc.qt.io/qt-5/qtextstream.html
and to UTF8 in Qt6, see
https://doc.qt.io/qt-6/qtextstream.html


amtriathlon:
TODO: adapt this to preserve current code inside
```
#if QT_VERSION < 0x060000
#endif
```
besides to resolve the conflicts.

PR #3962, opened 2021-07-26
Qt6: adapt checksum methods

andreasbuhr:
This is incompatible with Qt 5.15 and must not be merged before really going to Qt 6.

Probably it will change checksums of UserMetricSettings. If this is not acceptable, this needs more work.

### DIFF
--- a/src/Core/Measures.cpp
+++ b/src/Core/Measures.cpp
@@ -44,7 +44,7 @@ Measure::getFingerprint() const
     QByteArray ba = QByteArray::number(x);
     ba.append(comment);
 
-    return qChecksum(ba);
+    return qChecksum(ba, ba.length());
 }
 
 QString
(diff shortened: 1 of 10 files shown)


liversedge:
The code should be adapted to check the QT version, e.g.
```
#if QT_VERSION >= 0x060000
    // do the Qt6 code
#else 
   // do the current code
#endif 
```

andreasbuhr:
Ah, do you plan to be compatible with both Qt5 and Qt6 for a while?

liversedge:
Absolutely. It helps the transition not just for the team, but for building and distribution.
It is unlikely that Qt6 will be stable for a few months, so we are not likely to just jump to it.

andreasbuhr:
I changed this back to use the qChecksum(QByteArray, qsizetype size) methods. While they are deprecated, they still exist in Qt 6.2.

amtriathlon:
Thank you!

PR #3967, opened 2021-07-26
Qt6: Remove call to qRegisterMetaTypeStreamOperators

andreasbuhr:
Stream operator registration is now automatic, see
https://doc.qt.io/qt-6/qtcore-changes-qt6.html

Not to be merged before going to Qt6


liversedge:
Similar to qChecksum commit you should use the pre-processor to check the qt version rather than make the code not compile on earlier releases. Thanks.

amtriathlon:
Thank you!

PR #3968, opened 2021-07-26
QWebEnginePage::view was removed

andreasbuhr:
The view() method of the page does not exist any more in Qt 6.2.
I think it is redundant with the preceding call anyway, isn't it?


amtriathlon:
It looks like it is redundant, otherwise we could use the static method https://doc.qt.io/qt-6/qwebengineview.html#forPage to find the corresponding view in Qt6 using conditional compilation.

andreasbuhr:
If I remember correctly, I discussed this patch with someone in the WebEngine-Team of Qt. He told me it's redundant.

amtriathlon:
Thank you!

liversedge:
Classic case of re-using the same code- it is indeed redundant

PR #3969, opened 2021-07-26
Qt6: Remove usage of QHashIterator

andreasbuhr:
QHashIterator is no longer able to iterate over QMultiHash, even
though it is documented to be.

This patch removes usages of QHashIterator on QMultiHash.

An alternative solution would be to use QMultiHashIterator, but that is only available in Qt 6 and thus would require an ifdef on the Qt version. This solution works with both Qt6 and Qt5.


andreasbuhr:
I simplified the solution considerably. 
Added the ifdef to be Qt5 compatible.

amtriathlon:
Thank you!

PR #3973, opened 2021-07-26
Fix number to QString conversion in FixElevation

andreasbuhr:
Some clarification is required here

I am a bit lost with the old code. It does not compile with Qt 6.2 and I don't see what is intended here.


amtriathlon:
This looks like a bug, I will fix it in master. Thanks for the heads up.

PR #3974, opened 2021-07-26
QMap::const_iterator cannot do +1 any more. Workaround this.

andreasbuhr:


andreasbuhr:
As https://codereview.qt-project.org/c/qt/qtbase/+/362150 got merged, this is probably no longer required.

amtriathlon:
Looking at the codereview I think it is better to merge this, thank you!

PR #3975, opened 2021-07-26
Use factory method for QLowEnergyController

andreasbuhr:
Directly calling QLowEnergyController's constructor was deprecated


amtriathlon:
Thank you!

PR #3976, opened 2021-07-26
Remove usage of QGLWidget

andreasbuhr:
QGLWidget was deprecated and is removed in Qt 6.


amtriathlon:
TODO: adapt this to preserve current code behavior inside
```
#if QT_VERSION < 0x060000
#endif
```

liversedge:
> TODO: adapt this to preserve current code inside
> 
> ```
> #if QT_VERSION < 0x060000
> #endif
> ```
> 
> besides to resolve the conflicts.

Let me test on Linux and Qt5 to see if it makes any real difference ?

andreasbuhr:
Sure

amtriathlon:
I resolved the conflict, the installer build failed due to the removal of Qt5OpenGL.dll, if we choose this route I will change the installer script, but perhaps it is better to keep without changes for Qt5 and remove QGLWidget for Qt6 only.

liversedge:
It certainly makes a difference on Linux/Qt5- performance gets sluggish quickly on the user charts and overview

andreasbuhr:
> It certainly makes a difference on Linux/Qt5- performance gets sluggish quickly on the user charts and overview

Can you clarify please, is this a big or a small problem?

liversedge:
Ale has covered this in the commit, so keep it for builds < Qt6
Its not a huge issue, it will be interesting to see how Qt charts perform on Qt6 given they are still dependant on OpenGL

PR #3977, opened 2021-07-26
There are two overloads for QStringList and QVector<QString>

andreasbuhr:
These are not required in Qt6, as QStringList is an QVector<QString>


amtriathlon:
Thank you!

PR #3979, opened 2021-07-26
Some more workarounds about iterators which cannot +1

andreasbuhr:


andreasbuhr:
As https://codereview.qt-project.org/c/qt/qtbase/+/362150 got merged, this is probably no longer required.

amtriathlon:
Thank you!

PR #3980, opened 2021-07-26
workaround circular dependency for qt6

andreasbuhr:
In Qt6, moc needs to know the types of properties. So one would need to include RideItem.h from GoldenCheetah.h. But we run into a circular dependency there, becasue RideItem.h includes GoldenCheetah.h.
To solve this, add this moc-include.


andreasbuhr:
I added the ifdef-s for Qt5 compatibility here.

amtriathlon:
Thank you!

PR #3986, opened 2021-07-29
Replace deprecated QSignalMapper::mapped signal

andreasbuhr:
The QSignalMapper::mapped signal does not exist any more and
was replaced by mappedString, mappedInt, and mappedObject.
This patch adapts its usages and only transitions to
new style signal-slot connection syntax.

This is required for porting to Qt 6.
The new signals were introduced in Qt 5.15. If you need only compatibility with Qt 5.15 and Qt6, this is fine as-is. If you want to continue supporting older Qt versions, this needs to be modified and ifdef-ed on Qt version.

### DIFF
--- a/src/Gui/ColorButton.cpp
+++ b/src/Gui/ColorButton.cpp
@@ -123,7 +123,7 @@ GColorDialog::GColorDialog(QColor selected, QWidget *parent) : QDialog(parent), 
 
     // map button signals
     mapper = new QSignalMapper(this);
-    connect(mapper, SIGNAL(mappedInt(int)), this, SLOT(gcClicked(int)));
+    connect(mapper, &QSignalMapper::mappedInt, this, &GColorDialog::gcClicked);
 
     // now add all the colours to select
     colorSet = GCColor::colorSet();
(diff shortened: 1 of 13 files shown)


src/Gui/ColorButton.cpp
```diff
@@ -123,7 +123,11 @@ GColorDialog::GColorDialog(QColor selected, QWidget *parent) : QDialog(parent),
 
     // map button signals
     mapper = new QSignalMapper(this);
+#if QT_VERSION >= 0x060000
     connect(mapper, SIGNAL(mappedInt(int)), this, SLOT(gcClicked(int)));
```
andreasbuhr, line 127: I am a bit confused how this worked. According to the documentation
https://doc.qt.io/qt-5/qsignalmapper.html#mappedInt
"mappedInt" was introduced in Qt 5.15. So this should not work with 5.14. Is this a pre-existing bug?

andreasbuhr:
I added compile-time switches so the code stays compatible with Qt 5.14.

amtriathlon:
I think Qt5.15 compatibility is enough now and the resulting code is simpler, I plan to resolve the conflicts using only the first commit.

PR #4030, opened 2021-08-30
User selectable map route line color and switchable route line out & back opacity

paulj49457:
To try and address comments 1&2 raised in : https://groups.google.com/g/golden-cheetah-users/c/LQ1bq3zGTeE/m/5AvoGKX6AQAJ

_1) Map tiles layer should be able to be dimmed or desaturated to make the ride's path more visible - The map details are great, but often visually overpower and make it hard to see the ride's path (map details have many colours and saturated colours).. especially if multicoloured "shaded zone" (which I find useful is on)
2) The colour and width of yellow outline of ride path should be configurable - depending on map tiles other colours might be preferable (don't overlap colours of map tile details).. increasing width (if highlight colour is transparent) make path more visible_

Trying to keep things simple and hopefully provide sufficient configurability I have proposed:

1) the option to switch off the out & back route line opacity, this can improve the visibility of the route line, the default is the standard GC 50% opacity display.

Capture1

2) and I have also added the option to change the route line colour (see below), note: this is only displayed for activities with no power information or when the hide shaded zones is selected (checkbox ticked), and again defaults to the standard GC red:

Capture2

Note the power zone line colours are already configurable, the difficulty is trying to choose them so they appear clear on some tile server displays :

Capture3

### DIFF
--- a/src/Charts/RideMapWindow.cpp
+++ b/src/Charts/RideMapWindow.cpp
@@ -71,6 +71,8 @@ RideMapWindow::RideMapWindow(Context *context, int mapType) : GcChartWindow(cont
     hideShadedZonesCk->setChecked(false);
     hideYellowLineCk = new QCheckBox();
     hideYellowLineCk->setChecked(false);
+    hideRouteLineOpacityCk = new QCheckBox();
+    hideRouteLineOpacityCk->setChecked(false);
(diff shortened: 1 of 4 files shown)


paulj49457:
Any chance this can get merged in 3.6 ?

Issue #3802, opened 2021-01-27
Control trainers using Bluetooth FTMS


amtriathlon:
I went and looked at [bluetooth.com](https://www.bluetooth.com/specifications/gatt) GATT specifications and found: 

-FTMP - [Fitness Machine Profile](https://www.bluetooth.org/DocMan/handlers/DownloadDoc.ashx?doc_id=423417)
-FTMS - [Fitness Machine Service](https://www.bluetooth.org/DocMan/handlers/DownloadDoc.ashx?doc_id=423422)

Both have been adopted and are listed as active. 

_Originally posted by @Frono in https://github.com/GoldenCheetah/GoldenCheetah/issues/2771#issuecomment-451747554_

mjunkmyjunk:
I would love to see support for FTMS as well.  Having looked through the documentation/spec, I am appalled at how little I understand the protocol and the way it's interpreted. (Somehow, reading ANT+ specs is much more intuitive)

erikboto:
I started looking at this. I've implemented the "other side" of the FTMS protocol in devices so I think I have all the pieces I need in order to get this done fairly easily, specially since the ground work of having controlled devices in BT40Device. I'll link this issue when I start a PR so that you'll know when there's something to test.

ZeniXue:
Hi., I'm offering a job just to get the Wahoo KICKR 2020 sensor's data like power, cadence, speed, resistance, and heart rate into a Unity game engine. If anyone is interested, please let me know. Thank you.
syafiqbahrin@gmail.com

PR #4045, opened 2021-10-03
WIP: Add support for FTMS devices

erikboto:
First shot at implementing FTMS support (#3802).

I only have access to a somewhat limited FTMS device, so I've only been able to test ERG-mode and gradient mode.

It would be great if we could have testers using a variety of different devices to help test the different features/modes.


MeisterP:
I've tested your patch with a kickr core running firmware 1.1.1 which supports FTMS

from https://support.wahoofitness.com/hc/en-us/articles/360000217839-KICKR-CORE-Firmware-Updates
> KICKR CORE Firmware v1.1.1 - 1 June 2021
>     Added: Bluetooth FTMS support

To me it looks like the trainer is discovered twice. Once by the old, wahoo specific BTLE code and then once more by your new FTMS code.
Not sure which one is used in the end.

I've edited the log files because there was a lot of bluetooth noise from other devices. Hopefully still useful.

[gc-no-ftms.log](https://github.com/GoldenCheetah/GoldenCheetah/files/7285088/gc-no-ftms.log)
[gc-ftms.log](https://github.com/GoldenCheetah/GoldenCheetah/files/7285087/gc-ftms.log)

erikboto:
@MeisterP Thanks for testing! Yeah the way the BT40 code works is a bit weird at the moment, since it will connect to everything it finds. At some point it would make sense to at least be able to choose which type of BLE services you want to use, so that you could either choose to the kickr-interface or the FTMS interface. And also have to option to maybe not connect a power meter if you use ftms/kickr, if you want to be sure you get the power from the trainer etc.

If you have the time and energy to test more, you could comment out this line (and the following, since it's split into two): 
https://github.com/erikboto/GoldenCheetah/blob/ftms/src/Train/BT40Device.cpp#L254

Then I think the custom KICKR interface should not be used, so only FTMS would be tested.

MeisterP:
> If you have the time and energy to test more, you could comment out this line (and the following, since it's split into two): https://github.com/erikboto/GoldenCheetah/blob/ftms/src/Train/BT40Device.cpp#L254
> 
> Then I think the custom KICKR interface should not be used, so only FTMS would be tested.

This does disable the custom KICKR interface indeed. Will report back after my next indoor workout.

amtriathlon:
Hi Eric, thanks for your work, I can't test this but I started a thread on users forum to recruit volunteers. 

To call setWatts and setCadence from multiple sources concurrently is a no go, IIRC.

Until we have a proper pairing step for BTLE we should prioritize a which source uses setWatts/setCadence and use setAltWatts for the other, be in a fixed or configurable way.

erikboto:
I have tried to create a solution for devices that have more than one way of controlling them, so that each device will only be controlled using one method.

The idea is that when we have discovered the services a device has, we do through them and make sure we keep only one service that provides control. Then there's an extra check for the Wahoo Brake Control, since they just add a characteristic to the Cycling Power Service so we can't filter it out on a service level. So if we already have a controllable service at the time we have discovered details about the CPS service, we don't try to add the Wahoo Brake Control.

Since I still don't have any good way of testing this, I would very much appreciate if @MeisterP have the possibility to test if this prioritization works with the FTMS capable Kickr, and then we'll have to hope that someone with Kurt and Tacx trainers can help out with testing the "service level" prioritization.

amtriathlon:
Hi Eric, my comment above was about telemetry: if I understand your code correctly, you are setting watts and cadence from the FTMS service when present, but if there is also a Cycling Power service or a Speed and Cadence one, we will have double updates which I think are not well handled. Current code disable Speed and Cadence service when there is a Cycling Power service present to avoid this.

amtriathlon:
> Right, I think I understood that the first time I read it but then lost that thought when I was thinking about control priority.
> 
> Looking at the code it seems like the Kurt device and FTMS devices provide power/cadence via their own characteristics, while Tacx and Kickr rely on the Cycling Power Service. So I guess we could just avoid connecting Cycling Power Service if a device has Kurt or FTMS, using the same "filtering" as I do to avoid multiple controllable services. Then I would also set the has_power for Kurt and FTMS so that speed and cadence is avoided in those cases too.
> 
> Do you think that would work?

Yes, that seems reasonable to me.

FTMS service can't provide Speed?

Odald:
Hi,
I tested this on Windows 10 with the trainer "Elite Direto X".
I get "Failed to write BTLE characteristic for device "DIRETO X".
I attached a log:
[goldencheetah_log.txt](https://github.com/GoldenCheetah/GoldenCheetah/files/10474090/goldencheetah_log.txt)

PR #4046, opened 2021-10-04
Fix python-version in travis after_success.sh for linux build

erikboto:
The upstream AppImage for python3.7.x got upgraded from 3.7.11 to
3.7.12, causing the script to fail.

I'm still waiting for a build with this fix included to finish in my own fork with travis-ci enabled: https://app.travis-ci.com/github/erikboto/GoldenCheetah/jobs/541072491 , if that works as intended I guess it can be considered tested ok.


erikboto:
Well it seems like it passed the previous point of fail, the job was eventually killed due to "The job exceeded the maximum log length, and has been terminated.".

PR #4073, opened 2021-11-17
WIP: FTMS: Add support for Spin Down Calibration

erikboto:
A first attempt at implementing the spin down calibration support for FTMS devices that support it.

I don't have access to any FTMS trainer to test with, so any help testing this out is very much appreciated.


amtriathlon:
Hi Eric, it looks like a safe change since the new code will not be activated until the user starts calibration on a capable device, I think we could merge this to allow user testing in other platforms than windows.

PR #4092, opened 2021-12-02
Support to set track background color and fg/bg opacity + initial cmake build support

frli8848:
* This PR adds support to set the background track color  which previously was hard coded to yellow
* It also adds support to set the opacity of the background track line and the foreground line/segement opacity
*  There is also initial (experimantal) support to build GC using cmake. The cmake files do not touch any other code.

Here is a screen shot using the standard yellow background trackline
track_bg_line_yellow
:

Here is a screen shot using the a custom color of the background trackline
track_bg_line_custom

And here is a screenshot of the Chart Settings dialog with the color chooser open:
chart_settings_color


paulj49457:
Please take a look at the PR: User selectable map route line colour and switchable route line out & back opacity #4030 which uses the existing colour selection pallet within GC, and has been waiting to be merged since August 21.

PR #4097, opened 2021-12-05
Python automated processors fix, working as in manual mode

### LINKED ISSUE #4095: Enable Python Data Processors to be run in automatic mode
Metadata handling functions, introduced by https://github.com/GoldenCheetah/GoldenCheetah/commit/9315b301d739feb2302de47a54a5699ecccb4818 and https://github.com/GoldenCheetah/GoldenCheetah/commit/d7e1747a2b6e3493f9f50365768f7212cee92584, don't work when used in a Data Processor configured to run automatically on import or save.

Edit: the problem is more general than originally reported since read only access to metadata neither works in automatic mode.

peret2000:
Python automated processors cannot execute setTag/delTag/hasTag correctly, as well as meny other features, as the activity being processed is not the RideItem object in the context variable; in fact, it is set with the currently selected activity in the activities list.
This commit changes that, and allows python processors behave as manual processors.
This commit should be verified by some expert, as there can be side effects unknown by me. The most straightfoorward one is: a RideItem is created but not destroyed. The reason is that it cannot be deleted calling its destructor, as it would close and delete the file it has as a property, that was created beforehand. Maybe a trick would be to set to null its ride_ property before deleting it.
But thee correct solution, I think, would be to avoid the need of creating a RideItem object (reuse one if it is already created, or created but inserted in some list wherever it is taken into account,...)
Perhaps this code is not the correcct one, but can give a clue to deveelop a better solution.

Anyway, its behavior is correct, processors run and can read and modify metada, for example.

Resolves #4095

### DIFF
--- a/src/FileIO/FixPyDataProcessor.cpp
+++ b/src/FileIO/FixPyDataProcessor.cpp
@@ -37,6 +37,9 @@ bool FixPyDataProcessor::postProcess(RideFile *rideFile, DataProcessorConfig *se
     FixPyRunner pyRunner(rideFile->context, rideFile, useNewThread);
     bool result = pyRunner.run(pyScript->source, pyScript->iniKey, errText) == 0;
     rideFile->context->ride = prior;
+    ride->context = NULL;
+    ride->setRide((RideFile*)NULL);
+    delete ride;
     return result;
 }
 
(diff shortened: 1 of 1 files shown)


amtriathlon:
See https://github.com/GoldenCheetah/GoldenCheetah/issues/4095#issuecomment-1024244478

PR #4108, opened 2021-12-11
Corrected problems with LRBalance Import

grmpl:
I've made two corrections to the import of FIT files to correct problems with LRBalance-Data.

First, the formula for interpolation of data was wrong. It would count up from -255 when prevPoint->lrbalance is an RideFile::NA-value. 
I did change the algorithm not to interpolate values when prevPoint->lrbalance or lrbalance is RideFile::NA, but just to set all the intermediate values to RideFile::NA, too. IMHO it doesn't make sense to interpolate this value as LRBalance usually is not monotonic and changes within a few seconds.

Second I reverted the change which excluded all LRBalance-data without right-bit set.
I checked all the issues and I found no technical reason which made this change necessary. There were only two reasons for this change:
1) A concern by Alejandro in issue 3819 "Obviously it is easy to revert that change, but my concern is we loose the ability to detect LR balance marked as invalid."
2) The ANT-specification

The concern was never confirmed by any issue, nor any complaint from an user. The specification is a proprietary specification controlled by one manufacturer. Other manufacturers are using left-based values with right bit not set, it would not be ok to exclude these manufacturers.

### DIFF
--- a/src/FileIO/FitRideFile.cpp
+++ b/src/FileIO/FitRideFile.cpp
@@ -1895,10 +1895,10 @@ struct FitFileReaderState
                              break;
                     case 30: //LEFT_RIGHT_BALANCE
                              // When bit 7 is 1 value are right power contribution
-                             // not '1' the location of the contribution is undefined
-                             if (value & 0x80)
-                                lrbalance = 100 - (value & 0x7F);
-                             else
+                             //  some manufacturers use left power contribution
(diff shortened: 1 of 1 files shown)


amtriathlon:
The relevant part to fix #4117 was applied in https://github.com/GoldenCheetah/GoldenCheetah/commit/0108a05f5ae97e4b595f10e2fe85bc1a3ac836a9, thank you.

PR #4147, opened 2022-01-24
Update Spanish translation

jgpallero:
New update to the Spanish translation. This is the third one after https://github.com/GoldenCheetah/GoldenCheetah/commit/f906e8751c696c098513af57141a64e19cedfd4a and https://github.com/GoldenCheetah/GoldenCheetah/commit/0449988fb1f71b690b75ed4cf7e03993f92a9427

### DIFF
--- a/src/Resources/translations/gc_es.ts
+++ b/src/Resources/translations/gc_es.ts
@@ -6917,9 +6917,7 @@ Esto puede conducir a inconsistencias entre sus activides locales en GoldenCheet
     <message>
         <location filename="../../Gui/Pages.cpp" line="1276"/>
         <source>The quick brown fox jumped over the lazy dog</source>
-        <translation>El veloz murciélago hindú comía feliz cardillo 
-y kiwi. La cigüeña tocaba el saxofón detrás 
-del palenque de paja</translation>
+        <translation>Canta, oh Diosa, la cólera del pelida Aquileo</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


amtriathlon:
I don’t agree to translate weight as masa.
To change weight to mass is a decision to be taken at source code level, not translation level.

jgpallero:
The reason because I've translated "Bike Weight" as "Masa de la bicicleta" in FixDerivePowerConfig is that in SimBicyclePage all the original strings says "Mass", so in this case I've preferred maintain the word "Masa" for all data related to the bike.

I've used the word "Peso" for body measures, so I use always "Peso del atleta" for "Athlete Weight". The only exception in body measurements is the use of "Masa" for "Masa magra del atleta", and related measurements. The reason is that traditionally in Spanish this is the term used (I don't know the reason). For the person weight it is used "peso", but for the components the term "masa" is employed. And this is used in this form in Spain (https://www.cun.es/diccionario-medico/terminos/peso-corporal, https://www.cun.es/diccionario-medico/terminos/masa-corporal-magra) and also in America (https://revistas.unlp.edu.ar/raab/article/view/3148).

Strictly speaking, we must use always mass, as this is the data we read in a scale (although internally it measures force, F=m*a, the data we read is an approximation to mass), but the word weight is consolidated for the body mass. There are the reasons because I've used the term "masa" in some strings

PR #4149, opened 2022-01-25
Bug fix in wheel's moment of inertia computation

jgpallero:
I've put in contact with the author (Eric Christoffersen), and he confirmed it is a bug


amtriathlon:
Intuitively the change looks wrong.
Do you have a reference for the new formula?

jgpallero:
Yes your are right, my change is wrong and the original formula is correct. I made a mistake when I deduced the formula for a disc with a hole. Cancel the pull request, please

PR #4154, opened 2022-02-05
src/Charts/PythonChart.cpp: allow LocalContentCanAccessRemoteUrls

MeisterP:
new versions of qtwebengine block those requests by default

fix "Access to fetch at *** from origin 'file://' has been blocked by CORS policy:
   Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, chrome-untrusted, https."

see https://forum.qt.io/topic/132956/qwebengineview-javascript-fetch-cors-error-but-works-on-chrome-desktop


MeisterP:
it's the same setting as in https://github.com/GoldenCheetah/GoldenCheetah/blob/a13da9e91751566fec544673b4b2509b12821a3b/src/Train/WebPageWindow.cpp#L158

Maybe there are other places where that setting would be useful too... so far, I just had issues with some of my python charts. Hence this pull request.

PR #4172, opened 2022-02-15
Release 3.5.0

mrflyf:
I am a huge fan of the "Summery" in the "Trends" fan and I would really like to compare 2 og more periods for instance Q1 in 2020, 2021 and 2022.
I use the Date range to set up each period per year but I need a feature wher it is possible to choose more than one date range like you can do with the workout code.
That would be great to have.


amtriathlon:
Active development happens on master branch, there will be no new v3.5 release and no PRs will be accepted on that branch.

PR #4200, opened 2022-03-22
corrected Japanese translation 

Overcomb3:
translation corrections in the season tub of Trends page


amtriathlon:
Hi, welcome to the project, a couple of comments:
1) You shouldn't change the source texts, they are generated by Qt lupdate tool from source code.
2) when translations are finished they should be marked as that, not let as unfinished.
To avoid this and other errors, the recommended way to work on translations files is to use Qt linguist tool, as explained in https://github.com/GoldenCheetah/GoldenCheetah/wiki/Guidelines-for-translators

Overcomb3:
I'm sorry for my mistakes.
I'm going to study and try correcting again.

PR #4207, opened 2022-04-16
Addition of CoreTemp sensor support

ihewitt:
This patch adds ANT+ support for CoreTemp body sensors. 

As discussed on #4121 this is now updated to use the XDATA fields rather than changing the saved json.

If there is core temperature data then it uses that for "tcore" data instead of using the estimated calculation.


src/FileIO/JsonRideFile.y
```diff
@@ -105,6 +105,8 @@ static QString protect(const QString string)
 #define scanner jc->scanner
 
 %}
+// generated by the scanner
+%define api.header.include {"JsonRideFile_yacc.h"}
```
amtriathlon, line 109: This change breaks ci builds since it requires a newer bison version

ihewitt: Apologies. No idea what made that change, removed.

ihewitt:
Spotted a bug while fixing the CI issue, tidied the change. :smile:

liversedge:
If you want to edit the data remember to update RideEditor::insColumn() and RideEditor::whatColumns()

amtriathlon:
Hi Ivor, this is marked as WIP and now has a conflict with Cycling Dynamics support, do you still want to proceed with this?

src/Core/DataFilter.y
```diff
@@ -48,6 +48,7 @@ extern int DataFilterparse();
 extern Leaf *DataFilterroot; // root node for parsed statement
 
 %}
+%define api.header.include {"DataFilter_yacc.h"}
```
liversedge, line 51: Why is this included in this patch ?

src/Core/RideDB.y
```diff
@@ -41,6 +41,8 @@ void RideDBerror(void*jc, const char *error) // used by parser aka yyerror()
 
 %}
 
+%define api.header.include {"RideDB_yacc.h"}
```
liversedge, line 44: Again, this should not be included in this patch

src/FileIO/FitRideFile.cpp
```diff
@@ -3080,10 +3090,10 @@ struct FitFileReaderState
...
         swimXdata->unitname << "";
 
-	hrvXdata = new XDataSeries();
-	hrvXdata->name = "HRV";
-	hrvXdata->valuename << "R-R";
-	hrvXdata->unitname << "msecs";
+        hrvXdata = new XDataSeries();
+        hrvXdata->name = "HRV";
+        hrvXdata->valuename << "R-R";
+        hrvXdata->unitname << "msecs";
```
liversedge, line 3096: This reformatting will break git blame- should be removed

src/FileIO/FitRideFile.cpp
```diff
@@ -4034,8 +4060,8 @@ uint16_t crc16(char *buf, int len)
 
 void write_header(QByteArray *array, quint32 data_size) {
     quint8 header_size = 14;
-    quint8 protocol_version = 16;
-    quint16 profile_version = 1320; // always littleEndian
+    quint8 protocol_version = 20;
+    quint16 profile_version = 2173; // always littleEndian
```
liversedge, line 4064: did the format spec change ?

src/Train/RealtimeData.cpp
```diff
@@ -43,694 +43,624 @@ RealtimeData::RealtimeData()
...
-{
-    this->altWatts = (int)watts;
-}
-void RealtimeData::setWatts(double watts)
-{
-    // negs not allowed (usually from virtual power)
-    if (watts < 0) watts = 0;
+void RealtimeData::setName(char *name) { strcpy(this->name, name); }
+void RealtimeData::setAltWatts(double watts) { this->altWatts = (int)watts; }
+void RealtimeData::setWatts(double watts) {
```
liversedge, line 48: reformatting will break git blame, but in this case the file doesn't change so often and this is in line with the rest of the code style so is ok

src/Train/RealtimeData.cpp
```diff
@@ -43,694 +43,624 @@ RealtimeData::RealtimeData()
...
-void RealtimeData::setLapMsecsRemaining(long x)
-{
-    this->lapMsecsRemaining = x;
-}
+void RealtimeData::setHr(double hr) { this->hr = (int)hr; }
+void RealtimeData::setSpeed(double speed) { this->speed = speed; }
+void RealtimeData::setWbal(double wbal) { this->wbal = wbal; }
+void RealtimeData::setVirtualSpeed(double speed) { this->virtualSpeed = speed; }
+void RealtimeData::setWheelRpm(double wheelRpm, bool fMarkWheelRpmTime) {
+  this->wheelRpm = wheelRpm;
```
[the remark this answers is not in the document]
liversedge: On the whole we really should look to get this merged and my comments are for feedback but not neccessarily needed to action- although it would be good to remove the yacc header changes which should not be in this patch.

Realtime_Data.cpp changes seemed to be 90% reformatting and there is a LOT of that. Its ok to let that ride, but maybe don't get too carried away in future- the fact it breaks git blame seems trivial but when you're chasing down bugs it can be quite annoying !

ihewitt:
Ignore! Let me change to WIP... just rebasing.

src/Train/RealtimeData.cpp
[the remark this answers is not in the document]
ihewitt: Thanks, yeah please ignore for a bit! I'm just getting the rebase sorted out and I'll get rid of formatting changes etc when I'm done.

amtriathlon:
Hi Ivor, if you want to build using Bison 3.7 or higher see https://github.com/GoldenCheetah/GoldenCheetah/commit/5a771809579c619aefdd52d3ea596f0a2372f315 to avoid breaking compatibility with older versions.

ihewitt:
@amtriathlon patch tidied up, working well - this should be good to review/go in now.

amtriathlon:
Thank you!

PR #4209, opened 2022-04-16
SimRider V1. Ride live map against existing activities.

human705:
This PR adds simulated riders to GC and builds on the SimRider code created by Erik. In V1, the capability to workout against previous rides on the same route was added. Working on V2 to add AI capability to allow the simulated rider to pace and attack at random intervals on any route. 

The feature is based on the existence of the route tag in the activities folder of the athlete. The tag value (route name) will be read from the selected workout and used to find matching activities from the athlete activities folder. If any routes are found, they will be sorted by time so the fastest one is first. If more than 4 are found, the top 4 will be selected and displayed in the chart. It is recommended that you zoom in the live map to see all markers. If 4 or less are found, they will all displayed in the chart. If no files are found with the specific route tag, only the current user will be displayed.

The following steps are required in order for the feature to function as designed:

        1. Simulate Speed from Power must be enabled in the preferences section in options-> training-> Preferences
        2. Previous activities must have power (Watts) data. If they don't, the feature will be disabled.
        3. Review existing activities and add the same route name to all activities created using the same workout.
        4. Export one of the tagged activities as a GC json file in the athlete workouts folder.
        5. Add the workouts folder in the search path for workouts, video files and videosync files.
        

Don't forget to "Save settings" after changing configurations to apply the changes before the workout starts.

Here is a sample view of the new chart
GC SimRiderV1


amtriathlon:
Thank you for your contribution, but I think this is no ready to be merged.

PR #4214, opened 2022-04-26
Fix for DP #4095 - Python Fixes work on auto import (save option)

paulj49457:
This will allow the Python Fixes scripts to work on auto import (save option), resolving #4095

During Auto import, the Data processors are called twice for each imported activity, once for the "Import" setting and once for the "Save" settings. Unfortunately the "Import" data processing call occurs before the RideFile context is configured (at this point it set to the previous ride if it exists), but when they are called by the "Save" option at the end of the processing the context is correct for the ride being imported, BUT the code didn't pass the context to the Python Data processor.

I don't want to try and restructure the AutoImportWizard to try and fix the "Import" case, when the "Save" option works correctly (when this PR is applied). This PR now passes the context to the Python data processor. 

The resulting auto import functionality for Python Data Processors is:
i) "Import" option: will process all imported activities except the last one  :(
ii) "Save" option: will process all imported activities  :)

I have tested this and it works fine, but there maybe unintended consequences of this approach, so it needs review by GC expert.

### DIFF
--- a/src/FileIO/FixPyDataProcessor.cpp
+++ b/src/FileIO/FixPyDataProcessor.cpp
@@ -23,7 +23,7 @@ bool FixPyDataProcessor::postProcess(RideFile *rideFile, DataProcessorConfig *se
 
     QString errText;
     bool useNewThread = op != "PYTHON";
-    Context* context = (op != "DELETE" && rideFile) ? rideFile->context : nullptr;
+    Context* context = (rideFile) ? rideFile->context : nullptr;
     FixPyRunner pyRunner(context, rideFile, useNewThread);
     return pyRunner.run(pyScript->source, pyScript->iniKey, errText) == 0;
 }
--- a/src/FileIO/FixPyDataProcessor.cpp
+++ b/src/FileIO/FixPyDataProcessor.cpp
@@ -23,7 +23,8 @@ bool FixPyDataProcessor::postProcess(RideFile *rideFile, DataProcessorConfig *se
 
     QString errText;
     bool useNewThread = op != "PYTHON";
-    FixPyRunner pyRunner(nullptr, rideFile, useNewThread);
+    Context* context = (op != "DELETE" && rideFile) ? rideFile->context : nullptr;
+    FixPyRunner pyRunner(context, rideFile, useNewThread);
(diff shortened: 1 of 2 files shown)


amtriathlon:
I like the idea to pass the context to Python DPs when available, I think this is a clean fix for the save case, and I see no reason to exclude the delete case since deletion happens afterwards (currently Python DPs have no access to the op parameter, but it could be potentially useful).
OTOH, I don't like the idea to remove the import option, since it works when you only need to change/add data series, provided no metadata is accessed, and it has the advantage to run only once, no on every activity save. Until we find a proper solution for this I would let this unchanged.

paulj49457:
Thank you for the feedback,

So after re-testing with the DELETE guard removed there are no issues
I have backed out the changes to the import/save/manual menu in Pages.cpp
I have updated the description explaining the behaviour after the PR is applied, so now it's just a two line change :)

The resulting auto import functionality for Python Data Processors is:
i) "Import" option: will process all imported activities except the last one :(
ii) "Save" option: will process all imported activities :)

It would be great if this could make the v3.6 cut-off

amtriathlon:
Thank you.

peret2000:
Testing it. Import Dataprocessors seem to run ok for all the activities but the last. That is known.
Should we configure python scripts to automatically run on 'Save' instead of 'Import' to have all the activities processed? I do not know if there are other side effects in such case.

OTOH, it is a curiosity, but changes in metada made by the dataprocessors that are running during import are not seen in 'Details' tab. Their values are correct, as they can be seen in the json file of the imported activity, and also they can be seen in the mail tab if configured to show those metadata; but in 'Details' tab, the fields are empty.
As a simple example, I run a data processor that sets 'SubSport' to 'VirtualRun' in case 'Sport' is 'Run' and there are no GPS data. The result, if executed manually, is ok: SubSport is 'VirtualRun' in mail tab and in Details tab.
However, importing the activity, as long as it is not the last imported activity, SubSport is also set to 'VirtualRun', but in Details, SubSport is empty, even closing and reopening GC. I have not found the issue, I do not know if it is something to look at, or just a worthless stetic effect.
Using the 'test' code  #4095, the behavior is different: metadata fields are updated in main tab and Details tab.

Anyway, thanks for the implementation

PR #4219, opened 2022-05-13
Workout chart support for imperial units

human705:
These changes will allow the workout chart to show distance in kilometers and miles based on the user options.

### DIFF
--- a/src/Train/ErgFilePlot.cpp
+++ b/src/Train/ErgFilePlot.cpp
@@ -537,7 +537,7 @@ ErgFilePlot::setData(ErgFile *ergfile)
                 setAxisScale(xBottom, (double)0, maxX, step);
                 QwtText title;
                 title.setFont(stGiles);
-                title.setText("Distance " + ((GlobalContext::context()->useMetricUnits) ? tr("(Km)") : tr(" Mi")));
+                title.setText("Distance " + ((GlobalContext::context()->useMetricUnits) ? tr("(km)") : tr("(mi)")));
                 QwtPlot::setAxisFont(xBottom, stGiles);
                 QwtPlot::setAxisTitle(xBottom, title);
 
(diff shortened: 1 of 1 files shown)


src/Train/ErgFilePlot.cpp
```diff
@@ -523,21 +524,21 @@ ErgFilePlot::setData(ErgFile *ergfile)
 
         // set the axis so we use all the screen estate
         if (context->currentErgFile() && context->currentErgFile()->Points.count()) {
-            double maxX = (double)context->currentErgFile()->Points.last().x;
+            double maxX = -1;
```
amtriathlon, line 527: This change is initialization is irrelevant for the bydist case after your other changes, and potentially wrong otherwise, it should be removed.

src/Train/ErgFilePlot.cpp
```diff
@@ -523,21 +524,21 @@ ErgFilePlot::setData(ErgFile *ergfile)
...
                 if (maxX <= 1000) step = 100;
                 else if (maxX < 5000) step = 1000;
 
                 // axis setup for distance
                 setAxisScale(xBottom, (double)0, maxX, step);
-
                 QwtText title;
                 title.setFont(stGiles);
-                title.setText("Distance (km)");
+                title.setText("Distance " + ((GlobalContext::context()->useMetricUnits) ? tr("(Km)") : tr(" Mi")));
```
amtriathlon, line 540: Changing capitalization voids translations, besides the use of () should be consistent

amtriathlon:
I don’t know how much you have tested this, but it looks incomplete and wrong.
For example: x/y values are not converted (see the bridge a the start), laps marked positions are not adjusted, and x-axis mixing of feets/miles likely will not work.

human705:
Thx Ale. I missed the laps as none of my workouts have any. I'll close this and re-submit when it is done.

PR #4225, opened 2022-05-24
Wourkout Chart with imperial unit support

human705:


amtriathlon:
Have you tested these changes with ERG type workouts, or at least considered how they will be affected?

human705:
I did run a test, but I must've forgotten to switch from metric to imperial. It appears that the lap creation code will run without checking "bydist".  I added a check to validate the mode for plotting both correctly. I will push in a few.

amtriathlon:
> I did run a test, but I must've forgotten to switch from metric to imperial. It appears that the lap creation code will run without checking "bydist". I added a check to validate the mode for plotting both correctly.

Not just lap creation, the "bridges" providing x data also convert units without checking it is distance when imperial is choosen.

human705:
Thanks Ale. What would be the preferred approach? Adding the distance check to the bridge providing x or pass the "bydist" as a parameter to the bridge?

amtriathlon:
I am not sure what is your question, but you could use (context->currentErgFile()->format == CRS) to compute bydist on the fly without a new member.

human705:
In the suggestion above you had mentioned that you would add a bydist member to both classes, ErgFileData and NowData. I added the member and the setter in the NowData class only and everything seems to work. Was there a reason that you would add the member in the ErgFileData class?

amtriathlon:
For the same reason you use bydist in other places: to distinguish when x/y are distance/altitude, and eventually need to be converted from metric to imperial, and when are time/power and should not be changed.

Issue #4223, opened 2022-05-18
Update Batch export... to Batch Processing....


paulj49457:
Rename the access menu option from  "Batch export..." to  "Batch Processing....".

  Rename the "Activity Batch Export" dialog to "Batch Processing", then add a radio box at the top with the following four options:

    i) Export as    xxxxxxx     to    xxxxx  Browse  overwrite  xx
 
    ii) Run data processor xxxxxxxxx 

    iii) Set metadata field (drop down?) to value (constant or formula?) for all selected

    iv) Delete all selected

update the Action column to display the option selected and change the Export button to Run or Execute.

paulj49457:
So I have "delete all selected" working, and the dialog looks like:

![CaptureDial](https://user-images.githubusercontent.com/46629337/169643676-c63cbf6e-d39a-4804-aa96-d1516856dd94.PNG)

So I have a few questions:

i) When the data processors are run from here, are we expecting them to use the configuration defined in the Options->Processing menu? Or should I make it optional to use the config or not?

ii) Meta Data, can you tell me how many fields I should be expecting? Should it be the same list as in Options->Fields menu's Field column ? Also can you point me in the direction of the type of meta data processing that already exists that you would like to be made available?

paulj49457:
To reduce confusion I think the "Settings" column heading (circled in red) should be renamed to "Auto & Save Settings", as the manual settings are entered in the individual data processor dialogs, what do you think ?

![Capture123](https://user-images.githubusercontent.com/46629337/169644991-fa02f1dd-b5fb-42b2-b921-089f94f952f7.PNG)

amtriathlon:
> To reduce confusion I think the "Settings" column heading (circled in red) should be renamed to "Import, Save & Batch Processing Settings", as the manual settings are entered in the individual data processor dialogs, what do you think ?

Those settings are also used as defaults for interactive DP execution, perhaps "Default Settings" would be clear enough.

paulj49457:
Hi Ale,

   Thank you for the feedback to my questions, but in the meantime I have created a PR for a version of the batch processing and submitted it #4226. So I'll look at the Meta Data info you have provided, and maybe you can try the PR I have submitted to see if anything else is required.

   As to the check boxes they are exclusive radio buttons.

Paul.

amtriathlon:
> As to the check boxes they are exclusive radio buttons.

[...]

PR #4226, opened 2022-05-25
Batch processing

paulj49457:
This PR addresses the “Update Batch export... to Batch Processing.... #4223” issue.

The new BatchProcessingDialog that replaces and extends the BatchExportDialog is as follows:

CaptureABC

**BatchProcessingDialog.[cpp/h]**
New dialog supporting:

- Batch Export (existing functionality)
- Batch core data processing
- Batch Python data processing
- Delete all selected

The reason for separating out the core and python processors is the addition of a Configure and Edit button respectively. These buttons allow the User direct access to edit the selected python data processor, or at least to view to decide whether it be applied to a batch of activities. Whilst the Configure button provides direct access to the core data processor configuration menu, so the parameter values can be set or confirmed before a batch application of the processor.

**Meta Data Processor:**
There was originally a request for a meta data processor option, but knowing what functions could be generally useful is difficult, and when simple powerful python fix processors can be written to change meta data fields there seemed little point. For example my Wahoo Elemnt Bolt V2 reports the device as “Wahoo fitness 43” which isn’t that obvious, so I have a Python Fix to correct it:

```
if (GC.activityMetrics()['Device'] == "Wahoo fitness 43"):
         GC.setTag("Device", "Elemnt Bolt v2")
```

Therefore I think most Meta Data changes can be addressed by simple Python Fixes which are easy enough to write for non-programmers. I’m happy to write a page in the Wiki if it would help?

**BatchExportDialog.[cpp/h]**
This dialog has been deleted, as the changes to implement were a significant rewrite, although the actual Batch Export code has been ported untouched.

**ProcessorPage class in Pages.[cpp/h]**
The data processors are now ordered into core & python processors, and displayed with the core ones at the top of the list and the python ones at the bottom. The rationale for this is the python processors do not have any configuration to modify, and the import/save/manual option is only infrequently changed. Also now when display directly from the Batch Processing dialog the ones the User is going to want to change are the core processor configuration values.

**RideCache.[cpp/h]**
The RideCache interface has an additional removeRide function which allows an activity to be deleted by filename to support the Batch Processing delete operation. The rationale for this addition is that selecting an activity to then delete it will be slow and cause excessive screen disturbance.

**Code Review**

ProcessorPage class in Pages.[cpp/h]  and  RideCache.[cpp/h]
It’s best to view the modified file rather than the git diffs, as diff seems pretty poor at providing a sensible output, and overstates the amount of change IMHO.

BatchProcessingDialog.[cpp/h]
•	Please check I have the Qt memory collection correct.
•	Please check whether I need to delete the temporary QDialog I have created in the CoreDpOptionsDialog class to hold the data processor ProcessorPage class items.

**Testing**
The Batch Processing Dialog has been tested in the following combinations:

- Complied and test with GC_WANT_PYTHON compiler directive
-- Tested with no Python Path defined
-- Tested with Python Path defined
--- Tested with no Python Fixes defined
--- Tested with Python Fixes defined


paulj49457:
I'll update this PR with the additional option of a Meta Data processor, so please wait for these changes before reviewing and processing this PR.

paulj49457:
All ready to be reviewed.

amtriathlon:
Great job, I think the general idea is good, but data processors and metadata options look too complicated.
I reviewed the Data Processor option and found it can be simpler -and likely more functional- if we model it after Edit menu behavior, I will push the changes for testing. I haven't looked at metadata yet.

paulj49457:
I'd have preferred to have kept the options to quickly configure the core data processors and edit the python fixes in the original design. Since the data processors have been re-enabled I have been using the latest GC version with the original PR added to tidy and correct my data, and I did modify my own python data processors for differently selected sets of data and it was quick and easy.

batch_Capture

Now I will have to navigate through to the python fixes menu, edit the fix, and then return back to the batch processing, also it was a quick way to confirm what updates I would be running, and similarly I think the quick way to adjust/check the core data processor configuration was user friendly.

IMHO I didn't see the original options as complex, rather it clarified what could be done, core data processors could be configured, while python fixes could be edited.

paulj49457:
> Great job, I think the general idea is good, but data processors and metadata options look too complicated. I reviewed the Data Processor option and found it can be simpler -and likely more functional- if we model it after Edit menu behavior, I pushed the changes for testing. I haven't looked at metadata yet.

Can we at least retain the option to open the pyfixes window to confirm/edit the changes that would be made by running the Data processor? Maybe a button that becomes enabled once a pyfix has been selected, to access the pyfixes window?  I can live without the core data processor configuration option, but the pyfixes I find really useful.

paulj49457:
Thank you for reinstating the pyfixes edit button.

amtriathlon:
> Thank you for reinstating the pyfixes edit button.

YW. WRT core processors parameters you can change them before to run the processor, and optionally save those parameters as defaults if you want to, no need to go to config for this operation purposes.

amtriathlon:
I removed metadata options to be able to squash and merge the options I think we can incorporate to v3.6, they could be reinstated afterwards or be included in a new PR.

amtriathlon:
You are right, buttons look better left aligned. I will merge this, thank you!

PR #4241, opened 2022-07-06
Make GCColor & Themes into Singletons

paulj49457:
GCColor has been defined with many static functions and accesses entirely static data, so is really a prime candidate for a Singleton pattern within the GC code. Therefore I have updated it to be a singleton, added defensive code, and additional overloaded function access so GColor macro doesn't always have to be used when accessing luminance, alternateColor, etc.

Themes is another candidate for a Singleton pattern, and its access is now independent of GCColor, as this was just providing a pass through access, and I have removed GCColor's unused Themes list to tidy the code.

This PR does NOT change any GCColor or Themes functionality in normal operation, although it does now defend against out of range colornum values in access functions, returning a defined out of range color.

This PR does impact numerous files accessing GCColor (and there are many!), but the changes are simple word replacements, so it looks far worse than it actually is, GCColor.[h & cpp] are where the changes have been made.

**NOTE:**
 I noticed two color definitions are missing from the LightDefaultColorList, I'm not sure whether this was intentional, but I've added a comment in the code to highlight this possible issue, but I've no idea what sensible values should be for these.

    LightDefaultColorList[102].color = QColor(134,74,255); // 102:Respiratory Frequency
    LightDefaultColorList[103].color = QColor(255,46,46); // 103:FeO2
    // CHOVER (104) are missing definitions !
    // CCHARTBAR (105) are missing definitions !
    LightDefaultColorList[106].color = QColor(180,180,180); // 106:Tile Alternate
    LightDefaultColorList[107].color = QColor(238,248,255); // 107:Tile Vibrant

### DIFF
--- a/deprecated/BingMap.cpp
+++ b/deprecated/BingMap.cpp
@@ -170,7 +170,7 @@ void BingMap::createHtml()
     }
 
     QColor bgColor = GColor(CPLOTBACKGROUND);
-    QColor fgColor = GCColor::invertColor(bgColor);
+    QColor fgColor = GInvertColor(bgColor);
 
     // No GPS data, so sorry no map
     if(!ride || !ride->ride() || ride->ride()->areDataPresent()->lat == false || ride->ride()->areDataPresent()->lon == false) {
--- a/deprecated/GcScopeBar.cpp
+++ b/deprecated/GcScopeBar.cpp
@@ -190,7 +190,7 @@ GcScopeBar::paintBackground(QPaintEvent *)
     QRect all(0,0,width(),height());
 
     // fill with a linear gradient
-    QLinearGradient linearGradient = GCColor::linearGradient(23*dpiYFactor, isActiveWindow());
+    QLinearGradient linearGradient = GCColor::instance()->linearGradient(23*dpiYFactor, isActiveWindow());
     painter.setPen(Qt::NoPen);
     painter.fillRect(all, linearGradient);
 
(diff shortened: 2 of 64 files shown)


amtriathlon:
We are open to functional contributions (bug fixes or new features), refactoring are not that welcome, since this is a large long standing project with code from multiple contributors and refactoring can introduce new bugs for no appreciable benefit.
We have limited resources for reviewing, testing, deploying, support, etc. to add additional mostly unnecessary charges.
The fact you update deprecated folder seems to indicate you don't even understand the base code well enough, please don't change what you don't understand.

paulj49457:
> We are open to functional contributions (bug fixes or new features), refactoring are not that welcome, since this is a large long standing project with code from multiple contributors and refactoring can introduce new bugs for no appreciable benefit. We have limited resources for reviewing, testing, deploying, support, etc. to add additional mostly unnecessary charges. The fact you update deprecated folder seems to indicate you don't even understand the base code well enough, please don't change what you don't understand.

I assumed the "deprecated" software code referred to functions or elements that are in the process of being replaced by newer ones, and would be still valid in some cases. If it's not used, then it should be removed from the repo to prevent confusion. As for the refactoring I understand the review issue (and I'm willing to help), but not refactoring will mean the project will accumulate technical debt, and maintenance will become more problematic. I suggested the singleton as a starting point to centralise the colour management, and therefore simplify its long term maintenance.

amtriathlon:
It is easy to check deprecated code is not used in https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/src.pro

Anyway, if you are so uncomfortable with the way this project is managed, perhaps you can consider forking instead of insisting on try to force your criterium on each PR.

Issue #4201, opened 2022-03-22
Default Dark Theme Setting has some almost invisible items


schubidoo:
When Running GC (Version 3.6 RC, Feb. 1, 2022) with the system set to dark mode, some things in the user interface become basically invisible. Particularly, I saw the filter markers in the "Trends" Tab (Which are "black on very dark"), and the "invisible" text in the filter dialog (also black on very dark).

See the attached screenshot, this is on a standard Win 10 20H2 install (19042).

![grafik](https://user-images.githubusercontent.com/17938073/159447396-2a602b97-c63d-4c0b-87c0-9d938237ed31.png)

paulj49457:
I'm thinking that rather the current 8 or so pre-programmed themes it would be good if these were changed to be a set of files in a config directory. The GC installation would come with the 8 themes in file format which could be selected and loaded, also the user could save the values currently set in the colours tab as a file. This would make it easier to correct colour issues in GC themes, plus make it possible for people to create new themes and share them. If this would be generally useful I'm happy to work on improving this area.

paulj49457:
I've been looking at this and I'm going to code a change to move the currently hardcoded themes definitions to an XML file so new ones can easily be created. I have also created a snapshot feature that writes the current colour settings to an XML file, so they can be reloaded assuming the user has defined a custom set they want to keep.

liversedge:
Please don't do this. It will not be accepted.

paulj49457:
Putting the hard coded themes in an xml file parsed at start-up. why would this not be acceptable?

It would be before all the other modifications, it would just replace the code sections such as:

    // MODERN LIGHT (SAP Fiori Belize inspired)
    add.dark = false;
    add.stealth = false;
    colors << QColor(Qt::white)  // Plot Background
           << QColor(0xef,0xf4,0xf9) // Toolbar and Sidebar Chrome
           << QColor(0x26,0x84,0xf6) // Accent color (markers)
           << QColor(Qt::blue) // Selection color
           << QColor(Qt::darkMagenta) // Critical Power and W'Bal
           << QColor(Qt::red) // Heartrate
           << QColor(85,170,0) // Speed
           << QColor(255,170,0) // Power
           << QColor(0,204,204) // Cadence
           << QColor(Qt::magenta) // Torque
           << QColor(0xcb,0xdc,0xea) // Overview Background
           << QColor(255,255,255) // Overview Tile Background
[...]

PR #4242, opened 2022-07-06
Configurable Themes & Backup Colour Settings

paulj49457:
**Configurable Themes**

This PR moves the original eight defined Themes that are hardcoded into GC into a configuration file stored in the GCRoot directory. This will enable new themes to be easily created, and corrections to existing themes to be made. This only replaces the themes definition aspect, GC has internal code which defines how the themes are applied and there are some platform dependent override cases, these are not altered by this PR.

The new themes configuration file which needs to be in GCRoot is:
https://drive.google.com/drive/folders/17bL9G6DMg-T8JfGvd1FnvRMmyhtXrQlk?usp=sharing

I have expanded the themes swatch to display all 13 colors in the theme color definitions, which makes it easier to compare the differences between the themes before applying them:

Capture1

**Backup Colors**

When experimenting with the 109 possible colors in the Appearance->Colors tab there is no way to return to initial settings once "save" has been pressed, this is especially true if a theme was applied and then further modifications have been made in the Colors tab. So this PR introduces two buttons into the Colors tab which allow the current Color settings to be written to a "color-backup.xml" file in the GCRoot directory, and for the backup color settings to be imported.

Note1: The backup exports the colors currently defined in the Color tab, these maybe different to what GC is displaying if they haven't been applied by pressing "save".
Note2: Importing the color backup requires the user to press "save" to make them permanent within GC.

This change will also support copying Colors from one GC installation to another, the colors just need to be backed up, the color-backup.xml file copied to the same location in another GC installation, and then imported, and "save" pressed on the appearance dialog.

An example of the color-backup file created by the backup process (and written to GCRoot) can be found in:
https://drive.google.com/drive/folders/17bL9G6DMg-T8JfGvd1FnvRMmyhtXrQlk?usp=sharing

Capture2

Should one element of this PR be considered unacceptable, let me know and I can separate them, but I think they are both useful additions to support the user configurability of GC.

### DIFF
--- a/src/Gui/Colors.cpp
+++ b/src/Gui/Colors.cpp
@@ -25,6 +25,10 @@
 #include <QByteArray>
 #include <QDir>
 #include "Settings.h"
+#include <vector>
+#include <string>
+#include <sstream>
+#include <iostream>
 
 #ifdef Q_OS_WIN
 #include <windows.h>
@@ -87,9 +91,6 @@ static bool initStandardColors()
 }
 static bool init = initStandardColors();
 
-// the standard themes, a global object
-static Themes allThemes;
-
 // Number of configurable metric colors + 1 for sentinel value
 static Colors ColorList[CNUMOFCFGCOLORS+1],
               LightDefaultColorList[CNUMOFCFGCOLORS+1],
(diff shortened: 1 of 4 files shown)


amtriathlon:
https://github.com/GoldenCheetah/GoldenCheetah/issues/4201#issuecomment-1172876952

PR #4257, opened 2022-08-17
Athlete tab colour fix

paulj49457:
Fixes issue in discussion of v3.6: https://groups.google.com/g/golden-cheetah-users/c/Wlp4pBMxbAg

### DIFF
--- a/src/Gui/Colors.cpp
+++ b/src/Gui/Colors.cpp
@@ -444,6 +444,22 @@ QColor GCColor::alternateColor(QColor bgColor)
         return QColor(Qt::lightGray);
 }
 
+QColor GCColor::selectedColor(QColor bgColor)
+{
+     // if foreground is white then we're "dark" if it's
+     // black the we're "light" so this controls palette
+     bool dark = invertColor(bgColor) == QColor(Qt::white);
+     bool isblack = bgColor == QColor(Qt::black); // e.g. mustang theme
+
+     // on select background color
+     QColor bg_select = bgColor;
+     if (dark) bg_select = bg_select.lighter(200);
+     else bg_select = bg_select.darker(200);
+     if (isblack) bg_select = QColor(30, 30, 30);
+
+     return bg_select;
+}
+
 const Colors * GCColor::colorSet()
 {
     return ColorList;
--- a/src/Gui/Colors.h
+++ b/src/Gui/Colors.h
@@ -130,6 +130,7 @@ class GCColor : public QObject
         static double luminance(QColor color); // return the relative luminance
         static QColor invertColor(QColor); // return the contrasting color
         static QColor alternateColor(QColor); // return the alternate background
+        static QColor selectedColor(QColor); // return the on select background color
         static QColor htmlCode(QColor x) { return x.name(); } // return the alternate background
         static Themes &themes(); 
         static void applyTheme(int index);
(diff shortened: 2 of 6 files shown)


liversedge:
Its not an attractive solution. Need to look more closely at the selected color and bring it in line with the theme.

paulj49457:
Thanks for the feedback.

I have updated the code to track the selected colors in NewSideBar, is this better ?

IMHO classes within GC shouldn't need code such as in their MainWindow::configChanged(qint32) :

    // set foreground colors
    QColor fg_normal = GCColor::invertColor(GColor(CCHROME));

    // if foreground is white then we're "dark" if its
    // black the we're "light" so this controls palette
    bool dark = (fg_normal == QColor(Qt::white));

    // set background colors
    QColor col = GColor(CCHROME);
    bool isblack = (col == QColor(Qt::black)); // e.g. mustang theme
    QColor bg_normal = col;

    // on select
    QColor bg_select = GColor(CCHROME);
    if (dark) bg_select = bg_select.lighter(200);
    else bg_select = bg_select.darker(200);
    if (isblack) bg_select = QColor(30, 30, 30);
    QColor fg_select = GCColor::invertColor(bg_select);

The color set should be updated centrally elsewhere and then the component just use the colours, having comments about particular themes and how to handle them means it's difficult to maintain consistency. The NewSideBar component for example determining that it should be 200 darker or lighter seems wrong to me and maintenance headache.

paulj49457:
I see that once a theme is applied, CCHROME, CCHARTBAR & CTOOLBAR are all set to the same value, so maybe the code in should in MainWindow should use CTOOLBAR rather than CHROME for the case when the user has set them to different values and not applied a theme, I'll update the PR.

           ColorsPage::applyThemeClicked()

            ......

            case CCHROME:
            case CCHARTBAR:
            case CTOOLBAR: // we always keep them the same, but user can make different
                //  set to black for dark themese and grey for light themes
                color = theme.colors[1];
                break;

liversedge:
The bug in question is windows specific- the color choices reflect the theme- so if a user is unhappy with the color contrast of a dark stealth theme they should probably use a theme with a greater contrast.

The fix required is related to the way the style is cascaded on windows- unfortunately I am no longer able to build on windows so it is very difficult for me to check changes- but I can check if mac/linux builds are affected.

amtriathlon:
I can confirm this issue is Windows specific, I cannot reproduce it on Linux (both using Qt5.15.2) or macOS (Qt5.14.1).

Interestingly the annoying compare pane behavior (disappear when resizing and don't accept activities) is also Windows specific in my testing.

paulj49457:
I'm developing on Windows 10, and will take a look at : https://github.com/GoldenCheetah/GoldenCheetah/issues/4201 to se whether I can resolve this issue and add it to this PR.

paulj49457:
PR updated to fix https://github.com/GoldenCheetah/GoldenCheetah/issues/4201 for dark themes

amtriathlon:
It works on Windows, I will squash and merge this for better testing on remaining platforms.
We can revert otherwise.

liversedge:
image
This is horrible

liversedge:
Have fixed for Linux in 536057b

Will look at MacOS later today.

In general these changes need to be tested on all 3 platforms since local style rules will conflict with what we try to acheive. Fiddling with style sheets needs to be done with consideration for all platforms.

liversedge:
Additionally, search boxes are placed on dialog widgets and so we cannot assume the background color will be to toolbar color. This also needs to be considered and handled.

Issue #4249, opened 2022-07-24
Datafilter memory leak


liversedge:
Datafilters allocate a lot of memory and are never cleared, fortunately the data they manipulate and create are stored in Result objects that are created and deleted appropriately (generally on the stack).

Some of the more obvious issues:
* clearFilter never gets called in practice, there is no DataFilter::~DataFilter method for a start
* the leaf::clear() method is out of date and does not reflect the multitude of new types of Leaf, nor how the lvalue/rvalue are used/not used
* the runtime object is manipulated by UserMetrics in a spooky action at a distance way (that needs to be addressed)

There is likely a number of other issues hidden in the code- this defect needs to be reviewed in detail.

One example of how to test this:
* add a DataFilter destructor that calls clearFilter
* switch to trends view
* a SEGV will occur since SearchFilterBox::matches allocates a Datafilter on the stack, causing the delete method to be called as it goes out of scope

amtriathlon:
There is a reproducible crash in Leaf::clear reported at the forum: https://groups.google.com/g/golden-cheetah-users/c/rYZOMnYeltg/m/kXbtfoH7AgAJ

amtriathlon:
I think #4277 (almost) fixes this issue, but I am not sure about the runtime object part.

Recently I've found a crash related to calling clearFilter from DataFilter destructor, it happens when a compound statement has a syntax error, it is easy to reproduce when you edit User Charts trying to add a new method, you start to type the name and GC crashes.

The problem is a DataFilter is created in the stack by DataFilterEdit::checkErrors but, since the parse is incomplete, treeRoot points to a QList<Leaf*> instead to a Leaf* and clearFilter calls Leaf::clear incorrectly.

This can be avoiding setting treeRoot to NULL when there are parsing errors, s.t. the DataFilter constructor does when called with a script, in DataFilter::check and DataFilter::parseFilter, the price is memory will not be released in those cases, but this still is better than never to release it.

I will add this patch to avoid crashes, but ideas to distinguish the cases for properly delete on parsing errors are welcome.

PR #4277, opened 2022-10-11
clearFilter - release memory avoiding crashes

amtriathlon:
clearFilter - delete root Leaf
Leaf constructor: initialize left,right and cond to null pointers Leaf::clear avoid crashes when called with a null pointer parameter, and release memory in all cases allocated in the parser
TODO: still crashes if called from DataFilter destructor, see #4249

### DIFF
--- a/src/Core/DataFilter.cpp
+++ b/src/Core/DataFilter.cpp
@@ -1642,23 +1642,47 @@ bool Leaf::isNumber(DataFilterRuntime *df, Leaf *leaf)
 
 void Leaf::clear(Leaf *leaf)
 {
-    return; // TODO: fix this function to release memory in all Leaf cases avoiding crashes
+    if (leaf == NULL) return; // critical to avoid crashes
 
     switch(leaf->type) {
+    case Leaf::Script :
     case Leaf::String : delete leaf->lvalue.s; break;
     case Leaf::Symbol : delete leaf->lvalue.n; break;
     case Leaf::Logical  :
     case Leaf::BinaryOperation :
     case Leaf::Operation : clear(leaf->lvalue.l);
                            clear(leaf->rvalue.l);
-                           delete(leaf->lvalue.l);
-                           delete(leaf->rvalue.l);
+                           delete leaf->lvalue.l;
+                           delete leaf->rvalue.l;
+                           break;
+    case Leaf::UnaryOperation : clear(leaf->lvalue.l);
+                           delete leaf->lvalue.l;
                            break;
     case Leaf::Function :  clear(leaf->lvalue.l);
-                           delete(leaf->lvalue.l);
-                            break;
-    default:
-        break;
+                           delete leaf->lvalue.l;
+                           clear(leaf->series);
+                           delete leaf->series;
+                           foreach (Leaf* l, leaf->fparms) clear(l);
+                           leaf->fparms.clear();
+                           break;
+    case Leaf::Compound :  foreach (Leaf* l, *(leaf->lvalue.b)) clear(l);
+                           delete leaf->lvalue.b;
+                           break;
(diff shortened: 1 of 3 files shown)


src/Core/DataFilter.cpp
```diff
@@ -1642,23 +1642,47 @@ bool Leaf::isNumber(DataFilterRuntime *df, Leaf *leaf)
...
                            break;
     case Leaf::Function :  clear(leaf->lvalue.l);
-                           delete(leaf->lvalue.l);
-                            break;
-    default:
-        break;
+                           delete leaf->lvalue.l;
+                           clear(leaf->series);
+                           delete leaf->series;
+                           foreach (Leaf* l, leaf->fparms) clear(l);
```
liversedge, line 1665: should l be deleted after being cleared?

src/Core/DataFilter.cpp
```diff
@@ -1642,23 +1642,47 @@ bool Leaf::isNumber(DataFilterRuntime *df, Leaf *leaf)
...
+                           leaf->fparms.clear();
+                           break;
+    case Leaf::Compound :  foreach (Leaf* l, *(leaf->lvalue.b)) clear(l);
```
liversedge, line 1668: again, l deleted after being cleared?

src/Core/DataFilter.cpp
```diff
@@ -1642,23 +1642,47 @@ bool Leaf::isNumber(DataFilterRuntime *df, Leaf *leaf)
...
+                           clear(leaf->rvalue.l);
+                           clear(leaf->cond.l);
+                           delete leaf->lvalue.l;
+                           delete leaf->rvalue.l;
+                           delete leaf->cond.l;
+                           break;
+    case Leaf::Index :
+    case Leaf::Select :    clear(leaf->lvalue.l);
+                           delete leaf->lvalue.l;
+                           foreach (Leaf* l, leaf->fparms) clear(l);
```
liversedge, line 1681: delete after clear?

src/Core/DataFilter.cpp
amtriathlon: Yes, I wrongly assumed clearing/deleting the container should free the elements, I will fix it.

amtriathlon:
> TODO: still crashes if called from DataFilter destructor, see #4249

Last added commit fixes the crash I could reproduce when adding a DataFilter destructor calling clearFilter to release memory.

src/Core/DataFilter.y
```diff
@@ -458,6 +458,8 @@ expr:
                                                   $1->type = Leaf::Function;
                                                   $1->series = NULL; // not tiz/best
                                                   $1->function = *($1->lvalue.n);
+                                                  delete $1->lvalue.n; // not used anymore
+                                                  $1->lvalue.l = NULL; // avoid double deletion
```
liversedge, line 462: Oh wow- that is unbelievably obscure -- suspect that took a while to find :)
Chapeau !

amtriathlon: The crash was obscure because Leaf::clear received a pointer to a QString as parameter and tried to use it as a pointer to Leaf, but the previous Leaf::clear call in the stack had type function, name filename and lvalue.l was not NULL, that was the hint to look at how function nodes were initialized.
Do you agree to merge this PR?

liversedge: I think we should do more testing will try and find some time this weekend since it will be raining here in the UK

liversedge:
Spent some time yesterday and today testing this and couldn't get it to crash. Its definitely ready to merge.
I wonder if there are any leaks left where datafilters are not being deleted- I looked but couldn't find any.

amtriathlon:
Thank you, Mark!

PR #4303, opened 2023-01-02
Added support for old Daum devices before 2001

LateNightCoder0815:
Unfortunately, my old Daum Electronic device "Ergobike Cardio" (comparable to the 4008 series and many other bikes) from before 2001 uses a slightly different RS232 protocol in some of the cases, which can be found here: http://www.ergo-lyps.de/de/download/v01/schnitt.zip . Additionally, it uses a boud rate of 4800.

As the old devices can not be detected by the checkversion (was not implemented at that time) I used a _profile of "OLD_DAUM" to identify the device in the configuration setting.

With these changes everything works as intended. Thank you very much for the project and hopefully I can contribute back in this way.


src/Train/Daum.cpp
```diff
@@ -369,11 +390,18 @@ void Daum::requestRealtimeData() {
 bool Daum::ResetDevice() {
     QByteArray dat;
     dat.append((char)0x12).append(deviceAddress_);
-    return WriteDataAndGetAnswer(dat, 3).length() == 2; // device tells pedalling state too
+    int answer_len = 2;
+    if (profile_ == "OLD_DAUM"){
+        answer_len = 3;   // Old cockpits send 3 bytes answer
+    };
+    return WriteDataAndGetAnswer(dat, answer_len).length() == answer_len; // device tells pedalling state too
```
amtriathlon, line 397: This is different from existing code for other versions, this should not happen for this to be merged since we cannot test with previously supported devices.

src/Train/Daum.cpp
```diff
@@ -404,7 +432,11 @@ bool Daum::SetProgram(unsigned int prog) {
     QByteArray dat;
     if (prog > 79) { prog = 79; }   // clamp to max
     dat.append((char)0x23).append(deviceAddress_).append((char)prog);
-    return WriteDataAndGetAnswer(dat, dat.length() + 1).length() == 4; // device tells pedalling state too
+    int answer_len = 4;
+    if (profile_ == "OLD_DAUM"){
+        answer_len = 3;   // Old cockpits send 3 bytes answer
+    };
+    return WriteDataAndGetAnswer(dat, answer_len).length() == answer_len; // device tells pedalling state too
```
amtriathlon, line 439: I am not sure if this is the sama as existing code for other versions, for this to be merged that should be very clear.

src/Train/Daum.cpp
```diff
@@ -369,11 +390,18 @@ void Daum::requestRealtimeData() {
 bool Daum::ResetDevice() {
     QByteArray dat;
     dat.append((char)0x12).append(deviceAddress_);
-    return WriteDataAndGetAnswer(dat, 3).length() == 2; // device tells pedalling state too
+    int answer_len = 2;
+    if (profile_ == "OLD_DAUM"){
+        answer_len = 3;   // Old cockpits send 3 bytes answer
+    };
+    return WriteDataAndGetAnswer(dat, 3).length() == answer_len; // device tells pedalling state too
```
LateNightCoder0815, line 397: Ensured that the changed code calls the function exactly identically as before for older devices

src/Train/Daum.cpp
```diff
@@ -404,7 +432,11 @@ bool Daum::SetProgram(unsigned int prog) {
     QByteArray dat;
     if (prog > 79) { prog = 79; }   // clamp to max
     dat.append((char)0x23).append(deviceAddress_).append((char)prog);
-    return WriteDataAndGetAnswer(dat, dat.length() + 1).length() == 4; // device tells pedalling state too
+    int answer_len = 4;
+    if (profile_ == "OLD_DAUM"){
+        answer_len = 3;   // Old cockpits send 3 bytes answer
+    };
+    return WriteDataAndGetAnswer(dat, dat.length() + 1).length() == answer_len; // device tells pedalling state too
```
LateNightCoder0815, line 439: Now length is calculated from dat length as before so that it becomes obvious that this results in the same byte length.

src/Train/Daum.cpp
LateNightCoder0815: Thank you very much for the feedback and thorough review. I fixed this occurence!

src/Train/Daum.cpp
LateNightCoder0815: You are right directly from the code this is hard to see. I reverted this first part of the change so that it becomes clear that it is identical. Thank you again!

amtriathlon:
Hi, thank you for the update, the code is ok for me, except for some cosmetics:
1) keep the use of spaces consistent: `}else{` or `){` in some places don't match the rest of the code
2) `};` to end a block is redundant, please remove the `;` 

After that it would be ready to merge.

LateNightCoder0815:
@amtriathlon Thank you. I updated the formatting accordingly to match the rest of the code.

amtriathlon:
Thank you for your contribution!

PR #4309, opened 2023-01-15
Fix discrepancies in appveyor Python path

vlcvboyer:
When builing Windows executable, appveyor.yml is using 2 different path for Python:
C:\Python which seems to be initial version of appveyor.yml
and C:\python37-x64 which seems to be a recent update
However C:\Python is still partially in appveyor.yml
This generate errors during build process on my side which were solved by changing remaining occurences of C:\Python by C:\python37-x64.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -59,7 +59,7 @@ cache:
 - jom_1_1_3.zip
 - sip-4.19.8.zip
 - C:\R
-- C:\Python -> src\Python\requirements.txt
+- C:\python37-x64 -> src\Python\requirements.txt
 - c:\tools\vcpkg\installed\
 - qwt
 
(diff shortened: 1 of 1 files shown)


amtriathlon:
I can’t see the point of this change, it is not hard to see in AppVeyor logs the current script builds correctly the installer and in fails with your changes

vlcvboyer:
Ok. Sure keep it working !
You should have another build environment than mine.
@amtriathlon : FYI (maybe this error will appear in few month on your environment...),
I've errors below with recent master branch when building on ci.appveyor.com :
image
master dated 4 years back have no issue:
image
Only recent master is creating trouble:
image
I supposed the issue was related to multiple Python version in different path and (in my appveyor instance) my patch solved the problem.
Local builds are working fine on my own computer.

amtriathlon:
Currently we are generating installers on AppVeyor, all information is public except for the cloud keys.

The alternate Python 3.7.9 installed in C:\Python is used to provide an embedded Python with included modules for user not able or willing to install and manage their own Python, and your changes break the installer generation, as you can see in https://ci.appveyor.com/project/Joern-R/goldencheetah-knhd8/builds/45941590, that's the reason for rejection.

Issue #4311, opened 2023-01-15
Detect Virtual activity on hometrainer


vlcvboyer:
Segregate indoor cycling activities by detecting
- virtual activities: when GPS trace follows workout fake trace
- hometrainer workout when GPS signal ramain whthin the same area range

Then populate sub_sport tag accordingly.

amtriathlon:
GPS data is always fake for Train generated activities, and they can be classified using filters.

PR #4314, opened 2023-01-15
Determine virtual activity sub-sport based on gps trace

vlcvboyer:
Segregate indoor cycling activities by detecting

- virtual activities: when GPS trace follows workout fake trace
- hometrainer workout when GPS signal ramain whthin the same area range

Then populate sub_sport tag accordingly.


vlcvboyer:
#4311

amtriathlon:
I am closing this since Train supports ANT+ foot pods it cannot be assumed Sport is Bike, and GPS data is always "fake" when present since Train cannot receive lat/lon from sensors.

Issue #4313, opened 2023-01-15
Import and export sub_sport tag from/to FIT File


vlcvboyer:
At present sub_sport tag in FIT files is marked as not supported.
Adding sub sport in FIT file decoder and encoder will allow to segregate indoor/outdoor workouts without any additional manual action after import.

Issue #4312, opened 2023-01-15
Populate more sports and sub sports as per latest FIT SDK


vlcvboyer:
Update FIT file decoder by including up to date sports ID and sub_sport ID as per FIT SDK.

PR #4315, opened 2023-01-15
Update sport list and sub sports

vlcvboyer:


vlcvboyer:
#4313

vlcvboyer:
#4312

src/FileIO/FitRideFile.cpp
```diff
@@ -3827,12 +3827,17 @@ void write_session(QByteArray *array, const RideFile *ride, QHash<QString,RideMe
 
     // 6. sport
     // Export as bike, run or swim, default sport is bike. 
-    // todo: support all sports based on tag "Sport"
-    int sport = ride->isRun() ? 1 : ride->isSwim() ? 5 : 2;
-    write_int8(array, sport);
+    value=FitFileReaderState::getSportId(ride->getTag("Sport",""));
```
amtriathlon, line 4160: Here would be better to use
https://github.com/GoldenCheetah/GoldenCheetah/blob/68c795440c022edf133bf803691af9b57905d6aa/src/FileIO/RideFile.cpp#L246
which already does the sport tag "normalization" due to translations, etc.

vlcvboyer: correct
PR updated accoringly

amtriathlon:
Thank you!

Issue #4024, opened 2021-08-23
Swim R-R feature request


pepelisbon:
Attached a Garmin Forerunner 945 + HRM-Pro (after 7.40 firmware update) FIT file, with Swim activity R-R intervals.

[7321736658.zip](https://github.com/GoldenCheetah/GoldenCheetah/files/7033310/7321736658.zip)

pepelisbon:
Decoded Swim R-R

[SwimRR-133725.xlsx](https://github.com/GoldenCheetah/GoldenCheetah/files/7039288/SwimRR-133725.xlsx)

vlcvboyer:
I'm working on this and planned to extract the R-R data using existing format used for HRMTri (as far as I remember...):
Does this seems ok for you ?
		"XDATA":[
		{
			"NAME" : "HRV",
			"VALUES" : [ "R-R", "R-R flag" ],
			"UNITS" : [ "msecs", "bool" ],
			"SAMPLES" : [
				{ "SECS":0.59, "KM":0, "VALUES":[ 590, 1 ] },
				{ "SECS":1.166, "KM":0, "VALUES":[ 576, 1 ] },
				{ "SECS":1.724, "KM":0, "VALUES":[ 558, 1 ] },
				{ "SECS":2.274, "KM":0, "VALUES":[ 550, 1 ] },
				{ "SECS":2.822, "KM":0, "VALUES":[ 548, 1 ] },
				{ "SECS":3.366, "KM":0, "VALUES":[ 544, 1 ] },
				{ "SECS":3.912, "KM":0, "VALUES":[ 546, 1 ] },

pepelisbon:
Seems good, to me, but i am not a programmer. The decoded RR posted above comes from fit file with another application.

RR swim data will be great for: 
 Evaluate the presence of artifacts in the recording through the tools that GC already has, 
 Compare it with the HR data record, and eventually replace it, by calculating HR form RR 
 Important also is to ensure ensure both  (main and RR) time-frames are the same, synchronized and with the same size

Many thanks, Pepe

amtriathlon:
FWIW here is were current HRV reading from FIT files was added: https://github.com/GoldenCheetah/GoldenCheetah/commit/0cff7c286830779079cc39941cd9e0f66ce944b5 and an ulterior fixup: https://github.com/GoldenCheetah/GoldenCheetah/commit/2c0eb835d4a95827c8dff6167d344f27331a75d9

[...]

PR #4319, opened 2023-01-20
collect R-R data from gramin HRM swim

vlcvboyer:
Read R-R from Garmin HRM-Swim which records each HR event into internal memory.
ANT message 132 / field 10


vlcvboyer:
#4024

amtriathlon:
Thank you!

PR #4320, opened 2023-01-20
highlight current selection in train mode

vlcvboyer:
At present when you select a workout GoldenCheetah automatically select the associated video and synchronization parameter
But user is not aware of this selection change.
This patch put green backgroud on selections in order to know which files will be considered for workout
image


vlcvboyer:
Note: I have not created a new entry in GoldenCheetah color palette as, I suppose, that this kind of update shall be done during version release in order to include it in GcUpgrade.cpp and thus upgrade user config files

amtriathlon:
Problem is selection is barely noticeable when the list loses the focus using dark colors, switching to light colors "fixes" the issue, but is has to be changed manually since even light themes has dark colors for Train.
It is not exclusive to Train, it also happens in Trends sidebar when using a dark theme, at least on Windows.
I think we need a more general, and more subtle (color wise), solution.

PR #4326, opened 2023-02-02
CP Chart - restate access to Delta Compare

### LINKED ISSUE #4256: Reveal controls block charts upper UX
This affects most old charts, but it is specially annoying for charts having functions requiring access to that zone, s.t. Performance and Metrics Trends y-axis scale knobs.

As discussed in the forum, alternatives are to get rid of reveal controls making older charts simpler and more consistent with new ones, or to make the hot spot smaller and centered to be less obtrusive.

In the first case we need to convert show/hide console setting in R&Python charts from reveal to standard controls, for the rest reveal controls are redundant and can be directly removed or hidden.

amtriathlon:
Reveal controls were disabled by 53081eb to avoid UX interferences, but Delta Compare options are not available as standard chart settings, since they are specific to Compare Mode we could keep them always visible at the top of the chart instead of moving them to standard settings.

This is how the top of the chart looks like in compare mode:
image

and in normal mode:
image

It uses some vertical space, but it is similar to User Charts in that respect, and it makes the UI simpler and more intuitive, not all users are aware of reveal controls, IMHO.

### DIFF
--- a/src/Charts/CriticalPowerWindow.cpp
+++ b/src/Charts/CriticalPowerWindow.cpp
@@ -73,22 +73,15 @@ CriticalPowerWindow::CriticalPowerWindow(Context *context, bool rangemode) :
     rDeltaPercent->setText(tr("as percentage"));
     rDeltaPercent->hide();
 
-    QVBoxLayout *checks = new QVBoxLayout;
-    checks->addStretch();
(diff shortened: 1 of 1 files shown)


amtriathlon:
Closed because I fell this needs more thought and discussions.

Issue #3840, opened 2021-03-11
ANT+ recording: Power phase metrics not recorded.


awolff64:
I have a new pair of Favero Assioma power meter pedals.
When recording power with my Garmin Fenix and then import the fit-File I see all metrics in particular the power phase metrics.
When recording power with GC during a training I see all metrics but the power phases.

Would be great if they would be recorded and be present for analysis.

Thanks,
Andreas

asendobry:
This would be fine for me as well. I have the same setup.

Thanks in advance,
Alexander

amtriathlon:
This issue is not specific to the Assioma but Cycling Dynamics was not implemented in Train, there is a Pull Request adding this feature available for testing, an installer for Windows can be downloaded from https://ci.appveyor.com/project/Joern-R/goldencheetah-knhd8/builds/50047523/artifacts, feedback is welcome.

PR #4327, opened 2023-02-05
Fit file: decode cyclist position (seated/standing)

vlcvboyer:
Decode rider position from FIT file and record into GoldenCheetah ride as "POSITION" XDATA.
IDs for positions: 0="seated", 2="standing", 1 to 3="transition" are from FIT SDK. IDs 4 to 9 are kept as spare in case of usage in future FIT developments. In addition, 10 to 11 are used for "aero" (determined by fixRiderPosition when cadence is zero for a 'long' period) and "off" (when speed is zero).
This will allow to include rider position during analysis by adjusting CdA accordingly.
Aerolab may be interessted by this also.
fixRiderPosition will generate the fake "aero" and "off" position and adjust effective position as per FIT SDK (rename "transition" to effective position when confirmed by sensor).


vlcvboyer:
This is the first step to go forward on #3840 (and avoid git conflicts during merge, etc.)
All Cycling Dynamics are already there (in FIT file import and GC rides) but rider position was missing.
I'm working on #3840 assuming that this PR is OK

amtriathlon:
Hi Vianney, I cherry picked and merged your commit https://github.com/GoldenCheetah/GoldenCheetah/pull/4327/commits/bf16b1f1682c6304e26ea6611a298946890f4a02 and added 2 fixed in https://github.com/GoldenCheetah/GoldenCheetah/commit/6a88d775b050f96988650eb38c849b228627bd78.
I think you fix tool would be interesting to remove transition positions, but I don't think the fake positions make too much sense here, I will merge it if you reinstate it without them.
Thank you!

marcenGC:
Also no POSITION Xdata series. I checked Garmin Connect and found that the Position data series is displayed by my own files.

PR #4331, opened 2023-02-05
Code clean out (removing few warnings from build log)

vlcvboyer:
few code cleanup:
* Change few local variable from int to size_t in order to silent warning related to possible loss of data
* Use some static cast for explicit conversions from size_t to int in order to silent warning related to possible loss of data
* Tag some provisional formal parameters with Q_UNUSED when not part of code in order to silent unreferenced formal parameter
* Use wchar instead of char when 2 bytes characters are used
* Adds a comment related to GC CSV file (one field was missing)

Only few warnings were adressed by this patch...

### DIFF
--- a/src/Charts/OverviewItems.cpp
+++ b/src/Charts/OverviewItems.cpp
@@ -1173,7 +1173,7 @@ DataOverviewItem::sort(int column, Qt::SortOrder order)
     // dates in German are weird, a month is "Mai", "Juli" or even "Jan."
     // even when requesting a date in format dd MMM yyyy
     // remember: a dot (.) inside brackets ([]) does NOT need to be escaped
-    QRegExp redate("^[0-9][0-9] [.A-Za-zÀ-ž\u0370-\u03FF\u0400-\u04FF]+ [0-9][0-9]*$");
+    QRegExp redate(QString::fromWCharArray(L"^[0-9][0-9] [.A-Za-zÀ-ž\u0370-\u03FF\u0400-\u04FF]+ [0-9][0-9]*$"));
     QRegExp retime("^[0-9:]*$");
     QRegExp renumber("^[0-9.-]*$");
 
@@ -5454,6 +5454,7 @@ VScrollBar::setAreaHeight(double n)
 void
 VScrollBar::setPos(double x)
 {
+    Q_UNUSED(x);
     // xxx todo
 }
 
--- a/src/FileIO/PowerTapDevice.cpp
+++ b/src/FileIO/PowerTapDevice.cpp
@@ -104,6 +104,7 @@ PowerTapDevice::doWrite(CommPortPtr dev, char c, bool hwecho, QString &err)
 int
 PowerTapDevice::readUntilNewline(CommPortPtr dev, char *buf, int len, QString &err)
 {
+    Q_UNUSED(len);
     int sofar = 0;
     while (!hasNewline(buf, sofar)) {
         assert(sofar < len);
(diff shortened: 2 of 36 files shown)


src/FileIO/PowerTapDevice.cpp
```diff
@@ -104,6 +104,7 @@ PowerTapDevice::doWrite(CommPortPtr dev, char c, bool hwecho, QString &err)
 int
 PowerTapDevice::readUntilNewline(CommPortPtr dev, char *buf, int len, QString &err)
 {
+    Q_UNUSED(len);
     int sofar = 0;
     while (!hasNewline(buf, sofar)) {
         assert(sofar < len);
```
paulj49457, line 110: If len is unused, but in NDEBUG mode it is checked in the assert, so should the assert be removed?

src/Charts/OverviewItems.cpp
```diff
@@ -1173,7 +1173,7 @@ DataOverviewItem::sort(int column, Qt::SortOrder order)
     // dates in German are weird, a month is "Mai", "Juli" or even "Jan."
     // even when requesting a date in format dd MMM yyyy
     // remember: a dot (.) inside brackets ([]) does NOT need to be escaped
-    QRegExp redate("^[0-9][0-9] [.A-Za-zÀ-ž\u0370-\u03FF\u0400-\u04FF]+ [0-9][0-9]*$");
+    QRegExp redate(QString::fromWCharArray(L"^[0-9][0-9] [.A-Za-zÀ-ž\u0370-\u03FF\u0400-\u04FF]+ [0-9][0-9]*$"));
```
paulj49457, line 1176: Can you explain why the "L" after the opening bracket is required, or what it does ?

paulj49457 (commented):
Please see comments

src/FileIO/PowerTapDevice.cpp
vlcvboyer: Q_UNUSED will not have side effect except hiding the "unused parameter" warning during build process:
- which may sometimes be useful when we expect a clean code and inspect build log on regular basis tracking incomplete code
- but sometimes filling up the build log with such warnings will be useless. Then placing Q_UNSUSED to disable this warning for this specific parameter is a good practice (when we prepare the code structure for future features or when due to object inheritance, etc.)

In our case I find it cleaner to use Q_UNUSED for this parameter in this function.

Please note that in any case, even when Q_UNUSED is used, the parameter will still be usable in the function.

src/Charts/OverviewItems.cpp
vlcvboyer: in the code we are using wchar_t: '\u0370', '\u03FF', '\u0400', '\u04F' which raise a warning when compiling on windows:
"warning C4566: character represented by universal-character-name '\u0370' cannot be represented in the current code page (1252)"
as we are using wchar_t string we have to use wchar_t* pointer which is defined by adding "L" in front of double quotes.

Details were included in the commit comments:
https://github.com/GoldenCheetah/GoldenCheetah/pull/4331/commits/86b84e31216b5e29306ad7e99f886cf9539e2a25
image

src/FileIO/PowerTapDevice.cpp
vlcvboyer: I don't know about this assert thus I will not propose to remove it without further investigation that I've not conducted.
I did not plan to spend time on such subject. On the other hand I took time to clean out partially the build log as I'm likely to have a look at it before sending a PR.

src/Charts/OverviewItems.cpp
paulj49457: Thank you for the explanation.

amtriathlon:
Thank you!

PR #4336, opened 2023-03-04
Enable Python DPs use Metadata in automatic mode 

amtriathlon:
Data Processors running on import are applied before the activity
is added to RideCache and metrics are computed, this behavior is
by design, likely to optimize resource usage on bulk import.
So activityMetrics API is not available and a new getTag API was
added for this uses. setTag/delTag/hasTag were changed to work
in this context too.
When the Python fix is executed on activities already in the cache
either via Edit menu, Filters or other Python Fix changes are
notified via the corresponding RideItem.


peret2000:
There is a compile error in Python/SIP/Bindinds.cpp, in [Bindings::getTag()](https://github.com/amtriathlon/GoldenCheetah/blame/PyFixAuto/src/Python/SIP/Bindings.cpp#L1728): it returns a QString, but in lines [#1731](https://github.com/amtriathlon/GoldenCheetah/blame/PyFixAuto/src/Python/SIP/Bindings.cpp#L1731) and [#1734](https://github.com/amtriathlon/GoldenCheetah/blame/PyFixAuto/src/Python/SIP/Bindings.cpp#L1734) it is returning bool values. I am modifying locally the behavior, returning "" in such cases, and testing again

peret2000:
I think it is working fine (I have not tested your last commits that solve the compilation error and the notification issue).
I have modified all my python scripts to use GC.getTag() instead of GC.activityMetrics(), and the behavior is correct when importin activities, after condiguring scripts to run on import.
In fact, it seems that GC.activityMetrics() does not fetch the correct value if the the tag has been modified and the activity has not been saved yet (unless your last commit solves it)

So, it seems the PR woeks perfectly, and we should replace GC.activityMetrics() by GC.getTag()
Thanks for this very useful functionality!

amtriathlon:
Thank you for the feedback, I will merge this for extended testing.

paulj49457:
When I run a python fix on a single activity it doesn't update the data in the "Data" tab or the list of activities:

i) When I select another activity and then return to the one I ran the python fix on, the "Data" tab has updated, but the activity list still displays the old data.
ii) Selecting "Save all modified activities" causes the activity list to update to match the new "Data" field value.

amtriathlon:
I can’t reproduce this, may be you have not included the last commit dealing with current activity notification. It is merged now so better test using master head or latest snapshot build.

paulj49457:
It all looks fine, I'd taken the PR before the final commit.

Thank you for re-instating this functionality, and fixing it at a more fundamental level than my previous attempts, I'll close my PRs relating to the rideImportWizard approach.

PR #4337, opened 2023-03-09
Cycling dynamics in train mode

vlcvboyer:
Configure power sensors in order to enable cycling dynamics data broadcast
Read the same during hometrainer session
Record them into csv and then into activity json
includes commits with few cosmetics in code
and evolution to CSV parser/writer in order to be more flexible to add fields in the future
and part of #4327 (one commit) which is required by this commits sequence


vlcvboyer:
Related to Feature request #3840

lennartalff:
Thanks for this PR @vlcvboyer :blush:
I tested this PR with my Garmin Vector3 pedals. From my first impression it seems to work as expected.
The power phase data appears in the rides basic data and xdata with the rider position, i.e. seated/standing, is created.
So I guess, from a user's perspective you get what you can expect from this PR.

I assume that enabling cycling dynamics has implications on battery life time of the recording device. I don't know how much and how relevant this is for the concerned users. If we consider it relevant, it might be useful to make cycling dynamics optional and enable/disable them via the UI.

@amtriathlon in #3840 you asked for feedback. Do you have any specific aspect in mind I could examine besides what I already did?

amtriathlon:
Thank you, I was looking for users testing this feature for functionality and possible regressions.
If you can attach some of your recorded tests (exported json), it would be great.

lennartalff:
Here you are, @amtriathlon .
The first file is from yesterday with just a few seconds recorded. The second one is from today's normal training session. I was cycling in standing position for a few seconds several times for the sake of getting some position xdata.
[2024_09_24_18_30_48.json](https://github.com/user-attachments/files/17136680/2024_09_24_18_30_48.json)
[2024_09_25_19_13_19.json](https://github.com/user-attachments/files/17136681/2024_09_25_19_13_19.json)

amtriathlon:
Thank you!

PR #4357, opened 2023-04-28
Allows optional cloud service sync on auto import

### LINKED ISSUE #2710: Feature Request: auto import raw *.fit files from dropbox folder
Hi, it would be nice if we could auto import *raw data files* from a dropbox cloud folder and not only from a local folder. I could mount one dropbox account with my raw *.fit files to my local drive and auto import it then into my athlete profile. As far as I know it is unfortunately not possible to mount a second dropbox account on my pc and therefore not possible to manage a second or more athletes in this way. By the automatic import from a dropbox cloud folder, this topic could be solved.

paulj49457:
This PR aims to resolve #2710 by adding an option in the auto import dialog to incude the configured cloud services.

Note: Due to the secret key cloud service key, I cannot test it locally so hopefully I can use the version built by the GC build process.

cloudSync

### DIFF
--- a/src/Core/Athlete.cpp
+++ b/src/Core/Athlete.cpp
@@ -366,6 +366,11 @@ Athlete::configChanged(qint32 state)
 void
 Athlete::importFilesWhenOpeningAthlete() {
 
+    if (autoImportConfig->importCloudSyncEnabled()) {
+        // sync with cloud
+        cloudAutoDownload->checkDownload();
+    }
+
     autoImport = NULL;
     // just do it if something is configured
     if (autoImportConfig->hasRules()) {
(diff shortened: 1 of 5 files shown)


liversedge:
When you configure a cloud service you can specify if you want to sync with it on import.
Screenshot from 2023-04-28 11-55-59

paulj49457:
> When you configure a cloud service you can specify if you want to sync with it on import. Screenshot from 2023-04-28 11-55-59

Sorry wasn't aware of this functionality, this solves the issue #2710 that was raised,

paulj49457:
Sorry wasn't aware of this functionality, this solves the issue #2710 that was raised,

PR #4365, opened 2023-06-03
Use configured bicycle weight for training in Simulated Speed

peret2000:
When computing simulated speed based on slope, altitude, etc, total weight (athlete + bicycle) is relevant, specially in up and down hills

Currently, It is using 10kg plus the athlete weight. Now, it considers configured bicycle weigth. Not very relevant, but consistent with the data GH has for training

### DIFF
--- a/src/Train/TrainSidebar.cpp
+++ b/src/Train/TrainSidebar.cpp
@@ -1990,7 +1990,7 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
             displayLongitude = rtData.getLongitude();
             displayAltitude = rtData.getAltitude();
 
-            double weightKG = context->athlete->getWeight(QDate::currentDate()) + 10; // 10kg bike
+            double weightKG = bicycle.MassKG();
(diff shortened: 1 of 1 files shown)


src/Train/TrainSidebar.cpp
```diff
@@ -1990,7 +1990,7 @@ void TrainSidebar::guiUpdate()           // refreshes the telemetry
             displayLongitude = rtData.getLongitude();
             displayAltitude = rtData.getAltitude();
 
-            double weightKG = context->athlete->getWeight(QDate::currentDate()) + 10; // 10kg bike
+            double weightKG = bicycle.MassKG();
```
amtriathlon, line 1993: This looks like a mistake, athlete weight is the main component of total weight and you are removing it.

peret2000: That is the first thoght I had; but [Bicycle::MassKG()](https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/Train/BicycleSim.cpp#LL280C4-L280C4) is considering also the athlete weight.
I could check it in debug.

PR #4372, opened 2023-06-15
Update FITmetadata.json

dconka:
Add new SIGMA ROX 12.1 EVO to the list.


amtriathlon:
This change will be discarded no the next FIT SDK update, see https://github.com/GoldenCheetah/GoldenCheetah/tree/master/util/fit

Issue #786, opened 2014-02-04
See multiple rides on one map


mikeheiss:
It would be nice to be able to see multiple (if not all) rides on a single map.  That would make route planning easier to use roads you know are already good for riding.

amtriathlon:
We don't have multiple selection in the activities list, but to enable compare mode for the Map chart would allow to do this.

amtriathlon:
Fixed by #4389

PR #4389, opened 2023-08-28
Implemented compare mode for RideMapWindow

thejockl:
* Related issue #786
* Multiple activities can be viewed on one map at the same time
* Implemented for OSM and Google


amtriathlon:
Hi Joachim, it looks very nice and it works as expected in my preliminary tests, I posted to users forum for feedback and requested @grauser review. Thank you!

grauser (approved):
Thank you for this new feature.

Issue #4394, opened 2023-09-08
UX enhancement when drawing intervals on the map


thejockl:
Drawing intervals on the map currently has two issues that reduce user experience:

1. Clicking on the track fails in most cases to find the closest point
2. Drawing does not work on portions of the track already marked as interval. As a result, intervals can only be enhanced, not shrinked when drawing

PR #4395, opened 2023-09-08
UX enhancement when drawing intervals on the map

thejockl:
Related issue: #4394


thejockl:
Will cleanup and recreate

PR #4399, opened 2023-09-22
Batch Processing addition of metric & metadata operations

paulj49457:
The metric & metadata fields can be set from the search/filter windows, but this PR adds metric & metadata operations to the BatchProcessingDialog with a simple UI: 

Batch processing

The BatchProcessingDialog now supports:

Metadata field update
Set Metric Override
Clear Metric Override

The Metadata fields and Metric fields have been separated, as the latter override the existing GC calculated metrics, whilst the metadata fields are overwritten, so two different behaviours.


amtriathlon:
Hi Paul, I think the Metadata option can be handy for users not too comfortable with formulas, I can think of examples s.t. setting Equipment for a set of activities, etc. Perhaps I would have looked to FormField for the UI, but the text editor with a hint is simpler and likely good enough.
OTOH I cannot think a single example where overriding a metric to the same fixed value can be useful, I would prefer to remove the Metrics options before merge.
Thank you for your contribution.

paulj49457:
Hi Ale, Thank you for the feedback, I'll remove the Metric option from this PR.

paulj49457:
@amtriathlon before I finalise the changes can you confirm the following is correct? I've been testing the metadata batch processing and noticed that some metadata names are the same as metric names, and are currently removed from metadata options list as follows, leaving only the metadata options at the end of this message. 


The following metadata fields have been removed from the metadata options list as they are very destructive for batch processing:
metadata: Start Time
metadata: Start Date
metadata: Data

I also notice that the following names are in both Metric & Metadata names, and have been removed from the metadata options list:

metric - Metadata: Duration
metric - Metadata: Time Moving
metric - Metadata: Distance
metric - Metadata: Work
metric - Metadata: Elevation Gain
metric - Metadata: Average Speed
metric - Metadata: Average Power
metric - Metadata: Average Heart Rate
metric - Metadata: Average Cadence
metric - Metadata: Aerobic TISS
metric - Metadata: Anaerobic TISS
metric - Metadata: TSS
metric - Metadata: GOVSS
metric - Metadata: SwimScore
metric - Metadata: xPower
metric - Metadata: Daniels Points
metric - Metadata: Daniels EqP

Also the following names are similar in both Metric & Metadata names and have been removed from the metadata options list:
metadata: BikeScoreÔäó           --   metric:BikeScore&#8482;
metadata: CP                            --   metric: Critical Power 
metadata: Recording Interval  ??

The Metadata fields that can still be changed by batch processing:
Metadata : Sport
Metadata : Notes
Metadata : Device
Metadata : Device Info
Metadata : Calendar Text
Metadata : Interval Notes
Metadata : Interval Goal
Metadata : < Plus any user defined Metadata Items, providing they do not match any metric names >

paulj49457:
@amtriathlon Thank you the explanation, I have applied your guidance in the latest commit which is ready for review/merge

amtriathlon:
Hi Paul, we have a class to go back and forth between external and internal names of fields and to check if they correspond to metrics: https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/Metrics/SpecialFields.h

paulj49457:
@amtriathlon I wasn't aware of it, but I'll update the PR to use the class save any duplication

amtriathlon:
> @amtriathlon I wasn't aware of it, so I have updated the PR to use the SpecialFields class in the latest commit.

You need to use it for translations too, I mean, the combobox should show displayName instead of internalName since they can be different when using other language than english, but internalName for the update.
I pushed a fix for SpecialFields so you can use isUser/isSpecial to filter fields which cannot be updated: https://github.com/GoldenCheetah/GoldenCheetah/commit/964797564305ef1c4c95623d4610c8ee6306b025

paulj49457:
@amtriathlon I have updated the PR to use the recent change to isUser function within SpecialFields class.

amtriathlon:
Thank you!

PR #4398, opened 2023-09-20
Feature request - Favorite workouts

thejockl:
This PR contains a possible fix for #411
It introduces a new schema for the TrainDB, filters for workouts and an info-widget in the TrainSideBar. Workouts can be rated and tagged.
The filter works on selected metrics of the workout as well as the description and all assigned tags.
As soon as the zones / CP of an athlete change, the TrainDB is automatically updated.
A manager to add / remove tags was added to the Train settings.

Additionally the change includes refactorings to improve type safety (converted some enums to C++11 enum class) and Extracted pure data baseclasses from ErgFile and VideoSyncFile to achieve faster turnaround times by developing outside GoldenCheetah.

Example for a filter: "duration 1:00..1:15, variability 1.1..1.2, dominantzone z4, stress 80.., rating 3..4"

Please excuse the size of this PR

Issue #411, opened 2012-12-30
Feature request - Favorite workouts


jjofre:
Allow capability to manage organize and create favorite workouts in the library

PR #4400, opened 2023-09-23
Train View: Improved the selection of workouts

thejockl:
This PR tries to provide a solution for #411

It introduces a new schema for the TrainDB, filters for workouts and an info-widget in the TrainSideBar. Workouts can be rated and tagged.
The filter works on selected metrics of the workout as well as the description and all assigned tags.
As soon as the zones / CP of an athlete change, the TrainDB is automatically updated.
A manager to add / remove tags was added to the Train settings.

Additionally the change includes refactorings to improve type safety (converted some enums to C++11 enum class) and Extracted pure data baseclasses from ErgFile and VideoSyncFile to achieve faster turnaround times by developing outside GoldenCheetah.

Example for a filter: "duration 1:00..1:15, variability 1.1..1.2, dominantzone z4, stress 80.., rating 3..4"

This PR is the replacement for #4398 after cleaning up the commit-history and realizing Marcs first comments:
* The contents of the folders Train/info and Train filter were moved into Train
* Removed a proprietary definition working only in GNU Bison
* Removed some debug-statements in TrainDB.cpp
* Changed the license of tags (100% authored by me) from BSD to GPLv2 and moved the code from contrib/tags to Train

Additional remarks:
* The new code in contrib/qtsolutions/flowlayout was NOT authored by me but taken from https://code.qt.io/cgit/qt/qtbase.git/tree/examples/widgets/layouts/flowlayout?h=5.15 (see also https://doc.qt.io/qt-5/qtwidgets-layouts-flowlayout-example.html) I updated this code slightly to vertically align the widgets in this layout.
* Please be aware that the code replaces the TrainDB-schema automatically


liversedge:
When its ready for commit let me know and will review with a view to pull

thejockl:
All issues I gathered from the users-list are resolved, so I would be happy if you could start the review

liversedge:
Will review this week - thank you for your efforts !

peret2000:
Gret job! I am enjoying this feature. After a few weeks, I noticed that there are situations where personal data of workouts are lost (rating, stars, tags and last training date).
When scanning for workouts, videos and videsyncs, even the workouts files remain in its location, database is rebuilt and workouts are loaded againg. This is not an issue with database version 1, where all data came from workout files.
With this proposed PR, workout and workout_tag tables also hold data entered by the user. So, whenever a scan is done (and it is a common use case, I guess) to add new workouts, without caring which are the new ones, just rescanning, those personal data are lost.
So, if you scan frequently (or even rarely), you are loosing these rich data.

I have done a small modification (in LibrarySearchDialog::updateDB()) that does not loose those data. It just not rebuild all the tables (it excludes workout, tagstore and workout_tag); then, it inserts all the workouts found (the existing ones are untouched), and finally it deletes the workouts in the database that have not been found during the scan.

I am attaching a patch file, in case you consider it is a good to include it in your PR. If you prefer, I could make a PR to your forked project; let me know.

Thanks.
[patch.zip](https://github.com/GoldenCheetah/GoldenCheetah/files/13992118/patch.zip)

thejockl:
Thanks for the patch, @peret2000! It looks good and should be integrated. To keep things simple, I would prefer adding it it to this PR. If you are fine with it, I will add it directly

peret2000:
> Thanks for the patch, @peret2000! It looks good and should be integrated. To keep things simple, I would prefer adding it it to this PR. If you are fine with it, I will add it directly

Sure, for me it is ok

amtriathlon:
Hi Joachim, I fixed a conflict due to a merged PR from another contributor and added a little fix for Qt6 compatibility, current conflict is with your own PR, so you will have a better idea about how to fix it.
I started to look at the code and didn't understand the need to introduce the ErgFileBase class, considering ErgFile base is still a monolithic clase handling all different modes and file formats, and not a set of subclases, am I missing something?

thejockl:
Hi Ale, happy to see you looking into this PR!

I just merged master into this branch, resolving all conflicts.

The reason for the *Base-classes (I honestly dislike the name but couldn't come up with something better) is that this helped me developing most of the code in a small testdriver outside Golden Cheetah, giving the following benefits:
* Reduced compile times
* Reduced startup times
* No navigation required to reach the functionality under development
* Showing the full models in a table, a great help while developing the display-widgets and filtering (why don't I find this workout?)
By that I had way faster turnaround times, and could better concentrate on the features.

thejockl:
As I experienced a crash related to auto completion when running with Qt6, please wait for a fix before further proceeding

thejockl:
The crash has been fixed

amtriathlon:
This is a very nice feature, Joaquim, thank you very much!

We will need to update the wiki for v3.7 in two places: https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Side-Bar_Train-view and https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Special-Topics_Workouts%2C-Media-and-VideoSync-Library

thejockl:
Hi @amtriathlon, thanks so much for taking care and merging, I highly appreciate all your efforts!

I will look into the Wiki and come up with proposals, probably on the dev-list

PR #4403, opened 2023-10-06
Better user experience when drawing segments on the map

thejockl:
* Increased the area in which points are searched when clicking / hovering a track, easing the process of starting segment creation and giving more precision while drawing
* Hovering the track when drawing segments is not blocked any more by highlighted segments / parts of the current segment (i.e. segments can be shrinked during creation)

See #4394
This is a cleaned up version of PR #4395


amtriathlon:
Thank you!

PR #4408, opened 2023-10-22
Mapview: Added interactivity to the Smallplot

thejockl:
* Added a tooltip showing data about the current position (Altitude, Power, HR, Time)
* Added a marker to the map corresponding to the current mouse position (both Google and OSM)
* Changed scaling of the smallplot: Chart is now between lowest point and highest point instead of sea level and highest point

Why did I change the scaling: I live about 650 m above sea level with climbs having max. 150 vertical meters. In the current implementation this leads to pretty flat profile with a lot of unused space below. The updated implementation gives a much more realistic view.


amtriathlon:
I works and it seems very useful, I will merge it, thank you!

Issue #2945, opened 2018-10-15
Cannot create workouts on MacOS V3.4


acormack93:
When creating a workout via the wizard or editor, the message: "workout name already exists in directory:0" then it prompts the removal of the file. End result is that it cannot be created. Also tested in the latest MacOS development build

amtriathlon:
This happens when your user doesn’t have write permission on the configured workouts directory, check your settings in Preferences.

acormack93:
Ergh PEBKAC, thanks closing

PR #4410, opened 2023-10-30
Fixed broken workout directory check for the first start

enngq11k:
Hi,

I'm new to Golden Cheetah and had a rough start, because I wasn't able to save workouts without errors.
warn

error

It wasn't really obvious to me that you have to change the athlete folder **after** creating an athlete for the first time and then create a new athlete for this folder. I found some bug reports (like #2945) with the described fix. After I checked the code, this was and still is a real bug.

This pull request tries to deliver a better first use experience. This required two changes:

Fix1:
appsettings->value default fallback is "0" if no fallback value was given. So a fresh GoldenCheetah installation will always use path "0" until a workout directory was set in the settings.
We now use an empty string as fallback to get the condition working

Fix2:
Instead of using the athlete->home parent folder, the old code returned the parent's parent folder. Until a workout directory was set in the settings, the user had to navigate to the (hidden) athletes folder every time by hand.

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -178,8 +178,9 @@ MainWindow::MainWindow(const QDir &home)
 
     // if no workout directory is configured, default to the
     // top level GoldenCheetah directory
-    if (appsettings->value(NULL, GC_WORKOUTDIR).toString() == "")
-        appsettings->setValue(GC_WORKOUTDIR, QFileInfo(context->athlete->home->root().canonicalPath() + "/../").canonicalPath());
+    if (appsettings->value(NULL, GC_WORKOUTDIR, "").toString() == ""){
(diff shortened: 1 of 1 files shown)


amtriathlon:
Hi, Fix 1 is a good catch and I am willing to apply it, thank you!
Fix 2 not so much, `context->athlete->home->root()` is the athlete's folder and its parent the athletes directory, which is a more reasonable default since workouts library is global for all athletes.

enngq11k:
Hi,

that's the reason for Fix2. Maybe this screenshot can help you:

workoutdir
As you can see, `workoutdir_bug` points to the parent of the default athletes directory.

I'm still confused that the second .canonicalPath() returns the parent... Maybe we can find a more understandable solution.

amtriathlon:
That’s right, [QFileInfo::canonicalPath](https://doc.qt.io/qt-5/qfileinfo.html#canonicalPath) excludes the file name pointing to the containing directory, which is already the parent for the directory case.

PR #4417, opened 2023-11-12
Added a configurable grayscale filter to RideMapWindow

thejockl:
* Added a CSS-grayscale filter to OSM based maps in RideMapWindow
* Added a configuration-option for the intensity level (0..10)
* Configuration is persisted per Tile-Server
* Fixed the visibility of Google/OSM specific settings (previously the settings were only correct after selection of the other Maptype)

Background of this change: OSM-Maps (especially OpenTopoMap) can be very colorful, resulting in hardly visible activities. By adding the CSS grayscale-filter to the tiles, the contrast between the track and the map can be greatly improved. The attached image shows an example for levels 0 and 9

grayscalefilter


amtriathlon:
Thank you!

Issue #4318, opened 2023-01-17
BTLE Sensors explicit pairing


amtriathlon:
Explicit pairing, like we have for ANT+, is a long standing feature request from users which comes recurrently, so let's use this issue as a placeholder. Current functionality have some limitations and it is explained in https://github.com/GoldenCheetah/GoldenCheetah/wiki/FAQ-TRAIN#how-to-pair-btle-sensors

Issue #4093, opened 2021-12-02
Bluetooth not working on macOS v 12.0.1


erikboto:
I'm not entirely sure, but I think this is related to macOS 12 (Monterey). If someone could confirm that e.g. the newest dev-release work with BLE devices on older macOS that would be nice.

The scanning for BLE devices do not work. If I start the app with --debug I see the following lines repeating when trying to connect to a bluetooth device in Train View:
`qt.bluetooth.osx: IOKit error code:  1
Error while scanning BT devices: QBluetoothDeviceDiscoveryAgent::UnknownError`

Possibly related to this issue in the Qt Issue Tracker: https://bugreports.qt.io/browse/QTBUG-96557

erikboto:
I tried to add the `NSBluetoothAlwaysUsageDescription` key to the .plist file, and gave `GoldenCheetah` permissions to access bluetooth in the macOS security and privacy settings, but still no luck. A bit annoying since adding that to the .plist for Qt's HeartRate Game-example fixed that app, so I really thought that would do the trick even for GoldenCheetah. One difference was that I built GC using Travis-CI and the HeartRate Game was built locally.

Apparently there's also a difference now in macOS if you run the binary inside the bundle, or if you launch the bundle itself. Launching the binary via the command line, like when passing --debug to enable debug output, will apparently now use the .plist file and therefore the app won't get access to things granted in that file.

amtriathlon:
> I tried to add the `NSBluetoothAlwaysUsageDescription` key to the .plist file, and gave `GoldenCheetah` permissions to access bluetooth in the macOS security and privacy settings, but still no luck. A bit annoying since adding that to the .plist for Qt's HeartRate Game-example fixed that app, so I really thought that would do the trick even for GoldenCheetah. One difference was that I built GC using Travis-CI and the HeartRate Game was built locally.

Travis-CI macOS builds get the plist from https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/Resources/mac/Info.plist.app

> Apparently there's also a difference now in macOS if you run the binary inside the bundle, or if you launch the bundle itself. Launching the binary via the command line, like when passing --debug to enable debug output, will apparently now use the .plist file and therefore the app won't get access to things granted in that file.

Does this difference also happens if you use the macOS open command to pass parameters like in https://github.com/GoldenCheetah/GoldenCheetah/wiki/Disable-R-Or-Python?

erikboto:
[...]

PR #4422, opened 2023-11-29
Added Bluetooth pairing to the device wizard

enngq11k:
Proposal for a simple Bluetooth paring with the device wizard. #4318 

The wizard will now search (but not connect) for available Bluetooth devices. The result is a simple list with all device names. The user can now choose the desired devices in a multi select list and save them as a "allow list" in the global training mode configuration. If no device is selected, the profile will use the old behaviour and the connect button searches for available devices.

There's room for improvement. Like, only scan for devices with specific capabilities, assign devices to specific roles (cadence / power / hr) and many more. This is just the first step and I'm always open for your feedback.

**Tested only on Linux.** I haven't had time to configure a macOS and Windows build yet.
ToDo: Update translations

**How does it look right now?**
After we found a Bluetooth device, the dialog will scan for sensors
search

You can now multi select the sensors (Normal sensors will report a name)
select

Name, address and uuid will be saved in the profile string
Done

### DIFF
--- a/src/Train/AddDeviceWizard.cpp
+++ b/src/Train/AddDeviceWizard.cpp
@@ -1025,14 +1025,20 @@ AddPairBTLE::scanFinished(bool foundDevices)
         {
             QTreeWidgetItem *add = new QTreeWidgetItem(channelWidget->invisibleRootItem());
 
+            // Remove chars used as separator in storage
+            QString deviceName = deviceInfo.name();
+            deviceName.replace(';', ' ');
+            deviceName.replace(',', ' ');
+            deviceName = deviceName.trimmed();
+
             // Save device info in item
-            add->setData(0, NameRole, deviceInfo.name());
-            add->setData(0, AddressRole, deviceInfo.address().toString()); // macOS
-            add->setData(0, UuidRole, deviceInfo.deviceUuid().toString()); // other OS
+            add->setData(0, NameRole, deviceName);
+            add->setData(0, AddressRole, deviceInfo.address().toString()); // other OS
+            add->setData(0, UuidRole, deviceInfo.deviceUuid().toString()); // macOS
 
             // Setup display text
             QLabel *status = new QLabel(this);
-            status->setText(deviceInfo.name());
+            status->setText(deviceName);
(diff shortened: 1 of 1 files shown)


liversedge:
This looks really promising!
When the sensors are found, is there any info that can be discovered about the device name or  capability-- working with those hex addresses feels a little obscure compared to what you'd typically see in any other bluetooth pairing process?

I'm pretty ignorant wrt bluetooth, so forgive me if this is harder than it sounds

enngq11k:
I don't really know what these "Hex"-devices are, but the QBluetoothDeviceDiscoveryAgent always reports them. I will investigate (maybe just a Linux issue).

If I'm using a real device it will show a proper name
h10-2

I will check if I can show some additional information about the device.

amtriathlon:
> **Tested only on Linux.** I haven't had time to configure a macOS and Windows build yet.

Hi, you can download a Windows installer from the AppVeyor ci builds, for your first commit it is available from https://ci.appveyor.com/project/Joern-R/goldencheetah-knhd8/builds/48640895/artifacts, others will appear on completion.

enngq11k:
Thanks. Windows also works, but QT decided to put the progress label next to the progress bar. This doesn't look good in my opinion, but the ANT+ dialogs have the same problem.

I added validation for the `deviceProfile` string. An empty profile was working fine, but selecting `No sensors found...` did create an invalid `DeviceInfo`

I also wrote a comment about the address and uuid strings. You cannot exchange profiles between Linux/Windows and macOS, because you **either** receive an `address` or `uuid` (if the comment in 'BT40Controller::159` is still correct).

search
devices
NothingFound

amtriathlon:
Great!
Once this is complete and tested on Linux and Windows I propose to merge and generate a snapshot build to enable user testing. WRT macOS we have this https://github.com/GoldenCheetah/GoldenCheetah/issues/4093 pending for modern versions, so this cannot make it worse.

amtriathlon:
This is the users forum thread about this feature: https://groups.google.com/g/golden-cheetah-users/c/5wkUZSEyWx0

PR #4423, opened 2023-12-01
Update FIT SDK definitions to 21.126

MeisterP:
FIT 21.126 SDK
Release Date: November 21, 2023

see https://developer.garmin.com/fit/download/


amtriathlon:
Thank you!

PR #4425, opened 2023-12-04
Dialog to show Bluetooth connection status

enngq11k:
_This is a follow up to Bluetooth Pairing._

Connect now shows a dialog with the current connection status, if only one Bluetooth profile with allow list is selected.
I created this dialog because it is difficult to see which devices are connected.

This dialog offers to:
- accept the current connections and start a workout (available after scan)
- retry to establish connections to configured devices (available after scan)
- cancel connect and disconnect all devices (always available)

To get a fast overview, I decided to use colors for the connection status
- Green =  connected
- Yellow = not found
- Red = disconnected again
- Black = searching or connecting

We could achieve a generic dialog for Bluetooth/ANT+ and so on, but this would require a bigger refactoring and I only can test Bluetooth devices right now.

I also only support single device selection (the default setting)

These are real devices, with real connection states. I just changed the name for the screenshot.
connection_dialog

### DIFF
--- a/src/Train/AddDeviceWizard.cpp
+++ b/src/Train/AddDeviceWizard.cpp
@@ -1037,9 +1037,9 @@ AddPairBTLE::scanFinished(bool foundDevices)
             add->setData(0, UuidRole, deviceInfo.deviceUuid().toString()); // macOS
 
             // Setup display text
-            QLabel *status = new QLabel(this);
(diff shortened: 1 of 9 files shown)


amtriathlon:
Hi, thanks for your contribution but I don't like the idea of device controllers having its own UI.

PR #3956, opened 2021-07-23
Move to qwt 6 2 multiaxes

andreasbuhr:
Upgrading qwt 6.2 seems to be the hardest part.

The released qwt 6.2 seems to be lacking a so called "multiaxes" feature which is required for GoldenCheetah.
Here I placed the current "qwt-multiaxes" of qwt into the qwt folder and then fixed all compiler errors.

Open issues:

1. There are linker errors. The methods the linker is asking for do exist, so somehow there is a problem in the build system.
2. QwtSpline is now abstract. It interface changed considerably. Currently, usage of QwtSpline is commented out. This has to be fixed.
3. The enableAxis method is gone. Qwt source code suggests it is now synonymous to setAxisVisible. This needs testing.

PR #4427, opened 2023-12-26
Upgraded Qwt to 6.2 (branch: qwt-multiaxes)

thejockl:
This PR is based on https://github.com/GoldenCheetah/GoldenCheetah/pull/3956 with the following additions / changes:
* Upgraded to the latest version of the multiaxes-branch, thus elminating crashes of GoldenCheetah on startup
* Disabled the emitting of Layout Requests on geometry changes of QwtScaleWidget - without this, CPU utilization was up to 100% on one core
* Added the class SplineLookup, reusing small portions of code from Qwt 6.1
* Re-added the splines in WPrime and RideFile (resampling), using the new interface of QwtSpline


liversedge:
Hi- a few questions on this one:
* What is the crash on startup you referred to ?
* What spline in Wprime and RideFile (resampling should be done using libsamplerate)

liversedge:
OK I just looked at the code and can see a spline was used- what a hack- really should fix that at some point, but not in this fix up.

Sorry about that.

thejockl:
The crash I am referring only occured on startup of GC after applying #3956 on Linux. This was resolved by pulling in the latest version of the multiaxes branch of qwt 6.2
I fully agree with the ugliness on the use of QwtSpline, especially in RideFile. However this PR resembles the status quo of 3.6 and does not try modifying it. If you would prefer taking the step removing the splines here first, please let me know, happy to help.

With this PR I got a stable version of GC on Linux, utilizing qwt 6.2 which I consider an important step away from Qt5 (EOS since May 2023) towards Qt6.

Please let me know if you see any hope for getting it merged. If not, I will quit the ugly job of setting up a Windows build environment to investigate the linker errors in Appveyor / Windows.

liversedge:
I will fixup the spline sometime in the future, its horrible and I wrote it !
On the windows errors- I will take a look now

liversedge:
The errors suggest that either QWT_DLL is not defined in qwt.pro and friends, or the way it works has changed in QWT 6.2 ?

thejockl:
I can't reproduce the link errors locally on Windows 10. Can I somehow obtain the configurations (qwtconfig.pri, gcconfig.pri, ...) from Appveyor?

amtriathlon:
Config is in:
https://github.com/GoldenCheetah/GoldenCheetah/blob/31636b196d5cb7bab0595f420697511672872a63/appveyor.yml#L68-L77 
I am on travel now, after 8/1 I can collaborate

thejockl:
Linking was successful, however it seems like the build timed out while uploading the artifacts

amtriathlon:
> Linking was successful, however it seems like the build timed out while uploading the artifacts

Sometimes just restarting the build finish, we are too close to the hard 1hr limit, once qwt is stable we need to reenable qwt caching to avoid rebuilding it on each build and it will be fine (?)

appveyor.yml
amtriathlon, line 1: It would be better to replace the  qwt line by qwt -> qwt/qwtconfig.pri.in so the cache gets refreshed on version update instead of removed.

thejockl: Done

amtriathlon:
I merged for more testing and will update snapshots shortly, thank you very much!

thejockl:
Thanks for the quick merge! I will continue testing and come back if I find any glitches

peret2000:
Sorry my lack of awareness; maybe this is a silly question. Building GC with these changes gives many errors (all of them when linking). I guess I need to install a newer version of Qt? According to travis/linux/before_install.sh, it seems that the automated builds continue to use 5.15, doesn't it?
So, must we adapt that script? I use it as the base of the builds.
I cannot come up with the software that we must update or install in the development environment to build with Qt6.2. Also, in QtCreator I only see, as expected, projects with 5.15.
Thanks for this upgrade!

amtriathlon:
Qt5.15 is ok, but you need to update your qwt/qwtconfig.pri file, we use a copy of qwt/qwtconfig.pri.in in ci builds.

PR #4428, opened 2024-01-07
Compass widget for Video training and Bearing telemetry field

peret2000:
I have developed the integration of a Compass to be used in the Video frame when training with a real time video.
I have always been curious about the bearing, when driving with the GPS, for example, and I found the same functionality nice for GoldenCheetah.

It uses the direction between the current position and the position within 2 seconds (with the current speed); it is a trade off between smoothness and accuracy of the bearing.

I am a complete ignorant about Qt, so maybe the code is not the best; I have copied code from other widgets and research a bit; in the end the code is rellay simple, but probably lacks some calls to QtPainter functions or some stuff.
If you consider that adding widgets is good for the app, I am happy to have the compass in the builds of GC.

[WorkoutVideo.webm](https://github.com/GoldenCheetah/GoldenCheetah/assets/19463568/f1e5acf6-de12-4745-9104-10a3271c04ed)

The configuration is simple; in _video-layout.xml_:
```
   <meter name="Compass" container="Video" source="Bearing" type="Compass">
       <RelativeSize Width="10.0" Height="10.0" />
       <RelativePosition X="60.0" Y="10.0" />
   </meter>
```
Also, a Telemetry field has been added: "Bearing". It can be seen as a numeric figure during the training.

It has been developed and tested with Qt5.15. the update to Qt6.2 is not ready in my environment, although there are no conflicts with the changes for 6.2

### DIFF
--- a/src/Train/DialWindow.cpp
+++ b/src/Train/DialWindow.cpp
@@ -522,6 +522,9 @@ DialWindow::telemetryUpdate(const RealtimeData &rtData)
     case RealtimeData::Altitude:
         valueLabel->setText(QString("%1").arg(value, 0, 'f', 1));
         break;
+    case RealtimeData::Bearing:
+        valueLabel->setText(QString("%1°").arg(value, 0, 'f', 0));
+        break;
 
     case RealtimeData::VO2:
     case RealtimeData::VCO2:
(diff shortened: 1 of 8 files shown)


amtriathlon:
Hi, thank you for your collaboration, but I will not merge this one. I think it is too specific to your needs, we already have the live map to see where the virtual ride happens.

PR #4431, opened 2024-01-12
Qt6: Removed an unused timer

thejockl:
Removed a timer object of type QTime that only utilizes methods removed in Qt6, but is never read


amtriathlon:
Thank you!

PR #4435, opened 2024-01-14
Qt6: Prevent the use of a deleted QVariant constructor

thejockl:
In Qt6.4++, the templated constructor QVariant(T) is marked as `= delete`, making compilation fail. By casting the return value to (const char*), the right QVariant constructor `QVariant const char *)` is explicitly selected instead of
`template <typename T,
              std::enable_if_t<std::disjunction_v<std::is_pointer<T>, std::is_member_pointer<T>>, bool> = false>
    QVariant(T) = delete;
`


amtriathlon:
Thank you!

PR #4436, opened 2024-01-14
Qt6: Fixed connections

thejockl:
In UserChart, the (non-existant) signal QComboBox::currentIndexChanged(QString) was connected to a slot without parameters. In Qt6 this created a warning. Changed to signal QComboBox::currentIndexChanged(int)


amtriathlon:
Thank you! There is another one in:
https://github.com/GoldenCheetah/GoldenCheetah/blob/bf28d7398ce224051a0538407f39d913b9eba15d/src/Gui/DownloadRideDialog.cpp#L53
the parameter is not used either, so is easy to fix too.

thejockl:
WIll add the DownloadRideDialog-Fix to this PR

thejockl:
Updated the PR

PR #4437, opened 2024-01-14
Qt6: Fixed a crash in RideMapWindow

thejockl:
Fixed a crash when dereferencing const_end()-iterator while searching latest selected interval


amtriathlon:
Thank you!

PR #4439, opened 2024-01-25
Added new season type "Year to date"

thejockl:
This PR adds a new semi dynamic season type "Year to date". This helps comparing progress for same timeranges of different years.
It re-uses the start of class Season and grows automatically by using the current day/month as end.
All other season-types should remain as is.

* Examples (Today: 25.01.2024)
  * Start (configured): 01.01.2023 => End (automatic): 25.01.2023
  * Start (configured): 01.01.2019 => End (automatic): 25.01.2019
  * Start (configured): 01.06.2021 => End (automatic): 25.01.2022

### DIFF
--- a/src/Core/Season.cpp
+++ b/src/Core/Season.cpp
@@ -36,6 +36,7 @@ static QList<QString> _setSeasonTypes()
 {
     QList<QString> returning;
     returning << "Season"
+              << "Year to date"
               << "Cycle"
               << "Adhoc"
               << "System";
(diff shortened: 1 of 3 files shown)


thejockl:
Hi @amtriathlon, let me think about your proposal, how this configuration could look like and how to use it. If I have a good idea I will come back to you. BTW, do you prepare the dev-list or a issue for such discussions?

amtriathlon:
As a general rule I prefer discussions at the forum, but since this has already been discussed there I created a follow up issue: https://github.com/GoldenCheetah/GoldenCheetah/issues/4512

PR #4440, opened 2024-01-29
bluetooth FTMS fix: if condition was always false, so no speed was re…

Odald:
…ported...

!FTMS_MORE_DATA == 0
flags & 0  == 0
-> if statement is always false

I tested this on a Elite DiretoX with "Tools..Options..Training..Preferences  Simulate Speed From Power disabled" (only then the problem is apparent.


amtriathlon:
It looks like a typo, thank you!

PR #4443, opened 2024-02-02
Qt6: Fixed a crash related to drag & drop

thejockl:
Qt6 changed the signature of the method mimeData in QTreeWidget and QTableWidget from
Q...::mimeData(const QList<...>) const
to
Q...::mimeData(const QList<...>&) const
therefore ignoring local implementations and falling back to the base-implementation with the default-serialization. This PR supports both Qt5 and Qt6 by a selecting the matching signature based on the Qt-version. Additionally the specifier override was added to Q...::mimeData, Q...::mimeTypes and some drag&drop-related event-handlers to prevent this kind of error for future versions.

### DIFF
--- a/src/Charts/LTMChartParser.cpp
+++ b/src/Charts/LTMChartParser.cpp
@@ -197,7 +197,7 @@ ChartTreeView::dropEvent(QDropEvent* event)
     }*/
 }
 
-QStringList 
+QStringList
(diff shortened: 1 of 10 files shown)


amtriathlon:
Excellent, now compare mode works in all 3 cases: Activities and Intervals in Activities View, Date Ranges in Trends View, thank you!

PR #4444, opened 2024-02-04
Charts/PfPvPlot (QA): allow vertical scales up to 2500N

gabrbedd:
The vertical axis on the QA plots (pedal force, Newtons) was stuck at 600N, even if the ride data exceeded it. The code that calculated the max force for the activity was apperently trying to throw out outliers (defined as "more than 2500N" -- 562 lbf)... but when finding the "max force" the value was seems to have accidentally been typed in as "255". This created logically dead code:

    maxAEPF = 600;
    ...
    if (aepf < 255 && aepf > maxAEPF) maxAEPF = aepf;

Thus, the max is never updated.

This patch changes the filter value from 255 to 2500 (as is indicated to be the intent elsewhere in the source file).


amtriathlon:
It looks like a typo, probably not detected because it only affects compare mode.
I will merge it, thank you!

PR #4445, opened 2024-02-09
Fixed some valgrind findings

thejockl:
Fixed some findings of valgrind (a tool to detect memory management bugs):
* Mismatched free() / delete / delete [] (HrPwPlot.cpp)
* Conditional jump or move depends on uninitialised value(s) (other files)

Some of the initializations where put into the header, some (if I saw better fit) to the constructor


amtriathlon:
Thank you!

PR #4448, opened 2024-02-20
Initialized mode when parsing a zwo-file

thejockl:
When zwo workout files are read, mode was not set but the original value kept. This could lead to wrong perspective switches in Train-mode.


amtriathlon:
Thank you!

PR #4449, opened 2024-02-25
Equipment feature

paulj49457:
This PR adds an equipment maintenance feature to GC, any feedback or guidance on future development would be useful.

**Design Aims:**
•	To be as generic as possible, support multiple equipment & components
•	To minimise the number of Metadata items added to activities
•	Minimise impact within GC, and simplify data/event management

**Description**
This PR will create a new unpopulated “EquipmentLink” metadata item in all activities to provide a mechanism to link an activity to an equipment defined in the new page. 

for example:
                                                  
image

Within the equipment pages, various user defined components can be attached to the equipment link with time-spans of use, allowing the calculation of distances covered by individual components.

The equipment page can be accessed via the new button on the sidebar

image

There are two navigators within the equipment page:
•	The equipment List
•	The equipment Links & distance references

image

**Equipment List**
Two types of entries can be created in this list, distance & time based equipment, they can be created by the Equipment List navigator menu:

image

Or once some equipment has been defined via the popup context menu via a right click:

image

**Distance Equipment:** supports the management of items which wear out, chains, cassettes, tyres, running shoes, etc
•	nonGCdistance: allows an initial value to be set for activities not recorded within GC
•	gcDistance: the distance contribution of all the relevant GC activities
•	Total distance: the total of  nonGCdistance + gcDistance
•	Replacement Distance: The entry is drawn in red text when this is exceeded, set a value of zero to disable warnings.

image

**Time equipment**: supports reminders for bike service intervals, hydraulic brake fluid changes, etc
•	These cannot have any children in the Navigator, as I couldn’t see a practical reason
•	The entry is drawn in red text when the replacement is overdue.

image

Note: Save must be pressed to store any text/date changes

**Equipment Link & Equipment References**
To link an equipment to activities, links need to be created in the Equipment List navigator menu using its menu option “Add Equipment Link”:

image

The example below shows an equipment link for “Kickr” and this name needs to match that in the activities EquipmentLink metadata item, it will automatically be added to the pick list to simplify selection. 

image

Once an equipment link is defined distance items from the equipment list can be dragged and dropped onto the equipment link to accumulate miles from that item. Time spans can also be set for equipment that might be used on one or more equipment (wheel, etc).

**Notes:**

•	There is no automatic recalculate of distances, this can be initiated from either the Equipment navigator menus, distances depend on your GC settings. With 2800 activities it takes less than a second to recalculate (using only 1/4 of the machines cores).
•	Equipment settings are saved on closing GC, they can also be save/loaded should you decide you want to undo a set of recent changes. The settings are store in the athletes config directory in equipments.xml, be careful if you want to directly edit this file.
•	This PR only defines & calculates distances for the current athletes’ activities. To expand this all for all GC athletes is a future enhancement should be deemed useful.
•	Instead of setting your equipment link to a bicycle (or similar) you could set all activities of one type with the same equipment link name (say running), then use the time spans to allocate distance to specific running shoes, see example above.
•	Distances are shown in [] on the list entries, and in the Equipment List they are totals, whilst in distance references they are distance for that use.
•	Durations are shown in <> on list entries.

Probably enough for now, please experiment with it and let me know your thoughts….


paulj49457:
I think in hindsight there is a simpler way, I'll create a new PR once its complete.

PR #4452, opened 2024-03-07
Qt6: Fixed flickering MainWindow

thejockl:
When opening a Mapchart for the first time after start of GC, the MainWindow used to shrink and be maximized again afterwards automatically, matching the description in
https://forum.qt.io/topic/141398/qwebengineview-closes-reopens-window-when-added-dynamically

By adding and removing a QWebEngineView before showing MainWindow, this flicker can be avoided

### DIFF
--- a/src/Charts/RideMapWindow.cpp
+++ b/src/Charts/RideMapWindow.cpp
@@ -137,6 +137,7 @@ RideMapWindow::RideMapWindow(Context *context, int mapType) : GcChartWindow(cont
     view->setContentsMargins(0,0,0,0);
     view->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
     view->setAcceptDrops(false);
+    view->page()->setHtml("<html></html>");
     layout->addWidget(view);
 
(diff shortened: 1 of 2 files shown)


amtriathlon:
Hi Joachim, regrettably it doesn't avoid the flicker on Windows builds.

thejockl:
Ok, will reactivate my windows-system and look into it

thejockl:
On my machines: Same behaviour on Linux and Windows
- Flicker is fixed when Golden Cheetah starts maximized
- Flicker is still there when Golden Cheetah starts unmaximized

amtriathlon:
Right, I see the same behavior. Previously I have tested non-maximized only, since this is the way I use GC daily.

thejockl:
Added empty HTML to the QWebEngineView of RideMapWindow, fixing the issue mostly for me. Exception: No map was present on startup in a non-maximized window and is only added later.
If this works also for you Ale, lets discuss which additional charts need a similar fix.

amtriathlon:
Yes, this works for me too, thank you!

PR #4453, opened 2024-03-08
Replaced the icons in the TrainBottomBar

thejockl:
* Replaced the icons from oxygen (png) by new ones from breeze (svg)
* Added support for dark- and light-mode
* Grouped the icons by function
* Added a hover-effect to the icon-buttons
* IMHO: The new theme fits better into the overall appearance

gc-dark
gc-light
gc-disconnected


thejockl:
Agree on the test with MacOS, unfortunately I only have Linux and Windows available

amtriathlon:
I tried a basic build on macOS and it doesn't look good, the main problem is the light version is selected even when Train view is dark (default), this doesn't happen on Windows or Linux in my tests.

thejockl:
Thanks for testing! Is it always light background with correct / visible icon or are the icons selected wrong (i.e. light background with icons intended for dark)? In the first case the fix will be easy (don't use CCHROME as background color). Can you please attach a screenshot?

amtriathlon:
I tried again and the bottom bar matches the sidebar, which I can't do is to change the sidebar/bottom bar to a darker color, so it looks like the problem is not related to your PR.

Screenshot 2024-03-21 at 19 36 29

thejockl:
I assume the issue is coming from Colors.cpp where CCHROME is set to #ececec for Mac only - and this is also used in NewSideBar.

Tomorrow I will update the PR to make the separators also visible in dark mode

amtriathlon:
> I assume the issue is coming from Colors.cpp where CCHROME is set to #ececec for Mac only - and this is also used in NewSideBar.

That is the default, the problem is it cannot be changed applying a new theme or selecting a different color, this is not related to your PR, off course.

amtriathlon:
After removing the "yosemite" special case and applying your last commits it looks like this:

Default:
Screenshot 2024-03-22 at 11 31 36

Modern Light:
Screenshot 2024-03-22 at 11 30 52

Modern Dark:
Screenshot 2024-03-22 at 11 29 32

PR #4455, opened 2024-03-10
Fixed rounding error when loading and saving ERG files

enngq11k:
Conversion from double to int can result in lower values. 
Example:
```cpp
int Ftp = 293;
double point = 246;
int watt = point / Ftp * Ftp; // watt = 245
```

This bug is hard to notice, since saving a workout will store the incorrect value, but
the UI still shows the correct value. You have to restart GC to spot it. Changing the FTP
setting and editing workouts can increase the difference even further.

Fix
```cpp
int watt = std::round(point / Ftp * Ftp); // watt = 246
```
I also added a check if a conversion is necessary.

This PR will only fix ERG files (GC Standard file format). `ErgFile.cpp` contains more
errors of this kind, but I can't reproduce and test them due to lack of data.

### DIFF
--- a/src/Train/ErgFile.cpp
+++ b/src/Train/ErgFile.cpp
@@ -26,6 +26,7 @@
 #include <QXmlSimpleReader>
 
 #include <stdint.h>
+#include <cmath>
 #include "Units.h"
 #include "Utils.h"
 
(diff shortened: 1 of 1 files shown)


amtriathlon:
Thank you!

PR #4462, opened 2024-03-26
ErgFile: Coalescing neighbouring sections

thejockl:
In erg-mode, some ErgFiles have neighbouring sections of same power. This is fine for creators / editors of workouts but gives "wrong" remaining section times from a athletes perspective, example:
* workout: 10 min at 200 watts followed by 5 min at 200 watts
* current situation: first counts down from 10:00 to 0, then from 5:00 to 0
* for an athlete, the countdown should be from 15:00 to 0 instead

This change
* creates an additional list of ErgFilePoints in ErgFile with neighbouring sections of same power coalesced
* uses the coalesced points in ErgFileQueryAdapter


thejockl:
This change proved highly useful to me (it is integrated into my local version since > 1 year). As I imported all ~2800 workouts from ErgDB, checking every single one manually would be a quite tedious task. This would alter the files on disk, creating deviations. A possible workoaround could be an external script, but this had to support all file formats (duplicating work already done in Golden Cheetah) and would exclude non-techie users.
Additionally, this PR supports a change I am currently working on: While the workout is not running, show the sections in ErgFilePlot in the colors of their power-zone plus hover effect and a tooltip (similar to the WorkoutEditor, but more verbose).
Therefore I would highly appreciate the inclusion.

Do you see possible issues as you mention the lack of alternatives? One thing I could enhance is to restrict the coalescing to watt based / erg-mode workouts.
The workout editor is not affected by this change as it still operates on the original Points.

amtriathlon:
> The workout editor is not affected by this change as it still operates on the original Points.

This is what I like the least: the workout being executed is different form what users can expect based on what they see in Workout Editor, and there is no option to preserve the original semantic (I mean, sections with same wattage can be separated for a reason).

thejockl:
Could you accept this PR if I added a global configuration option to toggle the feature?
* Keep the files on disk untouched unless saved with coalescing enabled
* If coalescing is enabled: Use these points globally, if saving a workout back: Write the coalesced points (maybe show a warning in the editor)
* If coalescing is disabled: Keep current behaviour

amtriathlon:
Providing the "Coalesce Neighboring Sections" runtime behavior be opt-in, I think your original proposal would be ok, since the files are not affected it would be trivial for the users to revert to the original behavior if they change their mind. A warning in Workout Editor may be in order when this option is enabled, perhaps in Properties, where the sections are easy to spot.

PR #4464, opened 2024-03-27
Bound global action to QKeySequence::Quit

thejockl:
This change allows to quit Golden Cheetah using a keyboard shortcut. QKeySequence::Quit is Ctrl+q on Linux and Cmd+q on MacOS but unbound on Windows, see https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts

Additionally set Qt::WaitCursor while shutting down Golden Cheetah to indicate the application is blocked


amtriathlon:
Alt+F4 still works as usual on Windows, thank you!

PR #4469, opened 2024-04-07
DialWindow: Scaling font in both dimensions

thejockl:
Currently the font in DialWindow is only scaled to make the text fit into the label vertically. This commit scales the font in both dimensions and prevents cutting off the content.

The font is rescaled
* If the DialWindow itself is resized
* If the text set and is longer than the previous one
* Every 10 text-changes

Before this PR:
Screenshot_20240429_173052

With this PR:
Screenshot_20240429_173040

### DIFF
--- a/src/Train/DialWindow.cpp
+++ b/src/Train/DialWindow.cpp
@@ -80,7 +80,10 @@ DialWindow::DialWindow(Context *context) :
     QVBoxLayout *layout = new QVBoxLayout;
     layout->setSpacing(0);
     layout->setContentsMargins(3,3,3,3);
-    valueLabel = new QLabel(this);
+    valueLabel = new ScalingLabel(this);
+    QFont vlFont = valueLabel->font();
+    vlFont.setWeight(QFont::Bold);
+    valueLabel->setFont(vlFont);
     valueLabel->setAlignment(Qt::AlignCenter | Qt::AlignVCenter);
     layout->addWidget(valueLabel);
     setChartLayout(layout);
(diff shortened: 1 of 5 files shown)


amtriathlon:
Thank you!

thejockl:
Hi @liversedge, I agree that the layout should be designed to show all content, but in practice it is not always that easy:
* Precision / Length differs between metrics
* Available space varies with sidebars / control pane shown or hidden, fullscreenmode enabled or disabled
* I have an old, low resolution but dedicated Notebook to control my Workouts. This also shifts the layout slightly

This PR helps me to at least see all requested information without anything cut off

I modified (only locally yet) my code to put all text on the same baseline, hoping to give a little relieve to your eyes. Could the PR become acceptable with this change?

Screenshot_20240519_202025

liversedge:
Looks horrible to me so maybe better to add a config to the dial window "fit contents to tile" ?

Then everyone is happy 😊

thejockl:
I added a global setting to "Options->Train->Preferences" that applies to all DialWindows. If you disagree, I am open to change it to a tile-specific setting, but this way it gives a consistent behaviour. Default is as before, i.e. Scale to fit height.
Additionally I changed the layout of this Page to QFormLayout, to be in line with the other Pages:

Screenshot_20240609_220628

liversedge:
I like this !

thejockl:
As this PR is already marked as closed by merge, should I open a followup-PR?

amtriathlon:
[GitHub doesn’t allow reopen once a PR is merged](https://stackoverflow.com/questions/12674304/github-reopening-a-merged-pull-request), you would need to create a new one.

Issue #4446, opened 2024-02-11
Drop down arrow not showing in Filter/Search input field


vischlatter:
The drop down arrow is not shown completely in the Filter/Search input field:
<img width="229" alt="image" src="https://github.com/GoldenCheetah/GoldenCheetah/assets/7930904/e8a87de7-2783-4576-81dd-fda7fbc9cf5a">

What I tried so far:

- Increase / decrease font size
- Use another screen
- Different window sizes

There is exactly one pixel that I can use to show the drop down. 

My setup:
MacBook Air M1, 2020, 8GB
MacOS: Sonoma 14.2.1

Golden Cheetah version: 
Build date: Aug 5 2023 14:06:41
Build id: 5000
Version: V3.6

Details:
<img width="291" alt="image" src="https://github.com/GoldenCheetah/GoldenCheetah/assets/7930904/2e9eb8d4-014b-42bf-8560-1b731e58c895">

amtriathlon:
There was an attempt to fix this on Windows, but it was reverted since it was broken on Linux and macOS, I would no expect this to be fixed anytime soon, unless Qt6 makes it easier, see https://github.com/GoldenCheetah/GoldenCheetah/commit/fc106615417b35469b97229767aaefcea7aab518

paulj49457:
Please see PR https://github.com/GoldenCheetah/GoldenCheetah/pull/4470 which aims to fix this issue.

PR #4470, opened 2024-04-08
Improved search/filter box drop down menu icon

paulj49457:
The following improves the accessibility of the search box/filter drop down menu, and it is now consistent with different themes, and hopefully resolves issue https://github.com/GoldenCheetah/GoldenCheetah/issues/4446 , see appearance below:

image

The QToolButton drop down triangle has been replaced by a QPushButton with an Icon, which is shown/hidden dependent on the search/filter box mode.

Can someone please test this on Mac/Linux build? As I'm not sure whether the MacOs specific style (line 55) should be **background-color : none** or the original **background: transparent**

### DIFF
--- a/src/Gui/SearchBox.cpp
+++ b/src/Gui/SearchBox.cpp
@@ -246,7 +246,11 @@ void SearchBox::resizeEvent(QResizeEvent *)
     searchButton->move(3*dpiXFactor, 6*dpiYFactor);
     setTextMargins(0, 0, frameWidth + tsz.width(), 0);
     clearButton->move(rect().right() - frameWidth - tsz.width() - cbz.width() - 3*dpiXFactor, 6*dpiYFactor);
+#ifdef Q_OS_MAC
+    toolButton->move(rect().right() - frameWidth - tsz.width() - 3*dpiXFactor, 3*dpiYFactor);
+#else
     toolButton->move(rect().right() - frameWidth - tsz.width() - 3*dpiXFactor, 6*dpiYFactor);
+#endif
 }
 
 void SearchBox::setFixedMode(bool fixed)
--- a/src/Gui/SearchBox.cpp
+++ b/src/Gui/SearchBox.cpp
@@ -33,12 +33,11 @@
 #include <QDebug>
 
 const int BUTTON_SIZE = 12;
-const int TB_ICON_SIZE = 10;
 
 SearchBox::SearchBox(Context *context, QWidget *parent, bool nochooser)
     : QLineEdit(parent), context(context), parent(parent), filtered(false), nochooser(nochooser), active(false), fixed(false)
 {
-    setFixedHeight(28 *dpiYFactor);
+    setFixedHeight(28*dpiYFactor);
     //clear button
     clearButton = new QToolButton(this);
     clearButton->setStyleSheet("QToolButton { border: none; padding: 0px; }");
@@ -52,7 +51,7 @@ SearchBox::SearchBox(Context *context, QWidget *parent, bool nochooser)
 
     // tool button
     toolButton = new QToolButton(this);
-    QIcon toolB = iconFromPNG(":images/sidebar/extra.png", QSize(TB_ICON_SIZE * dpiXFactor, TB_ICON_SIZE * dpiYFactor)); 
+    QIcon toolB = iconFromPNG(":images/sidebar/extra.png", QSize(BUTTON_SIZE*dpiXFactor, BUTTON_SIZE*dpiYFactor));
 #ifdef Q_OS_MAC
     toolButton->setStyleSheet("QToolButton { background: transparent; }");
 #else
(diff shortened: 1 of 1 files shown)


thejockl:
Looks good on Linux (Qt5, not yet tested with Qt6) and works as expected - much easier than before. The icon looks quite bulky to me, I would prefer a plain triangle.

One additional issue came to my attention while testing: After clearing the textfield, the placeholder is always dark, making it unreadable when using a dark theme. This was not introduced by your PR, master behaves the same.

marcenGC:
For testing,  I use Arch Linux with Qt6. It is really easier to find and no errors so far. I recommend using the current version where the triangle is available in the search and filter. This way, I can simply select a filter from the list when the search icon is selected, and GC will automatically switch to the filter icon.

paulj49457:
Thank you for the feedback, I have updated the PR:

image

- The filter drop down triangle is available in both search & filter (the original functionality)
- The triangle is now smaller, although the pick area is that of the icon, most of which is transparent.

The place holder text issue appears to be a Qt bug from my internet searches, I'll leave that for another day.

marcenGC:
Thanks for the update. It works well and the triangle is much easier to reach.

amtriathlon:
Regrettably it looks horrible on macOS:

SearchMenu

paulj49457:
Would someone with a MacOs like to work to correct the Mac specific display issues? As I think for Windows and Linux this PR has made accessing the drop down menu better, and also it display correctly across the different themes

- I think the white backgound on MacOs is probably due to not switching the background to either "background: none" or background : transparent", see src/Gui/SearchBox.cpp line 55

- I guess the sizing issue could be achieved by some Mac specific code to adjust the widget size, the intention of the PR is to have QPushButton the size of the Icon (  QSize(SEARCH_ICON*dpiXFactor, SEARCH_ICON*dpiYFactor)  ) without a text label, this only needs to be done once so would go in the constructor with the other widget creation.

marcenGC:
Perhaps it would be an option to move the drop-down menu to the right-hand side of the search field and make it the same height as the search field?

amtriathlon:
> Perhaps it would be an option to move the drop-down menu to the right-hand side of the search field and make it the same height as the search field?

That could be nice and functional, similar to GitHub search bar:
image

paulj49457:
Ok, sounds good to me, I'll move the filter drop down button to the right as suggested.

paulj49457:
Updated as suggested in commit https://github.com/GoldenCheetah/GoldenCheetah/pull/4470/commits/5df738925412c5762d0ec933897287ce094ed456, needs testing on Linux & MacOs. Please update the icon if you want a different appearance.

a1 - Screenshot 2024-04-14 162930
a2 - Screenshot 2024-04-14 163201
a3 - Screenshot 2024-04-14 165110
a4 - Screenshot 2024-04-14 165145

amtriathlon:
On macOS it looks much better than before, may be too subtle, perhaps we would need to change the icon for more visibility:

image

image

marcenGC:
Perhaps the Big Mac menu from the side bar. Most users know the function of the symbol.
Screenshot_20240414_204228-1

paulj49457:
I have changed the icon to the "Big Mac menu" and hopefully that will be better, see lastest commit:

Windows: 
image

marcenGC:
For testing, I use Arch Linux with Qt6.  Works with all three different search bars.
Screenshot_20240415_001118

amtriathlon:
It looks like the above in macOS too, both using light and dark themes, in the toolbar and dialogs, which is very nice, IMHO.

I get this warning:
```
Gui/SearchBox.cpp:246:11: warning: variable ‘sbz’ set but not used [-Wunused-but-set-variable]
246 |     QSize sbz = searchButton->sizeHint();
```
which should be removed before merging.

Also there seems to be a mix of positions multiplied by dpi[X|Y}factor and others in absolute numbers, this needs to be carefully reviewed since it can work when the factors are 1 but not in other cases. I am not sure the use of one time constants like SEARCHB_X makes the code easier to read too.

thejockl:
Looks mostly good on Debian / Qt5, light and dark themes work perfectly: 
image

"Mostly" because the three icons (filtertype, clear and burger) are not horizontally aligned: The clear button is above the two others, making it  look out of place. Besides that: A great improvement!

paulj49457:
I'll address the dpi[X|Y}factors, sbz & the vertical alignment later today and update the PR

amtriathlon:
Quick test on macOS, it looks ver nice:

Screenshot 2024-04-15 at 15 19 25
Screenshot 2024-04-15 at 15 19 58

Changing vertical position from 3 to 6 for the clear button seems to be even better:
Screenshot 2024-04-15 at 15 29 25
Screenshot 2024-04-15 at 15 28 35

paulj49457:
See PR https://github.com/GoldenCheetah/GoldenCheetah/pull/4475 for placeholder text fix

amtriathlon:
Thank you!

marcenGC:
Even if the PR is already merged in the overview tile, the filter and search bar is now missing.

PR #4475, opened 2024-04-16
Search/filter placeholder text fix

paulj49457:
This PR fixes:

- The placeholder text which disappears on dark themes now remains visible.

Screenshot 2024-04-16 201506

Screenshot 2024-04-16 201546

- The red text is now displayed (and removed) when invalid filter text is entered.

Screenshot 2024-04-16 201724

Screenshot 2024-04-16 201634

- The colour of the group-by menu has now been fixed, previously on Windows it appeared as:

Group-by menu

- The backgrounds of the "Manage Filters" dialogue QLineEdit fields for the name, searchbox and filter list would pickup the theme (noticeable on dark themes) depending where they were launched from:

**From the mainWindow Toolbar:**

Screenshot 2024-05-19 182832

**From the Trends, Manage Filters:**

Screenshot 2024-05-19 184500

- The "Manage Filters" option in the pull down menu has been removed from the "Manage Filters" dialogue, otherwise this just opens another instance of the dialogue for no benefit, and possible accidental filter overwrite problems.

- Added a close button to the bottom right of the "Manage Filters" dialogue, hopefully preventing accidentally deleted filters.

- Removed "column chooser" from  the "Manage Filters" pull down, as the option is already available from the ride navigator menu, and IMHO it is more logically associated with that location, I can add it back if you want it retained.

I have tested the above with most dark & light themes on Windows and all seems to work, so can it be tested on Linux & MacOs ?

Note: All popup windows and dialogues do not respect the selected theme, and use a default light theme, this this was part of the problem of hidden text, I expect it would a significant project to get them to adopt the theme colours.

### DIFF
--- a/src/Core/NamedSearch.cpp
+++ b/src/Core/NamedSearch.cpp
@@ -256,7 +256,7 @@ EditNamedSearches::EditNamedSearches(QWidget *parent, Context *context) : QDialo
     layout->addLayout(row2);
     QLabel *filter = new QLabel(tr("Filter"), this);
     row2->addWidget(filter);
-    editSearch = new SearchBox(context, this, true);
+    editSearch = new SearchBox(context, this, true, false);
     row2->addWidget(editSearch);
 
     // add/update buttons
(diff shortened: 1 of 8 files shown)


marcenGC:
I have tested the PR under Archlinux. One thing I noticed is that the red colour is not always reset.
1. enter a  wrong filter value that you see him in red.
2. call up the Manage Filters and close it
3. then delete the filter value with the X symbol
4. now the newly entered text is also red if the value is correct.

Here is a short video. Unfortunately, the Manage Filters is not displayed.
https://github.com/GoldenCheetah/GoldenCheetah/assets/125665140/bd0d859d-976a-4d77-a401-64557d272740

paulj49457:
@marcenGC Thank you for testing the PR (and the video), I'll take another look to see why the red colour is not reset in this case.

paulj49457:
@marcenGC The problem you found is now fixed in the latest commit.

I'd used a couple of static parameters to limit the number of updates in the new setSearchBoxStyle() function, but opening the Manage Filters creates another instance of the SearchBox and this caused the update limit functionality to fail, they are now private member variables.

marcenGC:
Thank you for the fix. It works fine now.

paulj49457:
I have updated the PR description with the latest fixes, please test on Linux & MacOs, thanks.

thejockl:
Test on Debian with Qt5 here. Looks good to me with one minor exception: When switching the color-scheme having an active error in the filter, the red text turns to the default color

paulj49457:
> Test on Debian with Qt5 here. Looks good to me with one minor exception: When switching the color-scheme having an active error in the filter, the red text turns to the default color

Thank you for testing, I hadn't tried that corner case :)  I guess that'll be the configChanged not checking the parsing error state, I'll wait to see what else gets reported.

amtriathlon:
It is harder to see the latest changes if you force push... Anyway, I think this PR is becoming increasingly complex and difficult to check it doesn't generate other issues just to fix a barely noticeable cosmetic issue, I am not willing to merge this but, off course, @liversedge has the last word.

paulj49457:
@amtriathlon, I've now split up the most useful changes into smaller PRs, and I'll aim to keep future PR's focused on a single problem, and avoid any force pushing of changes.

PR #4478, opened 2024-04-20
Changed the standard colors for compare-mode

thejockl:
Currently, the standard colors used for compare-mode do not blend harmonically into the other defaults. Also some of the (e.g. yellow) are hardly visible on a light background and had to be manually changed every time. This PR tries to address this issue.

Changed the standard colors, used for compare-mode:
* Nice looking and distinguishable colors
* Separate color-sets for light- and darkmode

Additional fix + refactoring:
* Fixed a typo that prevented the background-color of ComparePanes to follow the color-scheme (was always white on my machine)
* Moved the definition of standardColor(int) from ComparePane.cpp to Colors.cpp, matching the declaration in Colors.h

Screenshot_20240420_231757
Screenshot_20240420_231932

### DIFF
--- a/src/Gui/Colors.cpp
+++ b/src/Gui/Colors.cpp
@@ -63,28 +63,6 @@ int pixelSizeForFont(QFont &font, int height)
     return pixelsize;
 }
 
-//
-// A selection of distinct colours, user can adjust also
-//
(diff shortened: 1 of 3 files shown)


amtriathlon:
Thank you!

PR #4482, opened 2024-04-29
Initializing AllPlot::isPanning to false

thejockl:
AllPlot::isPanning was uninitialized, leading to unwanted (i.e. without prior middleclick) panning.
This PR initializes AllPlot::isPanning explicitly to false.


amtriathlon:
Thank you!

PR #4483, opened 2024-05-04
Qt6: Preventing crash in stacked LTMPlots

thejockl:
In Qt6 the internal handling of QList (now actually a QVector) was changed compared to Qt5. This results in reallocations / recreations of the LTMSettings, destroying the objects whose pointers already have been set to the LTMPlots.
This commit reserves the required space upfront, preventing the reallocations thus keeping the pointers valid and preventing the crash


amtriathlon:
Thank you!

PR #4488, opened 2024-05-09
Qt6: The signature of enterEvent changed

thejockl:
Adapted to the new signature of enterEvent when using Qt6 (before this change enterEvent was ignored with Qt6); added some additional override specifiers


amtriathlon:
Thank you!

PR #4492, opened 2024-05-20
Fixes the background colour of the group by sub menu

paulj49457:
Fixes the background colour of the group by sub menu:

image

### DIFF
--- a/src/Core/NamedSearch.cpp
+++ b/src/Core/NamedSearch.cpp
@@ -301,6 +301,9 @@ EditNamedSearches::EditNamedSearches(QWidget *parent, Context *context) : QDialo
     row4->addStretch();
     deleteButton = new QPushButton(tr("Delete"), this);
     row4->addWidget(deleteButton);
+    row4->addStretch();
+    closeButton = new QPushButton(tr("Close"), this);
+    row4->addWidget(closeButton);
 
     // Populate the list of named searches
     foreach(NamedSearch x, context->athlete->namedSearches->getList()) {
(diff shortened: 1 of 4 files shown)


amtriathlon:
This triggers the following warning on Linux:
`[20:21:11.289] warning: Creating a popup with a parent, QWidgetWindow(0x555b10d2c480, name="MainWindowClassWindow") which does not match the current topmost grabbing popup, QWidgetWindow(0x555b21da9ce0, name="QMenuClassWindow") With some shell surface protocols, this is not allowed. The wayland QPA plugin is currently handling it by setting the parent to the topmost grabbing popup. Note, however, that this may cause positioning errors and popups closing unxpectedly. Please fix the transient parent of the popup. (unknown:0)`

paulj49457:
Thank you for testing.
Can you tell me whether the sub menu colours are correct on Linux & macOs without this fix? If they are correct, then maybe this can implemented as a compiler switch, as I didn't see any warnings on the windows/visual studio compilation. Otherwise I'll investigate another solution.

amtriathlon:
Yes, colors are correct on Linux, I have not tested on macOS yet.
The warning above is not related to this change, it happens even without it.

marcenGC:
Under Arch linux the colours are only correct if KDE 6 and GC are set to dark themes. When KDE 6 light theme is selected and GC dark, the colors are incorrect.
Image shows when KDE 6 and GC are set to dark themes
Screenshot_20240521_191439

The image shows KDE 6 with a light theme and GC with dark themes.
Screenshot_20240521_191315

amtriathlon:
> This triggers the following warning on Linux: `[20:21:11.289] warning: Creating a popup with a parent, QWidgetWindow(0x555b10d2c480, name="MainWindowClassWindow") which does not match the current topmost grabbing popup, QWidgetWindow(0x555b21da9ce0, name="QMenuClassWindow") With some shell surface protocols, this is not allowed. The wayland QPA plugin is currently handling it by setting the parent to the topmost grabbing popup. Note, however, that this may cause positioning errors and popups closing unxpectedly. Please fix the transient parent of the popup. (unknown:0)`

Using menu.window() as parent of groupByMenu in
https://github.com/GoldenCheetah/GoldenCheetah/blob/5e3db39a4902855baeb56ed8dfc92db5986992df/src/Gui/AnalysisSidebar.cpp#L427
instead of `rideNavigator`(or `this`as proposed in this PR) fixes the above warning and the color issue on Dark themes, both on Windows and Linux in my tests.

PR #4493, opened 2024-05-20
Add close button to manage filters

paulj49457:
Add close button to manage filters, hopefully reducing the opportunity to delete filters

close button


amtriathlon:
Thank you!

PR #4494, opened 2024-05-20
Placeholder text & filter background color field fixes (simple version)

paulj49457:
Placeholder text & filter background colour field fixes:

The placeholder text which disappears on dark themes now remains visible.
image

image

The red text is now displayed (and removed) when invalid filter text is entered.
image

image

The backgrounds of the "Manage Filters" dialogue QLineEdit fields for the name, searchbox and filter list would pickup the theme (noticeable on dark themes) depending where they were launched from:

**From the mainWindow Toolbar:**

image

**From the Trends, Manage Filters:**

image

I have tested the above with most dark & light themes on Windows and all seems to work, so can it be tested on Linux & MacOs ?

Note: All popup windows and dialogues do not respect the selected theme, and use a default light theme, this this was part of the problem of hidden text, I expect it would a significant project to get them to adopt the theme colours.

### DIFF
--- a/src/Core/NamedSearch.cpp
+++ b/src/Core/NamedSearch.cpp
@@ -302,6 +302,26 @@ EditNamedSearches::EditNamedSearches(QWidget *parent, Context *context) : QDialo
     deleteButton = new QPushButton(tr("Delete"), this);
     row4->addWidget(deleteButton);
 
+    // The QlineEdit & QTreeWidget pick up the style from the location they launched and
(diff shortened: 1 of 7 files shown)


amtriathlon:
Hi Paul, sorry, but this will not be merged, as commented before:
> I think this PR is becoming increasingly complex and difficult to check it doesn't generate other issues just to fix a non-functional barely noticeable cosmetic issue.

May be when and if the changes necessary for dialogs to honor themes be implemented we can device a simpler solution.

PR #4497, opened 2024-05-22
Updated qwt to 6.3 (qwt-multiaxes branch)

thejockl:
Using qwt from https://sourceforge.net/p/qwt/git/ci/qwt-multiaxes/tree/

Applied the following changes to qwt 6.3:
* Added QwtZone to qwt_plot_curve.cpp
* Disabled the emitting of Layout Requests on geometry changes of QwtScaleWidget - without this, CPU utilization was up to 100% on one core


amtriathlon:
Thank you!

thejockl:
In sourceforge git, qwtconfig.pri did not change, but in my opinion QWT_VER_MIN should be bumped from 4 to 8 (this release reports 6.8.0 as its version). Should I change and push this?

amtriathlon:
Agree, and don't worry, I will do that, thank you!

PR #4501, opened 2024-05-24
Corrections to qt colors passed to GColor macro

paulj49457:
QT colors ( https://doc.qt.io/qt-6/qt.html#GlobalColor-enum ) being used in GColor macro when GC Color references are expected:

**OverviewItems.h**

    `BPointF() : x(0), y(0), z(0), xoff(0), yoff(0), fill(GColor(Qt::gray)), item(NULL) {}`
A gc color should be passed to the GColor macro not a qt color, so this is now:

    `BPointF() : x(0), y(0), z(0), xoff(0), yoff(0), fill(GColor(CRIDEPLOTXAXIS)), item(NULL) {}`
as Qt::gray has the value of 5, so its replaced with RIDEPLOTXAXIS which has the same value.

**ExtendedCriticalPower.cpp**

    `QPen e2pen(GColor(Qt::lightGray));`
A gc color should be passed to the GColor macro not a qt color, so this is now:

    `QPen e2pen(GColor(CRIDEPLOTYAXIS));`
as Qt::lightGray has the value of 6, so its replaced with RIDEPLOTYAXIS which has the same value.

### DIFF
--- a/src/Charts/OverviewItems.h
+++ b/src/Charts/OverviewItems.h
@@ -552,7 +552,7 @@ class IntervalOverviewItem : public ChartSpaceItem
 class BPointF {
 public:
 
-    BPointF() : x(0), y(0), z(0), xoff(0), yoff(0), fill(GColor(Qt::gray)), item(NULL) {}
+    BPointF() : x(0), y(0), z(0), xoff(0), yoff(0), fill(GColor(CRIDEPLOTXAXIS)), item(NULL) {}
 
     double score(BPointF &other);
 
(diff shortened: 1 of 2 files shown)


paulj49457:
@amtriathlon I have extracted the bug fixes from a couple of my recent PRs which got too large & complex and were closed.
So please take a look and apply those you think useful, otherwise close them.

amtriathlon:
I don't think complexity is a problem if it is necessary, the reason I closed your PRs was because that complexity wasn't, for example a whole refactor to change a static class into a singleton when there is no inheritance involved, just to add some additional static methods like you did recently is more reasonable.
I plan to merge that one and others fixing bugs, but not the ones intended to fix cosmetic issues due to dialogs not honoring themes since they add unnecessary complexity without fixing the real cause.

amtriathlon:
Thank you

PR #4502, opened 2024-05-24
Error found in GColor function readConfig

paulj49457:
Error found in GColor function readConfig, the name field is a text description of the gc color, but it is being compared against the setting value, therefore these defaults will never be applied:

        `// set sensible defaults for any not set (as new colors are added)`
        `if (ColorList[i].name == "CTOOLBAR") {`
        `    QPalette def;`
        `    ColorList[i].color = def.color(QPalette::Window);`
        `}`
        `if (ColorList[i].name == "CCHARTBAR") {`
        `    ColorList[i].color = ColorList[CTOOLBAR].color;`
        `}`
        `if (ColorList[i].name == "CCALCURRENT") {`
        `    QPalette def;`
        `    ColorList[i].color = def.color(QPalette::Highlight);`

Example names & settings:

ColorList[i].name = "Plot Background"                           ColorList[i].setting = COLORPLOTBACKGROUND
ColorList[i].name = "Performance Plot Background"     ColorList[i].setting = COLORRIDEPLOTBACKGROUND
ColorList[i].name = "Trend Plot Background"                 ColorList[i].setting = COLORTRENDPLOTBACKGROUND
etc....

### DIFF
--- a/src/Gui/Colors.cpp
+++ b/src/Gui/Colors.cpp
@@ -423,6 +423,7 @@ GCColor::readConfig()
     for (unsigned int i=0; ColorList[i].name != ""; i++) {
         QString colortext = appsettings->value(NULL, ColorList[i].setting, "").toString();
         if (colortext != "") {
+
             // color definitions are stored as "r:g:b"
             QStringList rgb = colortext.split(":");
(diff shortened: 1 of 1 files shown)


amtriathlon:
Thank you!

PR #4503, opened 2024-05-25
Remove duplicate theme color code

paulj49457:
This PR removes the duplicate theme color application code within ColorsPage::applyThemeIndex(int index) and CColor::applyTheme(int index), and they both now call the static function getThemeColor() within GColor class, hopefully this simplify any future maintenance.

This detected the following issue:

```
    case CHOVER:
        // stealthy themes use overview card background for hover color since they are close
        // all other themes get a boring default
        color = theme.stealth ? ColorList[96].color : (theme.dark ? QColor(50, 50, 50) : QColor(200, 200, 200));

        // color = theme.stealth ? theme.colors[11] : (theme.dark ? QColor(50, 50, 50) : QColor(200, 200, 200)); - Pages.cpp version
        break;
```

And the default case:

```
    default:
        if (theme.dark) color = DarkDefaultColorList[colorIdx].color;
        else color = LightDefaultColorList[colorIdx].color;

        // color = colorSet[colorIdx].color;  -- Pages.cpp version
        break;
```

I expect that as both pages.cpp and colors.cpp apply the theme at different times, then these cases should be identical? If not, a parameter can be added to getThemeColor to select between the differences in these options.


amtriathlon:
> I expect that as both pages.cpp and colors.cpp apply the theme at different times, then these cases should be identical? If not, a parameter can be added to getThemeColor to select between the differences in these options.

It looks like the Colors version is the more recent in both cases according to the commits that produced the difference: https://github.com/GoldenCheetah/GoldenCheetah/commit/c7d7972213f8bbae63417d84ce321650c5589f20 and https://github.com/GoldenCheetah/GoldenCheetah/commit/d83041282d2920c8cf538e00e6bf7fd9f1846b3f, the Pages version applies only when the them is selected and applied while the Colors version in all other cases (Startup and Upgrade) but I don't see a reason for they to be different, I think this refactor makes sense.

paulj49457:
I have removed the comments relating to the old pages.cpp code.

amtriathlon:
Thank you!

PR #4504, opened 2024-05-26
Correct popup menus QLineEdit backgrounds

paulj49457:
The backgrounds of the "Manage Filters" dialogue QLineEdit fields for the name, searchbox and filter list would pickup the theme (noticeable on dark themes) depending where they were launched from:

From the mainWindow Toolbar:

Screenshot 2024-05-20 194611

From the Trends, Manage Filters:

Screenshot 2024-05-07 072027


amtriathlon:
It looks like a simpler solution would be to change the parent of the manage filters dialog:

```
diff --git a/src/Gui/LTMSidebar.cpp b/src/Gui/LTMSidebar.cpp
index a98c4e1ef..aec36a67c 100644
--- a/src/Gui/LTMSidebar.cpp
+++ b/src/Gui/LTMSidebar.cpp
@@ -724,7 +724,7 @@ LTMSidebar::eventPopup()
 void
 LTMSidebar::manageFilters()
 {
-    EditNamedSearches *editor = new EditNamedSearches(this, context);
+    EditNamedSearches *editor = new EditNamedSearches(parentWidget(), context);
     editor->move(QCursor::pos()+QPoint(10,-200));
     editor->show();
 }
```
It doesn't seem to have negative side effects, but needs more testing.

amtriathlon:
See https://github.com/GoldenCheetah/GoldenCheetah/pull/4506#issuecomment-2137490349

PR #4506, opened 2024-05-26
SearchFilterBox background

paulj49457:
All the FilterSearchBox instances within dialog use a light theme independent of the GC theme except the FilterSearchBox located within the main window toolbar, this PR ensures the different instances have the correct colours.

Screenshot 2024-05-20 194611 v2

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -336,7 +336,7 @@ MainWindow::MainWindow(const QDir &home)
     HelpWhatsThis *helpPerspectiveSelector = new HelpWhatsThis(perspectiveSelector);
     perspectiveSelector->setWhatsThis(helpPerspectiveSelector->getWhatsThisText(HelpWhatsThis::ToolBar_PerspectiveSelector));
 
-    searchBox = new SearchFilterBox(this,context,false);
+    searchBox = new SearchFilterBox(this,context,false, true);
 
     searchBox->setStyle(toolStyle);
     searchBox->setFixedWidth(400 * dpiXFactor);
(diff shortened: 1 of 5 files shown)


amtriathlon:
Similar to the other case, It looks like a simpler solution would be to change the parent of the manage filters dialog:

```
diff --git a/src/Gui/SearchBox.cpp b/src/Gui/SearchBox.cpp
index ffd7507ac..cf09d93b4 100644
--- a/src/Gui/SearchBox.cpp
+++ b/src/Gui/SearchBox.cpp
@@ -354,7 +354,7 @@ void SearchBox::runMenu(QAction *x)
     if (x->text() == tr("Add to Named Filters")) addNamed();
     else if (x->text() == tr("Manage Filters")) {

-        EditNamedSearches *editor = new EditNamedSearches(this, context);
+        EditNamedSearches *editor = new EditNamedSearches(this->parent->parentWidget(), context);
         editor->move(QCursor::pos() - QPoint(460, -5));
         editor->show();
```
It doesn't seem to have negative side effects, but needs more testing.

amtriathlon:
On Windows and macOS changing the parent of EditNamedSearches to the MainWindow fixes the color issues without side effects, on my tests on Linux (Ubuntu 22.04 over WSL) the manage filters dialog appears centered on MainWindow, this likely depends on the windows manager and it looks like a minor inconvenience (a little more mouse movement is necessary).

paulj49457:
@amtriathlon Thank you for fixing the background colours, I noticed that context->mainWindow also works.

amtriathlon:
Yes, and it seems better, more explicit and robust, I will change that.

Issue #4508, opened 2024-05-31
Startup crash when there are no Power Zones for the athlete


amtriathlon:
This has been reported at the users forum and it happens after #4400, which assumes there is always power zones available for Bike and the current date, but this can be not true for pure runners, swimmers, etc.

There are at least 2 places where this can happen:

https://github.com/GoldenCheetah/GoldenCheetah/blob/804a9046d940ce5ad3f9db27f6bca138584f6882/src/Train/TrainSidebar.cpp#L290-L295

https://github.com/GoldenCheetah/GoldenCheetah/blob/804a9046d940ce5ad3f9db27f6bca138584f6882/src/Train/TrainSidebar.cpp#L702-L707

And it should be replaced by something like:

https://github.com/GoldenCheetah/GoldenCheetah/blob/804a9046d940ce5ad3f9db27f6bca138584f6882/src/Train/TrainSidebar.cpp#L92-L96

thejockl:
I will look into this

thejockl:
See #4509 for my proposed fix

amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/e592645cfb89d76d22af611a7cc627577da73d1b

PR #4509, opened 2024-05-31
Fixed crash on startup if no power zones for bikes are defined

thejockl:
Fix for #4508:
* Safeguarded reading zones in TrainSidebar.cpp
* Added additional check to prevent PowerZonesWidget from painting itself if zone list is empty
* Returning a empty list from multiple methods in class Zones if given range is out of allowed range (was: check only for range too high, below 0 was ignored):
  * Zones::getZoneLows(.)
  * Zones::getZoneHighs(.)
  * Zones::getZoneNames(.)
  * Zones::getZoneDescriptions(.)


amtriathlon:
Thank you! 
Zones object for Bike is warranted to exist now, son no need to check for the nullptr anymore.

PR #4515, opened 2024-06-22
ErgFileplot: Updating the zone-coloring on changed intensity

thejockl:
This is a followup to #4479, taking care of manually changed intensity:
* Adjusting the coloring when sections change their power zone
* Adjusting the W'Balance prediction

Screenshot_20240622_100417
Screenshot_20240622_100448


amtriathlon:
Thank you!

Issue #4516, opened 2024-06-24
Add tool tips to trainer control buttons


NickHastings:
The trainer control buttons have no text and the meanings of all of the icons are not immediately clear.
As such it would be helpful if these buttons had tool tips with text to more clearly indicate what they do.

PR #4517, opened 2024-06-24
Add tool tips to trainer controls

NickHastings:
Fixes https://github.com/GoldenCheetah/GoldenCheetah/issues/4516. Note that this implementation hides the tool tips if the corresponding buttons are disabled.

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -238,6 +238,7 @@
 #define TRAIN_USESIMULATEDSPEED         "<global-trainmode>train/usesimulatedspeed"
 #define TRAIN_USESIMULATEDHYPOXIA       "<global-trainmode>train/usesimulatedhypoxia"
 #define TRAIN_COALESCE_SECTIONS         "<global-trainmode>train/coalesceSections"
+#define TRAIN_TOOLTIPS                  "<global-trainmode>train/tooltips"
 #define TRAIN_TELEMETRY_FONT_SCALING    "<global-trainmode>train/telemetryFontScaling"
 #define GC_REMOTE_START                 "<global-trainmode>remote/start"
 #define GC_REMOTE_STOP                  "<global-trainmode>remote/stop"
--- a/src/Gui/Pages.cpp
+++ b/src/Gui/Pages.cpp
@@ -757,6 +757,9 @@ TrainOptionsPage::TrainOptionsPage(QWidget *parent, Context *context) : QWidget(
     coalesce = new QCheckBox(tr("Coalesce contiguous sections of same wattage"), this);
     coalesce->setChecked(appsettings->value(this, TRAIN_COALESCE_SECTIONS, false).toBool());
 
+    tooltips = new QCheckBox(tr("Enable Tooltips"), this);
+    tooltips->setChecked(appsettings->value(this, TRAIN_TOOLTIPS, true).toBool());
+
     telemetryScalingLabel = new QLabel(tr("Telemetry font scaling"));
     telemetryScaling = new QComboBox();
     telemetryScaling->addItem(tr("Fit to height only"), 0);
@@ -785,6 +788,7 @@ TrainOptionsPage::TrainOptionsPage(QWidget *parent, Context *context) : QWidget(
     all->addRow("", autoHide);
     all->addRow("", lapAlert);
     all->addRow("", coalesce);
+    all->addRow("", tooltips);
     all->addRow(delayLabel, startDelay);
     all->addRow(telemetryScalingLabel, telemetryScaling);
 }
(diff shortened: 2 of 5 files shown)


amtriathlon:
Hi, thanks for your contribution!
I think to add/remove tooltips according to button state is too cumbersome and not necessarily desirable in this case: it may be more useful to know what each button does before starting the session than while training.
Additionally text strings to be shown to users need to be enclosed by tr() macro to enable translations to other languages than english.

NickHastings:
Thanks for the feedback. I wrapped the non-null strings in the `tr()` macro.

I think it important to make it clear that I am *not* in any way trained in UI/UX.

When I first implemented this I just had static tool tips. But then I found that the tool tip for the connect/disconnect button was either wrong or needed to be something like (dis)connect" or "connect/disconnect" which I thought to be needlessly ugly/cumbersome. This lead me down a bit of a rabbit hole reading about the "correct" way to do tool tips. I couldn't find one. It seems that there are proponents for removing the tool tip for disabled buttons but there are also people who think that it confuses things since then it may not be clear why the button is disabled. I did find that in the UI/UX world disabled buttons seem to be frowned upon.

One option may be to keep tool tips for disabled buttons (which s discouraged in some literature) and dynamically change the tool tip for the connect/disconnect button.

I don't know what is the right way to do this, but I do know that the first time that I really would have appreciated tool tips for these buttons was while I was actually riding.

amtriathlon:
> I wrapped the non-null strings in the `tr()` macro.

Thank you

> One option may be to keep tool tips for disabled buttons (which is discouraged in some literature) and dynamically change the tool tip for the connect/disconnect button.

I think this a better option, most buttons are disabled before to start the session when the user is more confortable to find out what they do.
 
> I don't know what is the right way to do this, but I do know that the first time that I really would have appreciated tool tips for these buttons was while I was actually riding.

IMHO existing users will hate this feature due to being obtrusive and unnecessary once you already know what the buttons do, so they should be able to disable them in trainer options.

thejockl:
> > One option may be to keep tool tips for disabled buttons (which is discouraged in some literature) and dynamically change the tool tip for the connect/disconnect button.
> 
> I think this a better option, most buttons are disabled before to start the session when the user is more confortable to find out what they do.
> 
> > I don't know what is the right way to do this, but I do know that the first time that I really would have appreciated tool tips for these buttons was while I was actually riding.
> 
> IMHO existing users will hate this feature due to being obtrusive and unnecessary once you already know what the buttons do, so they should be able to disable them in trainer options.

Might it be an alternative to bind individual WhatsThis-Help entries to each button? Currently the help short text corresponds to the whole TrainBottom.
The user would have to click the "?"-icon on the upper right corner, first, then the button of interest.
Personally I am not the biggest fan of this feature but maybe it could be a compromise

Reminder to self: I have to update the screenshot on the wiki-page for v3.7

NickHastings:
> > I don't know what is the right way to do this, but I do know that the first time that I really would have appreciated tool tips for these buttons was while I was actually riding.
> 
> IMHO existing users will hate this feature due to being obtrusive and unnecessary once you already know what the buttons do, so they should be able to disable them in Train preferences.

Until accidentally clicking the corresponding button during a ride I had no idea that GC had a calibration facility: that is what finally prompted me to dive into the code and add the tool tips. FYI I've been using GC for indoor rides for over 7 years.

What about a config option to enable/disable tool tips. Have them enabled by default, then anyone sufficiently annoyed by them can toggle them off. I don't know how hard it would be to implement this but I'd be willing to look into it if would likely be accepted.

NickHastings:
> Might it be an alternative to bind individual WhatsThis-Help entries to each button? Currently the help short text corresponds to the whole TrainBottom. The user would have to click the "?"-icon on the upper right corner, first, then the button of interest. Personally I am not the biggest fan of this feature but maybe it could be a compromise

Yikes, I don't think I ever clicked on or even noticed that button!

I see how it could likely be used to provide the required information, but as far as I can recall I've never seen that style of "help" in a GUI. Am I odd one out here? Are most people familiar with that feature?

liversedge:
Like splash screens, whats this is very old school and largely replaced in modern UX
if the user needs to click to find out what something is then the design needs improving

NickHastings:
I made the update. Description in the commit message.

While testing I noticed that the preference dialog would automatically close after clicking [save] and that the ant connection to the trainer would be lost. I see this behavior also using b4eea7f (current head of the master branch) so it seems to be unrelated to my changes.

I do *not* see this behavior on the v3.7-DEV2404 appimage. Note however that is running on a different machine with a different config. I could do more detailed testing if needed.

amtriathlon:
Thank you for the changes, it is ok for me now, and I will merge it if there are no objections.

`What's this` currently is used no so much literally -granularity is too coarse for that- but as a -supposedly simple- way to access a link to the relevant part of the documentation, which the user will need to read sooner or later for any non-trivial use of GC, IMHO.

NickHastings:
Great. Should I squash the two commits into one?

amtriathlon:
No, GitHub squash and merge is very handy for PRs.

amtriathlon:
If you has to do changes please do a normal push, so other participants can spot them, if you force-push we have to review all the changes again.

NickHastings:
You can see the changes in a force push In the github PR UI.

amtriathlon:
Thank you!

PR #4518, opened 2024-07-02
Metadata tile initial creation update fix and time & date field support

paulj49457:
When metadata tiles are initially created they are blank, and only update when the config menu is opened and closed.

The reason for this is the AddTileWizard initially creates a default MetaOverviewItem which has fieldtype=-1, then it calls setData on the MetaOverviewItem when the wizard's tile creation is complete. But this wasn't updating the fieldtype and sparkline information leading to misconfiguration and a blank tile value. A work around was to open the tile's settings menu and close it which would cause its configChanged() to be called and the correct fieldtype/sparkline values to be calculated.

Additionally I have added metadata tile support to display Time and Date fields, so Start_Time & Start_Date fields now display correctly.

image     image

### DIFF
--- a/src/Charts/OverviewItems.cpp
+++ b/src/Charts/OverviewItems.cpp
@@ -1734,9 +1734,6 @@ TopNOverviewItem::setDateRange(DateRange dr)
 void
 MetaOverviewItem::setData(RideItem *item)
 {
-    // Ensure fieldtype & sparkline match the metadata symbol
-    configChanged(0);
-
     if (item == NULL || item->ride() == NULL) return;
 
     // non-numeric META
--- a/src/Charts/OverviewItems.cpp
+++ b/src/Charts/OverviewItems.cpp
@@ -1734,6 +1734,9 @@ TopNOverviewItem::setDateRange(DateRange dr)
 void
 MetaOverviewItem::setData(RideItem *item)
 {
+    // Ensure fieldtype & sparkline match the metadata symbol
+    configChanged(0);
+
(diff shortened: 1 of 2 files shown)


amtriathlon:
> When metadata tiles are initially created they are blank, and only update when the config menu is opened and closed.
> 
> The reason for this is the AddTileWizard initially creates a default MetaOverviewItem which has fieldtype=-1, then it calls setData on the MetaOverviewItem when the wizard's tile creation is complete. But this wasn't updating the fieldtype and sparkline information leading to misconfiguration and a blank tile value. A work around was to open the tile's settings menu and close it which would cause its configChanged() to be called and the correct fieldtype/sparkline values to be calculated.

I think it would be better to fix this in AddTileWizard since it may affect other tiles and there is no need to call configChanged on each setData which happens each time an activity is selected in the side bar, something like this should work:

```
diff --git a/src/Gui/AddTileWizard.cpp b/src/Gui/AddTileWizard.cpp
index 8a9ede630..2137cca73 100644
--- a/src/Gui/AddTileWizard.cpp
+++ b/src/Gui/AddTileWizard.cpp
@@ -206,6 +206,7 @@ AddTileFinal::validatePage()
             // everthing else is a bit smaller
             wizard->item->parent->addItem(0,0,1,7, wizard->item);
         }
+        wizard->item->configChanged(0);
         wizard->item->parent->updateGeometry();
     }
     return true;
```

paulj49457:
Thank you for improved method to fix the blank tile, PR updated as suggested.

PR #4466, opened 2024-04-05
Simplified Equipment feature (for feedback)

paulj49457:
This PR adds an equipment maintenance feature to GC, any feedback or guidance on future development would be useful.

**Design Aims:**

• To be as generic as possible, support multiple equipment & components
• To minimise the number of Metadata items added to GC activities
• Minimise impact to existing GC code.

**Description**
This PR will create a new unpopulated “EquipmentLink” metadata item in all activities to provide a mechanism to link an activity to an equipment link item defined in the new page. The diagram below tries to illustrate the relationships between the equipment items 
 
image

Within the equipment pages, various user defined components can be attached to the equipment link with time-spans of use, allowing the calculation of distances covered by individual components. The equipment page contains an equipment List:

image

**Popup Menus**

The Equipment List menu allows:
•	Distance Recalculation
•	Add Equipment Link
•	Load/Save Equipment
•	Expand/Collapse All

The Equipment List context menu (right click on an equipment item) allows:
•	Add Shared
•	Add Distance
•	Add Time
•	Add Banner
•	Delete* 
•	Top**
•	Shift Up/Down**
•	Bottom**
•	Shift Left**

Note:
*: Items with children cannot be deleted, move their children (drag & drop them), or delete their children.
**: These options are available when relevant.

**Equipment List –> Shared Equipment**
  
image

image
 
Items which are moved or shared across equipment (wheels, etc) should be created in this section, and then the items dragged to the relevant Equipment link headers. This will cause a reference to be created that will accumulate any distance and added to the shared equipment item.
{} – indicates the number of references to this object
[] – indicates the total mileage of the equipment (from its references)

**_Shared Distance Equipment Page:_**
•	Description: Equipment description
•	nonGCdistance: allows an initial value to be set for activities not recorded within GC
•	gcDistance: the distance contribution of all the relevant GC activities
•	Total distance: the total of nonGCdistance + gcDistance
•	Replacement Distance: Sets the replacement distance warning, set a value of zero to disable warnings.
 
image

•	Note: Save must be pressed to store any text/date changes

**Equipment List –> Equipment Links**
 
image

Equipment links “connect” to the activities.
All equipment items within an equipment hierarchy will accumulate the equipment links distance, allowing for any time restrictions.
() – indicates the number of “connected” activities
[] – indicates the total mileage of the “connected” activities.

**_Equipment Link Page:_**
•	Equipment Link: text that matches the Activities EquipmentLink Metadata item.
•	Activities: the number of “connected” activities.
[...]

PR #4519, opened 2024-07-05
Editable metadata tile

paulj49457:
This PR provides a quick way to edit already defined metadata values from an overview tile. The popup menu is activated by right clicking on a metadata tile, if it has pre-configured values then these will be available for selection via a popup menu. I find this feature useful for correcting the Sport and setting equipment used for the activity (EquipmentLink from #4466).

Screenshot 2024-07-03 073205

Note: Git diff is making the change to src/Gui/ChartSpace.cpp look far more complex than it is, I have just added an if statement for left & right mouse buttons.


amtriathlon:
Since right click is not used in this chart and the RPE tile already uses the left button, I think it would be more intuitive to use the left click over the text, to popup the list if values are constraint or -perhaps in the future- edit the value, but I will defer this to @liversedge who surely has a better idea about how to do this.

paulj49457:
FYI, The mechanism for this popup is included in https://github.com/GoldenCheetah/GoldenCheetah/pull/4533 to support the Tile move/clone functions within the proposed equipment pages.

lennartalff:
To add my two cents regarding the left/right mouse button question: I think this PR differs in its behaviour from the RPE tile. The RPE tile has a bar, where I click on the position of my chosen RPE value. So for this PR's feature I would consider a right click for opening the context menu and a subsequent left click to choose the appropriate option might be the equivalent behaviour. But besides that no strong opinion on this question.

For me this feature would be really useful. It would allow me to set the workout code and one or two other fields without leaving the overview dashboard without switching to the Data tab. So it gives me a more streamlined user experience.

Thanks @paulj49457 for the effort :blush:

paulj49457:
Hi @lennartalff  I decided on the right button as a menu is popped up with the options, and this follows many other gui interfaces. I agree its different to the RPE tile, but on that tile you are directly selecting the item so the left button makes sense. Additionally the left button is used for tile resize and drag operations so it could get a bit overloaded, but if the general feedback is that it should use the left mouse button I'll change it.

lennartalff:
I'm sorry if could not make my point clear: I meant to express that I agree with your choice of button assignment as you implemented it. 

Just wantes to argue for the fact that even though the RPE tile is different it follows actually the same principle. Maybe I got a bit tangled up in my elaboration :joy:

amtriathlon:
I think the right button is not being used in Overview charts/tiles for a reason, but I agree some visual cue/hotspot should be present for left click, perhaps a small "edit" icon near the cog icon, and it should cover all metadata type cases.

paulj49457:
The latest commit contains the updated metadate popup menu:

image

paulj49457:
@amtriathlon I think this is ready now for review. I'm sorry, I rebased this PR upon the latest GC commit, instead I should have merged, all the history of the commits has been retained. 

I noticed that the configure icon on the tiles accepts either the left or the right mouse as input, so the new edit icon follows this approach. The edit icon is only displayed for those metadata tiles that have auto complete values defined.

amtriathlon:
Hi Paul, it seems to be Ok for the case where the field has a list of values, but not when it is "*" which is interpreted as "suggest previous values" as explained in https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Preferences_Data-Fields.
The completer used by the Data chart covers both cases: https://github.com/GoldenCheetah/GoldenCheetah/blob/b62a2d31903af8ecd1338ecb01480ccb69ed462a/src/Metrics/RideMetadata.cpp#L1605
Besides that I think this could be much more useful if it worked for the general case, avoiding the need to go to Data tab for normal use, and you are close to that, IMO.
We would need a small popup dialog to edit the value, with a field like you implemented for the batch processing metadata update, and to replace the hamburger menu icon by a more standard edit icon since it will not popup a menu in the general case.

paulj49457:
Hi @amtriathlon, Thank you for the suggestion, I had been thinking along similar lines regarding a popup dialogue for entering notes, as notes and a couple of quick metadata edits are what I need to complete on a new activity. But I'll look at implementing a more general solution so that the user can just stay on the overview page for the majority of activity data edits. Can you post or reference an existing GC icon you'd like to see on the tile?

amtriathlon:
> Hi @amtriathlon, Thank you for the suggestion, I had been thinking along similar lines regarding a popup dialogue for entering notes, as notes and a couple of quick metadata edits are what I need to complete on a new activity.

That's my use case too, and I think it is pretty common.

> But I'll look at implementing a more general solution so that the user can just stay on the overview page for the majority of activity data edits. 

Great!

>Can you post or reference an existing GC icon you'd like to see on the tile?

I think there is none, but some of these could match the configure icon:

edit_pencil_write_pen_draw_icon

edit_underline_pencil_write_writing_icon

Here are them in png format, 512x512 with transparent background: [edit_icons.zip](https://github.com/user-attachments/files/17308995/edit_icons.zip)

paulj49457:
@amtriathlon I have updated this PR to use popup a dialog to allow editing of the metadata items, but I'm now starting to think that this additional field could just be added to the existing tile item configuration menu, so I'll create a separate PR with that option to allow the two approaches to be compared.

src/Gui/MetadataDialog.cpp
```diff
@@ -0,0 +1,230 @@
...
+    switch (field_.type) {
+    case FIELD_TEXT:
+    case FIELD_SHORTTEXT: text = ((QLineEdit*)metaEdit_)->text(); break;
+    case FIELD_TEXTBOX: text = ((GTextEdit*)metaEdit_)->document()->toPlainText(); break;
+    case FIELD_INTEGER: text = QString("%1").arg(((QSpinBox*)metaEdit_)->value()); break;
+    case FIELD_DOUBLE: text = QString("%1").arg(((QDoubleSpinBox*)metaEdit_)->value()); break;
+    case FIELD_CHECKBOX: text = ((QCheckBox*)metaEdit_)->isChecked() ? "1" : "0"; break;
+    case FIELD_DATE: {
+        if (field_.name == "Start Date") {
+            text = QDate(1900, 1, 1).daysTo(((QDateEdit*)metaEdit_)->date());
```
amtriathlon, line 175: This doesn't seems to make sense, are you trying to assign a number to a QString?

src/Gui/MetadataDialog.cpp
```diff
@@ -0,0 +1,230 @@
...
+        } else {
+            if (((QDateEdit*)metaEdit_)->date().isValid()) {
+                text = ((QDateEdit*)metaEdit_)->date().toString("dd/MM/yyyy");
+            }
+        } } break;
+    case FIELD_TIME: {
+        if (field_.name == "Start Time") {
+            text = QTime(0, 0, 0).secsTo(((QTimeEdit*)metaEdit_)->time());
```
amtriathlon, line 183: This doesn't seems to make sense, are you trying to assign a number to a QString?

amtriathlon:
> @amtriathlon I have updated this PR to use popup a dialog to allow editing of the metadata items

Thank you, I found a couple of errors which prevent compilation in my local build -in fact I don't understand why it passed ci-, but after commenting out that it looks pretty good, I will be testing this the next week, thank you!

paulj49457:
@amtriathlon Please take a look at https://github.com/GoldenCheetah/GoldenCheetah/pull/4557 I think this might be a better solution

src/Gui/MetadataDialog.cpp
paulj49457: Sorry, it's a mistake, I'll check my other PR.

amtriathlon:
I did, I think this is a better option, more aligned with how metadata works in GC.

amtriathlon:
Hi Paul, at least when building using Qt6 on Windows the metadata dialog positioning is problematic for tiles on the borders:

image

Config dialog positioned at the center works ok.

thejockl:
Could this pattern also be used to trigger other actions directly from a tile? I think about the Athletes View and functionalities like backup or delete user

paulj49457:
@amtriathlon Updated to ensure the metadata dialog remains within the GC window.

amtriathlon:
Hi Paul, it seems to be ok now, it looks like it is unnecessary to special case Start Date/Time in Metadata dialog since edition is not allowed anyway.

amtriathlon:
> Could this pattern also be used to trigger other actions directly from a tile? I think about the Athletes View and functionalities like backup or delete user

Yes, likely it can be generalized so Ivón and action can be provided by subclasses.

Also this could be extended to Metric tiles to allow metric override on Overview charts.

amtriathlon:
I will merge this for better testing, thank you!

PR #4522, opened 2024-07-15
Added support for the TrainerDay workouts/find API

thejockl:
* Added a new (optional) tab to the TrainerDay workouts download dialog
* Deferring loading of the classic list of workouts until this tab is activated
* Implemented the API for TrainerDays /workouts/find (see https://api.trainerday.com/api-explorer/)
* Visualizing the workouts using a colored ErgFilePlot
* Modified ErgFilePlot::setData to directly use the data of the given ErgFile instead of falling back to the ErgFile set in the context (required to display multiple plots at once)
* Added a simplified version of the existing workouts query sytnax (supported verbs: duration, dominantzone)
* Saving the workouts as trainerday-<hash>.erg
* Using the segments-section as input for the hash and to identify duplicates
* Functionality can be enabled in gcconfig.pri (GC_WANT_TRAINERDAY_API)
* Prepared the travis-scripts before_script.sh to inject the API key to Secrets.h
* Prepared appveyor.yml to inject the API key to Secrets.h (duplicated the secret of GC_AZUM_CLIENT_SECRET since I can't encrypt a key myself)
* To enable enable the functionality, an API key from https://trainerday.com/api is required

Screenshot_20240715_231935


amtriathlon:
Hi, Joachim, if you email me the plain API key I can return the encrypted version for AppVeyor and Travis-ci

thejockl:
Hi @amtriathlon, I requested a production API key from TrainerDay and will forward it to as soon as I received it (currently I only have a stage-key)

amtriathlon:
I will merge this so it can be tested by windows users since PR builds can’t use encrypted keys.

peret2000:
Hello, I found a compilation error on Ubuntu, in [TrainerDayAPIDialog.cpp](https://github.com/GoldenCheetah/GoldenCheetah/blob/90cb68623f3ff63e1d750affd54b017f34ed408a/src/Train/TrainerDayAPIDialog.cpp#L150): GC_TRAINERDAY_API_PAGESIZE is not defined. In gcconfig.pri.in its definition is commented.
Maybe that is not the proposal, or line [140](https://github.com/GoldenCheetah/GoldenCheetah/blob/90cb68623f3ff63e1d750affd54b017f34ed408a/src/Train/TrainerDayAPIDialog.cpp#L140) in TrainerDayAPIDialog.cpp is intended to be uncommented...
Can you check, in case I am wrong?

thejockl:
Hi @peret2000, GC_TRAINERDAY_API_PAGESIZE must be defined to allow compilation. Probably 25 is the value it should always be set to - but let me go back to Alex from TrainerDay for clarification. I would very much welcome to remove it but added it as I got the explicit statement that my stage-key was configured to deliver 25 results (do other keys deliver a different pagesize?)

peret2000:
Ok, I understand. Should it be uncommented in gcconfig.pri.in?

thejockl:
Lets await Alex answer, but most probably I will add something like
```
#if !defined(GC_TRAINERDAY_API_PAGESIZE)
#define GC_TRAINERDAY_API_PAGESIZE 25
#endif
```
to TrainerDayAPIDialog.cpp, still allowing an override but providing a sane default without configuration

Issue #4512, opened 2024-06-13
Custom relative Date Ranges in Trends


amtriathlon:
The ability to create custom relative date ranges in Trends, s.t. Last 24 months, is an old request at the users forum, see https://groups.google.com/g/golden-cheetah-users/c/nBR4OgVfmS4/m/kvCHmuk7AwAJ and https://groups.google.com/g/golden-cheetah-users/c/TVgXOYRCoUw/m/w_YZ6213CAAJ, for example.
Also a Year to Date range type was proposed to compare season advance across years (https://groups.google.com/g/golden-cheetah-users/c/CW-Xqvsc3Sg/m/LTk1IcoGBAAJ), there is a proposed implementation for this in https://github.com/GoldenCheetah/GoldenCheetah/pull/4439
I think the above is too specific but the ability to create a Relative Date Range covering both cases could be a nice addition.

PR #4534, opened 2024-08-23
Implemented user defined relative date ranges

thejockl:
See #4512
* Added additional time range options for user defined seasons
  * Relative start (e.g. 3 months ago)
  * Duration after start / Duration before end (e.g. 2 months)
  * Year to Date end, defining the end as the closest day with todays day and month after start
* Refactored Season.h/cpp and SeasonParser.h/cpp for better testability
  * Kept Season, Phase, SeasonEvent, SeasonOffset, SeasonLength in Season.h/cpp
  * Moved all UI related classes to SeasonDialogs.h/cpp
  * Renamed SeasonParser.h/cpp to Seasons.h/cpp, moved class Seasons there
* Replaced deprecated Qt classes
  * Replaced QRegExp by QRegularExpression in SeasonParser
  * Rewrote parsing seasons.xml to use QXmlStreamReader instead of QXmlSimpleReader
* Unittesting
  * Added a simple way to implement and execute (make check) unittests in Golden Cheetahs source tree based on Qt Test (https://doc.qt.io/qt-6/qttest-index.html), tested on Linux and Windows
  * Implemented unittests for classes (mostly before refactoring)
    * Season
    * SeasonParser
    * SeasonOffset


paulj49457:
I have a trends graph that requires all of my GC data, so is it possible to set a "season" date range from say 25/12/2010 to today ?

thejockl:
Yes, this is possible with absolute start and relative end:
image

paulj49457:
> Yes, this is possible with absolute start and relative end: image

Great, that will be very useful, thank you.

lennartalff:
Works like a charm for me. Greatly appreciated from my side :partying_face: . 

Would you consider extending this to phases as well? As I observed, for relative seasons the phases/events are disabled. Why is that so?

thejockl:
As phases and events are nested within seasons, I think they only make sense for fixed seasons.

What I found in the Wiki (https://github.com/GoldenCheetah/GoldenCheetah/wiki/FAQ-SEASONS-AND-EVENTS) is:
```
What is event used for?
It is only used to annotate a LTM chart -- so you can show an 'event' that happened at a particular date. There are no rules about what constitutes and event. It could be a race, a training camp or just a day that was particularly meaningful e.g. illness, crash. If you have a calendar service configured, new events will be uploaded to it.
```
So I see Phases as fixed sections within a Season to help preparing for Events - and as such they shouldn't move relative to other dates.
What are your usecases for relative Phases?

Thanks for your comment, this is exactly the type of discussion I want to have before features gets merged :)

lennartalff:
I would agree that it might be not that useful to have Events/Phases defined with relative dates.  But I would not immdediately see why I could not want to add an absolute event inside a (maybe partly) relative season. If this has some serious technical implications on how the ranges are handled or it simply increases the complexity without adding a significant benefit (I do not have an explicit use case in mind, where I would "insist" on this feature) it is perfectly fine to keep it the way it is currently. 

But also as part of the changes it is now conveniently possible to define the range with and absolute offset to the absolute end for example. Nothing you couldn't to with the old season dialog of course and I wouldn't consider selecting both dates in the date/calendar picker as too much of a burden. So as I said a mere convenience.

**Long story short**: I personally am quite happy with your current implementation and would be quite content to see it merged as it is.

amtriathlon:
Great job, thank you!

PR #4536, opened 2024-08-28
Trends View: Following the filter in compare mode

thejockl:
Before this PR: After adding seasons to compare pane in Trends View,, updating the filter is not affecting the activities selected by these seasons. To have only the filtered activities of the seasons
* the filter has to be set
* all seasons must be removed from compare pane
* all seasons must be re-added to the compare pane

With this PR: Modifying the filter is reflected immediately in the seasons available in the compare pane, similar to non-compare mode

Additionally included a fix for CPPlot that made Golden Cheetah crash if no data was available


amtriathlon:
Hi Joachim, current behavior is a feature in my book: it allows to compare filtered set of activities where the filter can be different for each date range, different equipment is one of the use cases.
The fact each compare item has its own filter is also an indicator this is by design, if the idea is to use context filters this would be redundant.
I understand the convenience in the use case you are thinking about, but there others and date ranges could even belong to different athletes and I am not sure what changing the filter for one of them should do for the others.
I am not willing to merge this, sorry.

Issue #1381, opened 2015-05-27
Auto CP


liversedge:
A tool to set CP config based upon model estimates.
Maybe check after import or model refreshes too ?

antoniusriha:
I like this idea. I think however, that date range of the the model feed data must be configurable. Currently it's set to the rolling bests of the last 12 weeks. This seems to be a rather long period in my opinion.
Another thing I'd like to suggest, is to add semi-automatic power zones management: The user gets to see a list of all model estimates and can then choose which to include in the config.

PR #4543, opened 2024-09-15
Semi automatic creation of ranges for power zones

thejockl:
This PR tries to be part of a solution to #1381 by adding semi automatic creation of ranges for power zones.

* Changed the UI of CPPage to inline-editing of all values in all tables
* Added a sports-specific selector for the model (cp2, cp3, ext, manual)
* Allowed to create new ranges either manually or based on the estimated values of the model
* Added option to reset each ranges values to those of the selected model
* Added message to create a new range if todays estimate differs from those of the currently active one

zones

Open issues of this PR include adapting the other tables in the athlete settings to the same inline editing mechanisms

Feedback is highly welcome


lennartalff:
Looks great :blush: . Really like the UI overhaul. To me it feels very handy to be able edit the values simply in-line. 

But I'm missing the "Closest Estimate" column shown in your GIF.

Currently the tooltip for the model fit only says "Estimate and settings **differ**". Maybe this could be changed to indicate the difference, e.g. "The set CP differs by +-NN Watts" or something similar. Otherwise, the only way to see the difference is to adopt the model and keep the old value in mind and compare it with the new one. And if I do not want to apply it after inspecting the change, I have to close the dialog without saving. 

Currently the values are checked for exact equality. What do you think about adding some tolerance there? Not sure about that myself, just something to consider.

thejockl:
I moved the "closest estimate" to the "Model Fit"-field as a second icon (clock) that is only shown if no estimate is available. The distance in days to this estimate is shown in the tooltip.
image

Background is that even I as the author (who really should know better) got distracted by "exact match". As I consider this the least important information, I moved it.
Both your points are great, I will add tolerance and think about the adoption process. Maybe as you proposed in the tooltip (but that would probably grow too big to look nice) or an additional dialog that shows all values - current and estimate. As an addition it could become possible to choose which values to take over.

thejockl:
Hi @lennartalff, this is my first proposal how the values can be verified before adoption (the input fields are read only):
image
The code needs some refinement before pushing, but it shows my current direction. PMax has no accept checkbox since it didn't change

lennartalff:
Look great. Especially the comparison symbols to immediately grasp the direction of change.
Also the option to accept only some values seems quite useful to have more control.

Maybe highlighting the respective values in the "current" and "estimate" column depending on acceptance? Or rather grey out the opposite value as the PMax. But that's probably more of a gimmick.

Nice work! :+1:

thejockl:
From my view this PR is ready for integration, however I am still playing around with the layout of the adoption dialog, showing only one number for unchanged values and disabling only the selected value for changed entries (currently the whole line is shown as disabled

Status Quo:
image

Might be:
image

Are there any opinions on which way to go?

thejockl:
Went for this option:
image

amtriathlon:
Hi Joachim, I like the UX modernization, great improvement!
WRT the original issue I always thought it to open a can of worms from a support POV, users tends to have a hard time understanding how model estimates work and why sometimes results are not what they expect, CP chart is way superior for setting zones, IMHO.
Having said that, the semi-automatic option could be useful for some users and I am willing to merge this, but the default option should be manual, for the reasons above.

thejockl:
Thanks for the feedback, will change the default later today. If fine for you, i will create two more PRs to
- align the UX in the measure page, also adding inline editing
- sort the entries in reverse to show the most recent ranges on top, preventing the need to scroll to bottom

amtriathlon:
Thank you!

liversedge:
At the very least the "Adopt" button should be "Adpopt..." to indicate there is a dialog- it won't just update on the spot. 

It is not clear at all what the differences are when on the main zone dialog- it really should tell you what the differences are since zones are carefully curated and no-one is going to blindly click on "Adopt" without knowing what they are adopting !

amtriathlon:
Perhaps Review would be better than Adopt for the main dialog

thejockl:
Will look into this after the weekend and create a followup PR

PR #4544, opened 2024-09-21
Fixed translated image filters

thejockl:
* Some translations (de, sv) for filters used in QFileDialogs contained commas to separate wildcards. As this is in violation with Qts syntax, the dialogs didn't show any files. These commas are removed by this PR
* No other translation contains these commas


thejockl:
TBH, I am not too happy about my PR as I think only the name (e.g. Images, Bilder, ...) of the filter should be translated, not the wildcards by themselves. My preferred solution would be something like
```
tr("Images") + " (*.png *.bmp *.jpg)"
```
Are there any opinions on that? If you agree, I would create this solution (including updating the existing translations)

amtriathlon:
Good catch and, yes, I agree with your last comment, it would be a more robust solution. Thank you!

thejockl:
Created #4545 as replacement for this PR

PR #4545, opened 2024-09-21
Removed wildcard from translatable part

thejockl:
Some translations (de, sv) for filters used in QFileDialgos contained commas to separate wildcards. As this is in violation with Qts syntax, the dialogs didn't show any files.

The change separates the wildcards from the name of the filter so that only the name can be manipulated by translation. All affected translations are updated.

This PR replaces #4544


amtriathlon:
Thank you!

PR #4550, opened 2024-10-10
Respect the Line Width Setting Inside the Workout Editor's Realtime Plot

lennartalff:
### What changed?
The realtime plot of the workout editor now applies the line width setting of the app. You can observe this in the screenshot below, the axis lines etc. have the hard coded with of 1 in comparison to the line width of 3 in my app's settings for the realtime data.

### Why?

- The hard coded line width of 1 might be to small for large displays with a high resolution (like in my case)
- I think it is appropriate for the workout editor plot to respect the app's settings.

### Alternative
Have a dedicated line width setting for the workout editor. I do not see the use case of this option.

### DIFF
--- a/src/Train/WorkoutWidgetItems.h
+++ b/src/Train/WorkoutWidgetItems.h
@@ -220,7 +220,8 @@ class WWTelemetry : public WorkoutWidgetItem {
                              const WorkoutWidget::WwSeriesType& seriesType)
         {
             QPen linePen(color);
-            linePen.setWidth(1);
+            double width = appsettings->value(workoutWidget(), GC_LINEWIDTH, 1.0).toDouble();
+            linePen.setWidth(width);
             painter->setPen(linePen);
 
             int index = 0;


amtriathlon:
Thank you!

liversedge:
The width should be scaled via dpiXFactor

lennartalff:
@liversedge instead of using the line width setting or in addition?

amtriathlon:
All other places where GC_LINEWIDTH is used are not scaled either, so this is consistent with existing code. Agree fixed values should be scaled.

liversedge:
Ah- thats probably wrong, but lets not break things changing it- apologies its been a while since I played with the older charts !

amtriathlon:
I think to scale GC_LINEWITH could make initial setup easier for new users on large displays, but we should do that consistently to avoid problems, and think what to do with existing setups. Perhaps an easier alternative would be to let them unscaled, but to scale default values on each use.

PR #4551, opened 2024-10-10
Sort the Dial Window's Combobox by Name

lennartalff:
### What changed?
The combobox for the data series selection of the training view's dial window is now sorted by name.
I checked the sorting for english and german and it looks correct to me.
### Why
There is quite an amount of data series to select from in the combobox. This makes finding the and selecting the desired data series quite tedious. If it is sorted by name, the desired series is much easier to find.
### Alternative
 For example group the items by their categories, like power related data, timing related data, etc. But I think these groups might quickly become non mutually exclusive so the group assignment gets ambiguous. Additionally, the QComboBox widget does not support grouping items. So more extensive changes would be needed. Probably not worth the effort.

### DIFF
--- a/src/Train/DialWindow.cpp
+++ b/src/Train/DialWindow.cpp
@@ -49,7 +49,15 @@ DialWindow::DialWindow(Context *context) :
     QLabel *seriesLabel = new QLabel(tr("Data Series"), this);
     seriesLabel->setAutoFillBackground(true);
     seriesSelector = new QComboBox(this);
-    foreach (RealtimeData::DataSeries x, RealtimeData::listDataSeries()) {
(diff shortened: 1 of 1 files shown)


amtriathlon:
Why just don’t change the QComboBox insertPolicy property?

lennartalff:
Because I did'nt know about this option. But according to the [documentation](https://doc.qt.io/qt-6/qcombobox.html#insertPolicy-prop) the policy is applied to user entered items.
A user in the [Qt Forum](https://forum.qt.io/post/295706) confirms that it does not apply to programatically inserted items and suggests to sort the list item's to be inserted before adding them. I just tested it myself and can also confirm that the `insertPolicy` does have no effect for the order of the added item's.

amtriathlon:
Thank you!

PR #4553, opened 2024-10-12
Make the Workout Editor's Settings Persistent

lennartalff:
### What changed?
The chart settings of the workout editor are now defined as `Q_PROPERTY`.
This way the settings are stored inside the perspective's layout xml and are restored after the app is restarted. Otherwise all settings get lost each time GC is closed.


amtriathlon:
Thank you!

PR #4554, opened 2024-10-12
Fix regression caused by #4551

lennartalff:
Starting from #4551 the DialWindow's `dataSeries` property does not match the ``RealtimeData::Dataseries`` enum values.

The enum value has been stored as userdata for each dataseries item of the ``seriesSelector`` combobox. But the getter/setter for the `dataSeries` has stored/restored the currently/previously selected series' index. This has become a problem since #4551 as the data series are added in another order.

I also assume that #4337 might cause problems for some users, since new data series enum values have been inserted and all following data series enum values have been changed. Hence, I moved them to the end of the enum.

I also checked the default train perspectives and fixed all occurences, where the DialWindows' title did not match the corresponding data series.

Edit: Note that this will **not** solve the problems for layouts that have been created before #4551, if the `dataSeries` property had differed from the data series enum value.

Looking at https://github.com/GoldenCheetah/GoldenCheetah/blob/60f9033617996c30a7b18a608adea27e8e370153/src/Train/RealtimeData.cpp#L533
I can observe that not all data series are enlisted. I do not know if this is on purpose or by accident. Anyhow, this PR should make the layouts more robust, since regardless of the data series subset the function returns, each will be correctly associated with its enum value by the `DialWindow`.

### DIFF
--- a/src/Resources/xml/train-perspectives.xml
+++ b/src/Resources/xml/train-perspectives.xml
@@ -36,7 +36,7 @@
 		<property name="showInstant" type="bool" value="1" />
 		<property name="avgsecs" type="int" value="1" />
 		<property name="avgType" type="int" value="32695" />
-		<property name="dataSeries" type="int" value="12" />
+		<property name="dataSeries" type="int" value="16" />
 		<property name="style" type="int" value="2" />
 	</chart>
 	<chart id="22" name="" title="W&apos; bal" >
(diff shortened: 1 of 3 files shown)


amtriathlon:
Hi Lennart, thank you for your analysis, I will merge the commit to fix #4337 regression.
WRT #4551 since it affect existing users and it is a mostly cosmetic change, I prevert to revert that.

lennartalff:
I'm fine with that. I would only suggest to at least add a comment in thr code to remind developers that changing the order of the listdataseries function breaks user condigurations. I feel like this is not immediately obvious at first glance. 

I'm sorry that my previously merged PR caused trouble.

PR #4557, opened 2024-10-13
Config metadata editing (a better option)

paulj49457:
The following PR includes an editable field within the existing tile configuration dialog, I think this is a better solution than having an additional dialog in PR https://github.com/GoldenCheetah/GoldenCheetah/pull/4519, it's a simpler implementation 

image

This PR and the alternative PR both fix the following metadata update issues detected whilst testing the dialogs:

- When viewing the details page with  TIME_FIELD & DATE_FIELDs, if they were not set for an activity they didn't clear the values from the previous activity. (RideMetadata.cpp)
- The date fields are now consistently displayed using dd/MM/yyyy format, previously the Start_Date was display this way, but the normal DATE_FIELD used dd.MM.YYYY format. (RideMetadata.cpp)
- Changing metadata data items on the details page didn't update the tiles displaying the same metadata fields, it required the config dialog to be opened & closed. (OverviewItems.[h|cpp] see rideMetadataChanged() & metadataChanged;
- Prevents the interval notes & goal from being edited as the tile cannot determine which interval should be updated, the batch processing option for these fields has also been removed.
- The configuration dialog is now positioned close to the config button from which is was launched.


amtriathlon:
Hi Paul, I really dislike this option, I think to conflate chart config and metadata update is a bad idea and may confuse users. Sorry.

PR #4558, opened 2024-10-14
Position tile config menu

paulj49457:
This PR matches the functionality in the Editable metadata tile https://github.com/GoldenCheetah/GoldenCheetah/pull/4519 to position the popup menu close to the location from which it was launched.

### DIFF
--- a/src/Charts/Overview.cpp
+++ b/src/Charts/Overview.cpp
@@ -178,8 +178,7 @@ nodice:
 void
 OverviewWindow::configItem(ChartSpaceItem *item, QPoint pos)
 {
-    OverviewConfigDialog *p = new OverviewConfigDialog(item);
-    p->move(pos.x()+10, pos.y()+10);
+    OverviewConfigDialog *p = new OverviewConfigDialog(item, pos);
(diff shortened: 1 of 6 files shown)


amtriathlon:
This change looks problematic on Windows since the dialog gets positioned mostly out of screen for tiles near the borders:

image

I don't think this is a good idea.

paulj49457:
@amtriathlon Please see updated version which ensures the dialog remains within the GC window.

amtriathlon:
Yes, Paul, it seems to work on Windows now, I have not test on Linux or macOS.
My concern is: while I understand it may be convenient to reduce mouse movement, I am not sure we should micro-manage this instead of just let the Windows Managers do their job. 
A priori I am not willing to merge this, but If others better versed than me on UX have an opinion, I would like to read it.

peret2000:
On Ubuntu 24.10 it works well. Menu is always visible and close to the tool icon.

amtriathlon:
I will merge this so it can be tested together with https://github.com/GoldenCheetah/GoldenCheetah/pull/4519, we can revert if necessary. Thanks Paul.

PR #4559, opened 2024-10-21
Editable metadata tile fixes expanding buttons and set minimum field …

paulj49457:
Fixed expanding cancel and ok butttons:

image

Set minimum width for the different fields types:

image

image


amtriathlon:
Thank you!

PR #4561, opened 2024-10-23
Zones: Sorting ranges in reverse

thejockl:
Power, Heartrate, Pace: Showing the most recent range at the top instead of at the bottom by adding a hidden column _rnum to the TreeViews


amtriathlon:
Thank you!

PR #4562, opened 2024-10-24
Harmonize the layout of athlete settings

thejockl:
Settings
* About
  * Show the athletes name
  * Mask the avatar-image as a circle
* Model
  * Add button to revert to defaults
* Auto Import
  * Use inline editing
* Backup
  * Add a "Backup now"-button

about-fill
about-narrow

Create new athlete
* Rename NewCyclistDialog to NewAthleteWizard and turned it into a 2-page wizard
* ChooseCyclistDialog: Open the NewAthleteWizard without interaction if no athlete is available

newathlete

Common
* Avatar image
  * Use Qt::SmoothTransformation for improved quality
  * Crop to the center of the original image
  * Keep aspect ratio of the original image
* All forms: Use QFormLayout to simplify the code and provide a desktop aware layout
* Switch to QDialogButtonBox for a button order aligned with the desktop


amtriathlon:
Hi Joaquim, the fill version looks weird for me:

> about-fill

I prefer the narrow version:

> about-narrow

but perhaps is a familiarity issue, may be others have an opinion.

thejockl:
Hi Ale, this comes from QFormLayout and tries to be compliant with the OS, based on the used QStyle. See https://doc.qt.io/qt-6/qformlayout.html:
image

If you prefer using the narrow version, I could set the style manually: https://doc.qt.io/qt-6/qformlayout.html#fieldGrowthPolicy-prop and https://doc.qt.io/qt-6/qformlayout.html#labelAlignment-prop

amtriathlon:
Thanks Joachim, I think the QMacStyle looks better, specially when the labels and size hints are wildly different for different fields, but the idea to use OS style defaults is sound too. May be the style could be a global option in Appearance config at some point.
Do you have a preference @liversedge?

thejockl:
The global config option sounds good to me and could be easily implemented by a common function "newQFormLayout()" which applies all the settings

thejockl:
Here is my proposal for the option, added as a new checkbox to the appearance tab:
image

To make development easier, this page was changed to have two QFormLayouts (left and right).

Topics I am unsure with:
1. The phrase "System styled Forms" doesn't sound self explanatory to me. Are there any better proposals?
2. The default currently is to use system style, are there any arguments to change this to the custom style?
3. On Mac the option is hidden. Probably it would be better to hide it on all platforms where the style in use provides the same look?

amtriathlon:
> Here is my proposal for the option, added as a new checkbox to the appearance tab: image
> 
> To make development easier, this page was changed to have two QFormLayouts (left and right).

It looks ok for me, my 2 cents below:

> Topics I am unsure with:
> 
> 1. The phrase "System styled Forms" doesn't sound self explanatory to me. Are there any better proposals?

I think "Mac styled Forms" would be easier to understand

> 2. The default currently is to use system style, are there any arguments to change this to the custom style?

I would try the Mac style, we can change default later if it is unpopular.

> 3. On Mac the option is hidden. Probably it would be better to hide it on all platforms where the style in use provides the same look?

With the above name on Mac it could be disabled or hidden, as you prefer.

amtriathlon:
Thank you!

marcenGC:
Hi,

I gave it a try under Arch Linux with KDE as QT 6 build. I'm not sure if it is a build error or the symbols are missing. In some tabs, there is the "+" and "-" symbol missing. 
Bildschirmfoto_20241101_100622

Current status
https://raw.githubusercontent.com/GoldenCheetah/GoldenCheetah/master/doc/wiki/PreferencesAthlete_Power_Overview.jpg

thejockl:
Ale, thanks for merging! This was (for now) my last planned PR related to the athlete-settings

PR #4564, opened 2024-10-31
Infowidget: Made more strings translatable

thejockl:
Wrapped more strings with the tr()-function (see https://groups.google.com/g/golden-cheetah-users/c/8jMyCkrvms0/m/K1T9-BVABAAJ)


amtriathlon:
Thank you!

PR #4567, opened 2024-11-02
Settings: Inline-editing for Measures

thejockl:
* Implmented inline-editing in Measures-tab, similar to Zones (Power, HR, Pace)
* Disabled editing of columns Datetime, Source, Original-Source

image

### DIFF
--- a/src/Gui/AthletePages.cpp
+++ b/src/Gui/AthletePages.cpp
@@ -653,64 +653,18 @@ MeasuresPage::MeasuresPage(QWidget *parent, Context *context, MeasuresGroup *mea
     metricUnits = GlobalContext::context()->useMetricUnits;
     QList<double> unitsFactors = measuresGroup->getFieldUnitsFactors();
 
-    QVBoxLayout *all = new QVBoxLayout(this);
-    QGridLayout *measuresGrid = new QGridLayout;
-    Qt::Alignment alignment = Qt::AlignLeft|Qt::AlignVCenter;
-
-    QLabel* seperatorText = new QLabel(tr("Time dependent %1 measures").arg(measuresGroup->getName()));
-    all->addWidget(seperatorText);
-
     QString dateTimetext = tr("From Date - Time");
-    dateLabel = new QLabel(dateTimetext);
-    dateTimeEdit = new QDateTimeEdit;
-    dateTimeEdit->setDateTime(QDateTime::currentDateTime());
-    dateTimeEdit->setCalendarPopup(true);
-    dateTimeEdit->setDisplayFormat(tr("MMM d, yyyy - hh:mm:ss"));
-
-    measuresGrid->addWidget(dateLabel, 1, 0, alignment);
-    measuresGrid->addWidget(dateTimeEdit, 1, 1, alignment);
-
     QStringList fieldNames = measuresGroup->getFieldNames();
-    valuesLabel = QVector<QLabel*>(fieldNames.count());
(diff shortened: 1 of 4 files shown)


amtriathlon:
Hi Joaquim, it looks nice but it seems to be unusable when you have several years of daily data.

I had noticed athlete settings was slower after the Zones changes, but since zone tables are usually not to large it was usable.

Now after changing measures it takes several minutes to load and to switch between Body (12 years of daily data) and Hrv (7 years of daily data) when previous to this change it was instantaneous.

thejockl:
Hi Ale, that is a clear blocker, probably caused by the QStyledItemDelegates. I will check if my implementation can be optimized.

My main concern is the space occupied by the edit-fields on top, allowing only some few visible rows in the list. Also thinking about your impressive amount of data, would it be helpful having a "real" tree, grouping the entries by year?

For the editing, I could make it work similar to the definition of these fields in the options, using a "Edit"-Button:
image

amtriathlon:
> My main concern is the space occupied by the edit-fields on top

As you propose above, an Edit button (default for the selected item, and with double-click on an entry shortcut) would be a viable solution, you can re-use the Add dialog for that purpose. We use this pattern in other parts of the code.

> Also thinking about your impressive amount of data, would it be helpful having a "real" tree, grouping the entries by year?

Personally I would prefer to keep the list simple and efficient as it was before, it is not that common that I look at old data here, I use charts for that purpose.

thejockl:
It seems the issue is caused by my naive implementation of DoubleSpinBoxEditDelegate::sizeHint which is called very often and creates a dedicated QDoubleSpinBox every time. The latest commit precalculates and caches the sizeHint. This improved performance massively in my testdriver (QTreeWidget with 8 columns / 10000 rows). With this change, inline-editing hopefully can be kept for measures, creating a unified UX for the settings.
Ale, if the performance improvement is good enough for you, I will add the same pattern to all delegates in src/Gui/StyledItemDelegates.

amtriathlon:
> It seems the issue is caused by my naive implementation of DoubleSpinBoxEditDelegate::sizeHint which is called very often and creates a dedicated QDoubleSpinBox every time. The latest commit precalculates and caches the sizeHint. This improved performance massively in my testdriver (QTreeWidget with 8 columns / 10000 rows). With this change, inline-editing hopefully can be kept for measures, creating a unified UX for the settings.

Bingo! Now performance is much better, in fact I see no differences in perceived performance with current master.

>Ale, if the performance improvement is good enough for you, I will add the same pattern to all delegates in src/Gui/StyledItemDelegates.

Dialog start still takes 6-7 secs, likely due to this same issue in Zones, I will test again when you apply that change.

thejockl:
Didn't implement sizeHint-precalculation for DirectoryPathDelegate as it requires runtime information (and it is not used in tables with too many rows)

amtriathlon:
Thank you, athlete settings dialog shows much faster now, I will use this and merge the next week.
Let me now when you are done with these changes so I can trigger snapshot builds with updated translation files so translators can use the binaries for testing translations without the need to build from source.

thejockl:
For now, this topic is fine for me. I will for sure jump back in when issues or questions arise

amtriathlon:
This PR introduced a bug when adding new measures using Imperial units, it was reported in https://groups.google.com/g/golden-cheetah-users/c/hAwFNWP3-G0/m/-pgPYkhfAwAJ and I can reproduce it using weight as example.

PR #4568, opened 2024-11-04
Changed util/antlog.py to be compatible with python3

peret2000:
It maybe not extensively used, but, when it is needed, you have to convert it to python3, unless you have or can create a python2 environment. To keep that tool undeprecated, it must be able to run in python3


amtriathlon:
Thank you!

PR #4572, opened 2024-11-13
Added option to hide Python Fixes from Edit-Menu

thejockl:
* Added flag "Automated Execution only" to Python Fix Editor
* Hiding Python Fixes from the Edit-Menu if they have the flag set
* Removed the submenu to manage Python Fixes from the Edit-Menu
* Added a new tab "Python Data Processors" to Options > Data Fields to manage the Python Fixes
* Manage Python Fixes: Showing automated only Python Fixes in italics
* Manage Python Fixes: Added support to create Python Fixes
* Manage Python Fixes: Editing Python Fixes on double click

See thread https://groups.google.com/g/golden-cheetah-users/c/U00RH76BJoI


thejockl:
Thanks for testing and the feedback

My motivation to add the checkbox to the Python-Editor was twofold:
* Keep all settings for the script (name, checkbox, code) at one place
* Have the option to add a description later to the Py Fixes (similar to the core processors showing it in the dialog on manual execution)

But lets discuss to find the best solution

amtriathlon:
> My motivation to add the checkbox to the Python-Editor was twofold:
> 
> * Keep all settings for the script (name, checkbox, code) at one place
> * Have the option to add a description later to the Py Fixes (similar to the core processors showing it in the dialog on manual execution)
> 
> But lets discuss to find the best solution

I think it’s ok to have all items defining a Python DP here: name, code and -if we decide to add them- description and parameters.

OTOH, configuration options s.t. execution modes and default arguments should go to DP configuration for both built in and custom Saps.

thejockl:
This is my (quick & dirty) proposal for adding the information / configuration options to the Processing-Tab:
image

I would like to keep the Apply-column to get a quick overview of processors configured for execution on import / save

For PyFixes a button to open the Python-Editor can be shown in the Setup-group. and +--Buttons could be shown on the bottom to add / remove Pyfixes.

A filter to hide the core-processors can simulate the current "Python Data Processors"-Tab even more

amtriathlon:
> This is my (quick & dirty) proposal for adding the information / configuration options to the Processing-Tab: image
> 
> I would like to keep the Apply-column to get a quick overview of processors configured for execution on import / save

Agree, perhaps it could be renamed Auto with manual replaced by None or Never, the checkbox in an additional column named Manual to indicate whether each DP is available in Edit menu and Batch processing list. There are several very specific built in DPs which some users may prefer to disable.

> For PyFixes a button to open the Python-Editor can be shown in the Setup-group. and +--Buttons could be shown on the bottom to add / remove Pyfixes.
> 
> A filter to hide the core-processors can simulate the current "Python Data Processors"-Tab even more

You could do that if you prefer to remove de Python DP tab altogether, but perhaps it is simpler/clearer to keep it separated, not sure about that.

thejockl:
> Agree, perhaps it could be renamed Auto with manual replaced by None or Never, the checkbox in an additional column named Manual to indicate whether each DP is available in Edit menu and Batch processing list. There are several very specific built in DPs which some users may prefer to disable.

In this PR I initially put the "automated only" config setting to the FixPyScript (was the simplest way), but this can't be used for the Core Processors (personally I don't use anything related to swimming or running). Currently I think about how to move it to DataProcessorConfig. Therefore please don't merge this PR yet, it should be generalized first.

> You could do that if you prefer to remove de Python DP tab altogether, but perhaps it is simpler/clearer to keep it separated, not sure about that.

Added action buttons to the Processing-tab in my local version. It is even possible to filter for PyFixes which makes the "Python Data Processors"-tab obsolete:
image
Please note that the right-hand side of this dialog is far from final

thejockl:
This is my proposal to add a flag `automatedOnly` to the DataProcessor, valid for both core processors and PyFixes: I added a pair of getter / setter (implemented directly in DataProcessor, (no need to overwrite) to retrieve the value from the global settings, based on the new virtual method id.
``` c++
class DataProcessor
{
    public:
        DataProcessor() {}
        virtual ~DataProcessor() {}
        virtual bool postProcess(RideFile *, DataProcessorConfig*settings=0, QString op="") = 0;
        virtual DataProcessorConfig *processorConfig(QWidget *parent, const RideFile* ride = NULL) const = 0;
        virtual QString name() const = 0; // Localized Name for user interface
        virtual QString id() const = 0; // Unique Id for internal handling
        virtual bool isCoreProcessor() const { return true; }
        bool isAutomatedOnly() const;
        void setAutomatedOnly(bool automatedOnly);
        virtual QString explain() const = 0;

        static QString configKeyAutomatedOnly(const QString &id);
        static QString configKeyApply(const QString &id);
};
```
The access to the flag is directly in the DataProcessor instead of the DataProcessorConfig to keep the two configuration scopes (technical and business) separate.

For the Core-Processors id returns the hardcoded classname, for PyFixes it will be the name.
A similar pair of getter / setter could be implemented to remove internal knowledge from the ProcessorPage and the DataProcessorFactory.

Based on this id, the need to give a name when registering the Processor could become unnecessary. However, this would require a migration of the current settings for the apply-values.

thejockl:
See #4576 for a more generic solution

amtriathlon:
Replaced by https://github.com/GoldenCheetah/GoldenCheetah/pull/4576

PR #4576, opened 2024-11-23
"Automated only"-flagging for DataProcessors

thejockl:
This PR provides a solution to mark all types of DataProcessors as "Automated execution only" and replaces #4572 

* Options > Data Fields > Processors & Automation (renamed from "Processing")
  * Modernized the UI
  * Added UI to set Automation (None, On Import, On Save) for all processors
  * Added UI to set Automated execution only for all processors
  * Showing the processors description
  * Showing the processors setting (if available) in a more userfriendly way
  * Added option to add / delete / edit custom Python processors
  * Enabled editing of Python processors via double click
  * Added option to hide code processors
* Removed the submenu to manage Python Fixes from the Edit-menu
* Renamed the Edit-menu to "Process"
* Hiding "automated only"-processors from Process-menu and Batchprocessing-dialog
* DataProcessors
  * Turned configuration of all core processors into forms
  * Changed the technical name of all core processors to match their classname
  * Added legacy technical name to all core processors to support existing scripts
  * Moved description from config-class to DataProcessor
  * Implemented a migration path for existing persisted processor configurations
* GSettings
  * Added method to remove settings by key

### DIFF
--- a/src/Core/DataFilter.cpp
+++ b/src/Core/DataFilter.cpp
@@ -7429,7 +7429,7 @@ Result Leaf::eval(DataFilterRuntime *df, Leaf *leaf, const Result &x, long it, R
                         QString dp_name = *(leaf->fparms[0]->lvalue.n);
 
                         // lookup processor
-                        DataProcessor* dp = DataProcessorFactory::instance().getProcessors().value(dp_name, NULL);
+                        DataProcessor* dp = DataProcessorFactory::instance().getProcessor(dp_name);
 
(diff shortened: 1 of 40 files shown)


thejockl:
Screenshot_20241123_183433

amtriathlon:
Thank you, Joachim, it looks very nice! I will take a more detailed look the next week.

amtriathlon:
It works as expected in my tests I found no regression, UX and code is a good improvement, so I will merge and generate snapshot to enable more testing.

Issue #3200, opened 2019-10-25
Editing settings for Training devices


TheSmartGerman:
It would be great to have the possibility to edit the settings for training devices.

Right now you can just add or delete. If you like to try different settings, it would be great to enter the gui again.

I finally edited the *.ini file with notepad++

PR #4578, opened 2024-12-07
Reworked the Options-Dialog

thejockl:
* Added a new class ActionButtonBox, similar to QDialogButtonBox to reduce code duplication
* Created a new QStyledItemDelegate for inline editing of lists
* Marking the currently selected category (General, Appearance, ...)
* Using the same reset button for Appearance and Measures
* Using the same style and behaviour as in athlete settings for all QTreeWidgets
* Using QFormLayout as much as possible, including support for the setting Mac styled Forms
* General: Reordered the settings, grouped them and added a (optional) scrollbar
* Data Fields
  * Fields
    * Enabled inline editing
      * Added autocompletion to Screen Tab
      * Added a uniqueness validator to Field
      * Added the list-delegate to Values
  * Colour Keywords: Added the list-delegate to Related Notres Words
  * Defaults: Added autocompletion to Field and Linked field
* Metrics
  * Swapped the tab order
  * Favourites
    * Made the lists wider
    * Moved the up-/down-buttons below the Favourites list
* Training
  * Swapped the tab order of Preferences and Train Devices
  * Remote Controls
    * Switched to inline editing
    * Made the Action non-editable
  * Virtual Bicycle Specifications: Redesigned the dialog

### DIFF
--- a/src/Gui/ActionButtonBox.cpp
+++ b/src/Gui/ActionButtonBox.cpp
@@ -1,0 +1,281 @@
+/*
+ * Copyright (c) 2024 Joachim Kohlhammer (joachim.kohlhammer@gmx.de)
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include "ActionButtonBox.h"
+
+#include "Colors.h"
+
+
+ActionButtonBox::ActionButtonBox
+(ActionButtonBox::StandardButtonGroups standardButtons, QWidget *parent)
+: QWidget(parent)
+{
+    layout = new QHBoxLayout(this);
+    layout->setSpacing(2 * dpiXFactor);
+    layout->setContentsMargins(0, 0, 0, 0);
+    if (standardButtons.testFlag(StandardButtonGroup::UpDownGroup)) {
+#ifdef Q_OS_MAC
+        up = new QPushButton(tr("Up"));
+        down = new QPushButton(tr("Down"));
+#else
+        up = new QToolButton(this);
+        up->setArrowType(Qt::UpArrow);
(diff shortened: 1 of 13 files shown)


thejockl:
As example the most changed dialogs:
image
image
image

liversedge:
I'd merge this just to get rid of the hideous virtual bicycle UI we currently have. Thank you for doing this !

thejockl:
Looking at the screenshots now with a little more distance - should I reduce the precision in the derived statistics?

liversedge:
Ha- if I had my way we'd get rid of the whole thing. I can bet you that the number of people adjusting these is ~0

thejockl:
If you ask me - this is the first time I touched it! And I use Golden Cheetah a lot for indoor training

amtriathlon:
Agree that level of customization is overkill, while I think physics simulation to obtain speed from power is a nice feature, so many parameters give an unjustified illusion of precision. Anyway, it is much better organized now and the 3/4 parameters that really matter are more accessible. It is a nice change. For training view config we also have https://github.com/GoldenCheetah/GoldenCheetah/issues/3200 pending.

In general I think the changes are an improvement, I haven't found any issue so I will use a couple of days and merge this if there are no objections.

amtriathlon:
Thank you!

adamcstephens:
I'm getting the following compilation error from some of this work. Am I holding it wrong or is this an actual problem?

```
golden-cheetah> Gui/Pages.cpp: In member function 'void ProcessorPage::processorSelected(QTreeWidgetItem*)':
golden-cheetah> Gui/Pages.cpp:2713:17: error: 'actionButtons' was not declared in this scope; did you mean 'ActionButtonBox'?
golden-cheetah>  2713 |             if (actionButtons != nullptr) {
golden-cheetah>       |                 ^~~~~~~~~~~~~
golden-cheetah>       |                 ActionButtonBox
```

thejockl:
> I'm getting the following compilation error from some of this work. Am I holding it wrong or is this an actual problem?
> 
> ```
> golden-cheetah> Gui/Pages.cpp: In member function 'void ProcessorPage::processorSelected(QTreeWidgetItem*)':
> golden-cheetah> Gui/Pages.cpp:2713:17: error: 'actionButtons' was not declared in this scope; did you mean 'ActionButtonBox'?
> golden-cheetah>  2713 |             if (actionButtons != nullptr) {
> golden-cheetah>       |                 ^~~~~~~~~~~~~
> golden-cheetah>       |                 ActionButtonBox
> ```

Are you sure you are on the latest commit? On the initial one I had forgotten to add the (new) class ActionButtonBox. Please do a git pull and check if you get the files src/Gui/ActionButtonBox.cpp|h
If they are already in your tree, try re-running qmake -r. Good luck - and come back to me if this didn't help

adamcstephens:
Yeah, I'm using 3.7-DEV2412 which is only missing a couple non-code commits. It's probably something in the nix build environment, so I'll dig in and see if I can figure out what's going on. My C++ isn't the strongest though. :)

PR #4579, opened 2024-12-14
RideNavigator: Always show colour bar

thejockl:
* Show the colour bar (see Options > Data Fields > Colour Keywords) in RideNavigator even when an activity is selected
* Same for intervals
* Use contrasting colours for text (especially when using option "Use for Background")
* Replaced focus with active (QStyle::State) to avoid different text colors
* Remove trailing whitespaces from the code

image
image

### DIFF
--- a/src/Gui/IntervalTreeView.cpp
+++ b/src/Gui/IntervalTreeView.cpp
@@ -160,41 +160,34 @@ IntervalTreeView::mimeData
     return returning;
 }
 
-void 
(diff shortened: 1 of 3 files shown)


liversedge:
I prefer the current ui personally

thejockl:
> I prefer the current ui personally

From a colour-bar perspective, text colour - or both? I could add a config-switch for the bar

amtriathlon:
We will not merge this, Joachim, sorry.

PR #4581, opened 2024-12-18
Workoutfilter: Support for categories

thejockl:
* Added support for the new filter keywords
  * category
  * subcategory
  * categoryindex
* Importing category, subcategory, categoryindex and tags from zwo files
* Updated the TrainDB-Schema to v3
* Added a migration mechanism to update the TrainDB and automatically add the new fields
* Refactored reading zwo-files from QXmlSimpleReader to QXmlStreamReader
  * Attention: QXmlStreamReader is stricter and fails on attributes containing &-signs not starting entities; zwo-files containing this must be fixed (& -> &amp;)
* Added a unittest for ZwoParser

See https://groups.google.com/g/golden-cheetah-users/c/raK7cUfgEg0


src/Train/ModelFilter.cpp
```diff
@@ -215,7 +226,7 @@ ModelStringContainsFilter::accept
     }
     bool contains = _all;
     QString dataValue = data.toString();
-    for (const auto &item : qAsConst(_values)) {
+    for (const auto &item : std::as_const(_values)) {
```
peret2000, line 229: There is a compilation error:
`Train/ModelFilter.cpp:229:34: error: ‘as_const’ is not a member of ‘std’; did you mean ‘is_const’?
`

thejockl: My thought was to switch from the deprecated qAsConst to the replacement std::as_const. But obviously this is not an alternative for now

amtriathlon:
I don’t like this PR, is too complex, specific to zwo files and doesn’t seem to fulfill the original request according to the group’s thread.

PR #4582, opened 2024-12-19
Text in Trainbottom window is word wrapped (Training Window)

peret2000:
Messages can be seen in several lines if they are large, not truncated
Specially useful when training a workout with long text messages during the training


amtriathlon:
Thank you!

PR #4584, opened 2024-12-23
Fixed calculation of measures when using imperial units

thejockl:
See https://groups.google.com/g/golden-cheetah-users/c/hAwFNWP3-G0


amtriathlon:
Thank you!

PR #4585, opened 2024-12-23
Fixed compilation error when python is disabled

thejockl:
See https://groups.google.com/g/golden-cheetah-users/c/d-A1M2FlcZI


amtriathlon:
Thank you!

PR #4588, opened 2025-01-05
Field colouring using expressions

paulj49457:
Given the recent discussions on Sport & SubSport fields I intended to update my activities to use this structure, but this would prevent me having my existing activity navigator colouring scheme, as keyword colouring is based upon a single metadata item, not a combination of Sport & SubSport....

**Warning: Before testing this PR please backup your metadata.xml file, as the following PR will automatically convert your keyword definitions to expressions, and update the keyword section of the metadata.xml file to a new format.**

I have therefore created this PR which updates the keyword colouring to be based upon filter expressions, please test and let me know if a) it useful, b) it needs any changes/improvements.

image

Notes:
i) The existing "Reverse" keyword can be used in place of an expression, and when "Use for background" is checked it will set all the text to the colour defined for "Reverse"
ii) Tokens are not supported by this PR, as the expressions should support multiple keyword options.

### DIFF
--- a/src/Charts/TreeMapWindow.cpp
+++ b/src/Charts/TreeMapWindow.cpp
@@ -65,9 +65,8 @@ TreeMapWindow::TreeMapWindow(Context *context) :
 
     // read metadata.xml
     QString filename = QDir(gcroot).canonicalPath()+"/metadata.xml";
-    QString colorfield;
     if (!QFile(filename).exists()) filename = ":/xml/metadata.xml";
-    RideMetadata::readXML(filename, keywordDefinitions, fieldDefinitions, colorfield, defaultDefinitions);
+    RideMetadata::readXML(filename, keywordDefinitions, fieldDefinitions, defaultDefinitions);
 
     //title = new QLabel(this);
     //QFont font;
--- a/src/Core/Athlete.h
+++ b/src/Core/Athlete.h
@@ -57,7 +57,6 @@ class RideAutoImportConfig;
 class RideCache;
 class IntervalCache;
 class Context;
-class ColorEngine;
 class AnalysisSidebar;
 class AthleteTab;
 class Leaf;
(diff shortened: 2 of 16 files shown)


paulj49457:
Currently the existing metadata tokens are ignored on the automatic update to expressions, if this PR might get merged, I could revisit the PR and add the conversion of tokens to expressions so no information is lost.

amtriathlon:
Hi Paul, I have not review the changes but, conceptually, I have mixed feelings.
While I think this is kind of a natural generalization, it goes in the direction we are frequently criticized for: it makes configuration more complex making a simple and direct cosmetic feature in one more complex to understand, use and maintain (not to mention if current settings are invalidated, which is a no go), for a benefit only sophisticated  users would appreciate.

liversedge:
I think we are bike shedding here- the fitness landscape is enormous and there are so many missing features in GC that would be transformative (e.g. planning) but instead we are focusing energy on something unimportant because it is comfortable and familiar.

paulj49457:
> I think we are bike shedding here- the fitness landscape is enormous and there are so many missing features in GC that would be transformative (e.g. planning) but instead we are focusing energy on something unimportant because it is comfortable and familiar.

No problem with field colouring PR, to be honest after creating the PR I achieved the same result using the existing functionality :)

@liversedge @amtriathlon On the subject of adding missing features, I have been using the equipment tiles (#4533) for the last 6 months and I find the feature very useful for bike maintenance and determining running shoe mileage, so would take a look at the PR ?

amtriathlon:
> @amtriathlon On the subject of adding missing features, I have been using the equipment tiles (#4533) for the last 6 months and I find the feature very useful for bike maintenance and determining running shoe mileage, so would take a look at the PR ?

Hi Paul, I have not looked at that PR in detail because current functionality is enough for the level of equipment tracking I am interested in, I understand it may be different for other users, mainly those coming from SportTracks, but it is not my thing.
Superficially, I am not sure we need that facility at the view level and to store data in charts is a good idea, but I don't plan to review it, honestly I have more than enough with muy current involvement in the project to add more demanding tasks, sorry.

paulj49457:
Regarding #4533 can we retain it as an open PR, as I'm happy to maintain it so it tracks the latest GC changes and remains available for people to use?

PR #4590, opened 2025-01-10
Batch processing metadata auto completion field

paulj49457:
Added auto completion to the batch processing metadata option:

image


amtriathlon:
Thank you!

PR #4596, opened 2025-01-18
Move workout filter to the toolbar

### LINKED ISSUE #4591: View dependent toolbar search box
This is motivated by @liversedge comment in https://groups.google.com/g/golden-cheetah-users/c/VL0ELt-L1jQ/m/rSvo46WuBgAJ, currently toolbar search box is view independent, but it only makes sense for Activities and Trends, Athletes lack this functionally for now and Train has a newly added search box in the sidebar.
Making it view dependent we could better reflect current functionally in the UX.

amtriathlon:
Now the search/filter box in the toolbar depends on the active view:
- No search/filter box for Athletes view
- Workout filter box for Train view
- Activities search/filter box for Activities/Trends/Diary views

Fixes #4591

PS: I kept the copyrights unchanged since there is not much code of my own, mostly a copy and paste refactor.

### DIFF
--- a/src/Core/Context.h
+++ b/src/Core/Context.h
@@ -121,6 +121,7 @@ class Context : public QObject
         bool showSidebar, showLowbar, showToolbar, showTabbar;
         int style;
         QString searchText;
+        QString workoutFilterText;
         bool scopehighlighted;
 
         // ride item
--- a/src/Core/Context.h
+++ b/src/Core/Context.h
@@ -72,6 +72,7 @@ class AthleteTab;
 class NavigationModel;
 class RideMetadata;
 class ColorEngine;
+class ModelFilter;
 
 
 class GlobalContext : public QObject
@@ -185,6 +186,9 @@ class Context : public QObject
         void setFilter(QStringList&f) { filters=f; isfiltered=true; emit filterChanged(); }
         void clearFilter() { filters.clear(); isfiltered=false; emit filterChanged(); }
 
+        void setWorkoutFilters(QList<ModelFilter*> &f) { emit workoutFiltersChanged(f); }
+        void clearWorkoutFilters() { emit workoutFiltersRemoved(); }
+
         // user metrics - cascade
         void notifyUserMetricsChanged() { emit userMetricsChanged(); }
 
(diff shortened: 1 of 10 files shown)


amtriathlon:
Hi @thejockl, as the original author, can you check this?

thejockl:
Hi @amtriathlon, the functionality is mostly unchanged, only leading spaces don't open the autocompletion any more.

See my attached patch to style the WorkoutFilterBox similar to the pre-existing SearchBox: [workoutfilterbox.diff.gz](https://github.com/user-attachments/files/18468859/workoutfilterbox.diff.gz)

From a UX standpoint it feels strange typing on the top-right and seeing the filtered result on the left, especially since it filters as you type. This was my main motivation putting the filterbox where it was, reducing required eye movement. Could it be an option to at least center the searchboxes in the toolbar? Yes, the screenshot shows a quick hack, it is not perfectly centered:
image


In combination with the rejected PR #4581 I have implemented a filter manager. Please let me know if there is interest, otherwise I will keep it for myself.

amtriathlon:
> Hi @amtriathlon, the functionality is mostly unchanged, only leading spaces don't open the autocompletion any more.
> 
> See my attached patch to style the WorkoutFilterBox similar to the pre-existing SearchBox: [workoutfilterbox.diff.gz](https://github.com/user-attachments/files/18468859/workoutfilterbox.diff.gz)

I applied this patch, it looks much better now, thank you!

> From a UX standpoint it feels strange typing on the top-right and seeing the filtered result on the left, especially since it filters as you type. This was my main motivation putting the filterbox where it was, reducing required eye movement.

I think this behavior is familiar for GC users, but perhaps to apply filters as you type is overkill and we could wait until enter to make it more similar to search/filter box.

 Could it be an option to at least center the searchboxes in the toolbar? Yes, the screenshot shows a quick hack, it is not perfectly centered: image

May be, on of the advantages of the toolbar placement is more space, sidebar tends to be kept small, but I would wait for @liversedge option as he tends to have a more global view of the project.

> In combination with the rejected PR #4581 I have implemented a filter manager. Please let me know if there is interest, otherwise I will keep it for myself.

Yes, something like named filters in the Search/Filter box would be interesting, #4581 was too specific for a particular use case, IMHO.

thejockl:
I don't like the latest commit (Apply the filter when editing finished)

amtriathlon:
Honestly, I am not sure, filter as you type feels weird to me, but perhaps I am just too used to GC search/filter behavior.
I have no problems to revert that, but I will wait for @liversedge review before doing anything else about this.

liversedge:
Its quite nice when filtering on free text but really weird when using expressions- my preference is to just be consistent and don't make the user think- so process it when they hit return

amtriathlon:
Thank you. I have to look at context changes and context help before merging.

PR #4216, opened 2022-05-01
travis/linux: set QTWEBENGINE_DISABLE_SANDBOX=1

MeisterP:
There is an issue with qtwebengine (chromium) and glibc 2.34

See: https://codereview.qt-project.org/c/qt/qtwebengine-chromium/+/374204
Fixes: https://github.com/GoldenCheetah/GoldenCheetah/issues/4196

PR #4597, opened 2025-01-20
Revert "travis/linux: set QTWEBENGINE_DISABLE_SANDBOX=1 (#4216)"

MeisterP:
This reverts commit 9b29fb2f8de05eb72cd1d30403533a09cdd24fa2.

Disabling the sandbox like this was a good workaround at the time. But we shouldn't stick with it for security reasons.
Ubuntu 20.04 is EOL in April 2025 and was the last version using the problematic glibc 2.34

This is not urgent. But maybe add it to the [3.7 milestone](https://github.com/GoldenCheetah/GoldenCheetah/milestone/13)  so it doesn't fall under the radar.

### DIFF
--- a/travis/linux/after_success.sh
+++ b/travis/linux/after_success.sh
@@ -18,15 +18,6 @@ mkdir -p appdir
 # Executable
 cp GoldenCheetah appdir
 
-# AppRun file
-cat >appdir/AppRun <<EOF
-#!/bin/bash
-HERE="\$(dirname "\$(readlink -f "\${0}")")"
-export QTWEBENGINE_DISABLE_SANDBOX=1
-exec "\${HERE}/GoldenCheetah" "\$@"
-EOF
-chmod a+x appdir/AppRun
-
 # Desktop file
 cat >appdir/GoldenCheetah.desktop <<EOF
 [Desktop Entry]


amtriathlon:
Hi, can you check if this works using Hi, can you check if removing `export QTWEBENGINE_DISABLE_SANDBOX=1` from AppRun works  using https://github.com/GoldenCheetah/GoldenCheetah/releases/download/snapshot/GoldenCheetah_v3.7-DEV_x64_new.AppImage? I haven't got the environment to reproduce the origina issue.

amtriathlon:
> On all three of them I get:
> 
> ```
> error while loading shared libraries: libQt5Core.so.5: cannot open shared object file: No such file or directory
> error while loading shared libraries: libQt5WebEngineCore.so.5: cannot open shared object file: No such file or directory
> ```
It's weird, I extracted the AppImage and both files are there and correctly pointed, can you check doing something like:
```
./GoldenCheetah_v3.7-DEV_x64_new.AppImage --appimage-extract
ls -l squashfs-root/lib/libQt5WebEngineCore.so.5
ldd -v squashfs-root/GoldenCheetah|grep libQt5WebEngineCore
```

amtriathlon:
Great, thank you! Here is the AppImage with the fixed applied: https://github.com/GoldenCheetah/GoldenCheetah/releases/download/snapshot/GoldenCheetah_v3.7-DEV_x64.AppImage

peret2000:
AppImage is working fine in Ubuntu oracular (24.10), tested map charts.
As a side note, Appimage contains `export QTWEBENGINE_DISABLE_SANDBOX=1` in AppRun, is that what we should test?
I have set it to 0, and also executed directly the binary once extracted the appimage, and confirmed that environment variables is set to 0 writing its value in a python chart. The map chart is ok. However, the log continues to say 'Debug: Sandboxing disabled by user.'

peret2000:
Thanks, @MeisterP, then, confirmed that undeclaring QTWEBENGINE_DISABLE_SANDBOX, it works fine with Ubuntu 24.10

Issue #4592, opened 2025-01-13
Review the new Splash Screen


amtriathlon:
I think https://github.com/GoldenCheetah/GoldenCheetah/pull/4468 is an improvement vs the terse version we had, but @liversedge doesn’t agree and we will need to review/revert this before release, this issue is to discuss these changes.

liversedge:
a) its WAY too big (especially on a 4k monitor) b) it doesn't say anything useful c) its ugly- the GC imagery is well past its sell-buy date.

thejockl:
a) As I don't have a 4k monitor, I can't comment on this one. On my machines it is on par with other applications: Larger than LibreOffice but smaller than Krita or Gimp
b) GoldenCheetah is a slow starter. Therefore the splashscreen indicates to the user during the startup phase: All good, we are still loading and nothing crashed. The old counter didn't say anything about the application it belongs to and disappeared long before the main window showed up
c) This reflects the style (including the website) at the time I implemented it. The image can be replaced easily if something better is available

ZajtiM:
I agree with @thejockl in all points.

But i also think colors should be updated a bit. Current color combination looks like some old wallpaper in grandmas house. A bit more vibrant colors can do a lot.

Here is a link for website where you can find vectors (domain names says it all). And it show triathlon options.
[https://www.vecteezy.com/free-vector/triathlon](https://www.vecteezy.com/free-vector/triathlon)
Some vectors are free, some not. I'm not saying new Splash screen should use something from that website. But it might help to get some ideas what to use / create.

Is GC allowed to use logos / equipment from certain companies? Adding some sort of charts, etc. might be good. Like when you start "fresh" GC and don't have any activities.

Btw, i was also thinking it might be interesting to have two or three images which are switched after certain amount of time. Sort of like banner carousel on websites. You can also see this during OS installation. That way image(s) are not too crowded / busy.

Just a few ideas.

amtriathlon:
> a) As I don't have a 4k monitor, I can't comment on this one. On my machines it is on par with other applications: Larger than LibreOffice but smaller than Krita or Gimp

Other parts of the code use dpiXfactor/dpiYfactor for size adjustments, they are initialized from https://github.com/GoldenCheetah/GoldenCheetah/blob/f7e91e7fae5faddb847fba86db2d3b88a635784a/src/Core/Settings.cpp#L785

[...]

PR #4599, opened 2025-01-21
Updated the splashscreen

thejockl:
* See discussion in #4592
* Artwork created and contributed by @ZajtiM


amtriathlon:
It looks nice in my first test:
image

ZajtiM:
Just a couple of question regarding image file format. Why using png? Is it because of OS support? 
PNG gets distorted when not in original size. I use svg for all my icons and splash image in my Qt app. But i'm running / using it only in Linux. So i'm not sure if there would be a problem with svg.

Main point of using svg is, it is vector. So you can resize it as much as you want and it won't be distorted.

You can see version and progress text is clear, but bottom part is clearly distorted.

thejockl:
@ZajtiM , fair point and good idea. Will give it a try.
And for the reasoning why PNG: I just did what I had done before (TM)

liversedge:
I am much happier now, thank you !

amtriathlon:
I am not sure I could detect the difference in my use case:
image
but it seems a good idea, anyway.

ZajtiM:
@amtriathlon i think it is your windows setting that is affecting how it is displayed (distorted). I only use windows at work, otherwise i use linux like 99% of my time. So i'm not 100% what it is. But if i remember correctly you can set zoom. Used mainly for 4K monitors, since everything is very small. And when you use zoom, it distorts images, etc. But not text. That's why text is nice and clear, but splash image is dostorted.

I did some testing. First i used original svg file. And then i also made extra svg with size 150x166px.
Here is much smaller version of splash svg file:
splashscreen1

And this is splash using small splash svg during GC load. It looks perfectly clear.
splash_screenshot

amtriathlon:
> @amtriathlon i think it is your windows setting that is affecting how it is displayed (distorted). I only use windows at work, otherwise i use linux like 99% of my time. So i'm not 100% what it is. But if i remember correctly you can set zoom. Used mainly for 4K monitors, since everything is very small. And when you use zoom, it distorts images, etc. But not text. That's why text is nice and clear, but splash image is dostorted.

Yes, I am using zoom on Windows 11, and that distortion is irrelevant for me, TBH.

I will merge this, if there are no objections.

PR #4600, opened 2025-01-21
Change CoreTemp suffix to 'e'

ihewitt:
Hadn't noticed/realised that I'd still used 't' for both Tempe and CoreTemp, so this changes the CoreTemp suffix to "e".


ihewitt:
Many thanks @amtriathlon

PR #4601, opened 2025-01-22
Enable compiling VideoWindow with QT6 toolset

ihewitt:
Qt6 has removed QMediaContent, so minor update, adds GC_VIDEO_QT6 flag to enable QT6 compilation with Qt video window.

### DIFF
--- a/src/Gui/GcCrashDialog.cpp
+++ b/src/Gui/GcCrashDialog.cpp
@@ -286,6 +286,8 @@ QString GcCrashDialog::versionHTML()
             .arg("quicktime")
 #elif defined GC_VIDEO_QT5
             .arg("qt5")
+#elif defined GC_VIDEO_QT6
+            .arg("qt6")
 #elif defined GC_VIDEO_VLC
             .arg("vlc")
 #else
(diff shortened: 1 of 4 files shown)


ihewitt:
I wasn't sure if the QT Video window is going to be continued with, but I build with Qt6 and noticed it failed if I used it instead of VLC video.

amtriathlon:
Hi Ivor, for v3.6 we evaluate switching from VLC to Qt5 video, main issue was seek performance making video sync unusable.
I will merge this and it would be interesting to test again to see if it seek performance was fixed in Qt6 since it will make building much simpler.
Thank you!

amtriathlon:
Yes, it is Qt6 only and it should be guarded, building with Qt5.15:

`Train\VideoWindow.cpp(152): error C2039: 'setAudioOutput': is not a member of 'QMediaPlayer'`

No problem with the .h change OTOH.

ihewitt:
Cheers @amtriathlon, got the QT6 video window working with the meters overlays nicely, performance *seems* ok, but I need to test properly, unfortunately I'm full of cough and cold at the moment 🤒 so limited real-use so far.
Will tidy up over the weekend.

Issue #4606, opened 2025-01-25
Train view - Filter - GC crash


ZajtiM:
I downloaded latest snapshot build, created few hours ago. 

When using filter in Train view GC crashes. When i insert my text and press enter, it finds workout and shows it in train view. But then after a few seconds it crashes.
It looks like it "selects" first workout from Workouts list that matches filter query. I would assume that default action would be to only filter Workouts list, not selecting a workout as well. It is QTreeView, so it could have nothing selected.

Here is gdb backtrace:
```
Thread 1 "GoldenCheetah" received signal SIGSEGV, Segmentation fault.
0x00007fffee0b2711 in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
(gdb) backtrace
#0  0x00007fffee0b2711 in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
#1  0x00007fffee0b2cac in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
#2  0x00007fffee0b30ad in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
#3  0x00007fffee0b408e in QSortFilterProxyModel::invalidateFilter() () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
#4  0x00005555560a7f02 in MultiFilterProxyModel::setFilters(QList<ModelFilter*>) ()
#5  0x00005555561b9474 in TrainSidebar::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) ()
#6  0x00007fffee114660 in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
#7  0x000055555615e099 in Context::workoutFiltersChanged(QList<ModelFilter*>&) ()
#8  0x00005555560ab4d8 in WorkoutFilterBox::editingFinished() ()
#9  0x00007fffee114660 in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
#10 0x00007ffff6c4c0ef in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Widgets.so.5
#11 0x00007fffee114660 in ?? () from /tmp/.mount_GoldenGcEpnB/lib/libQt5Core.so.5
[...]

PR #4607, opened 2025-01-25
Fix crash caused by WorkoutFilterBox (see #4606)

thejockl:
Root cause of the issue:
* MultiFilterProxyModel takes ownership of the filterlist (QList<ModelFilter*>) and deletes the previously set version, including the contained ModelFilters when clearing or setting a new list
* WorkoutFilterBox parses the filter (i.e. creates the filterlist) on textChange and keeps it in a member, even after setting it to the MultiFilterProxyModel
* When the event editingFinished is triggered in WorkoutFilterBox without reparsing, the filterlist is set again to MultiFilterProxyModel with already deleted ModelFilter*


amtriathlon:
Thank you!

PR #4611, opened 2025-01-30
Improve CoreTemp data

ihewitt:
Adds support for new "heat strain index" data from CoreTemp ANT messages

Details: [Heat Strain](https://corebodytemp.com/blogs/news/examples-of-the-heat-strain-index)
Not sure if the GoldenCheetah graphing could be used to create something like this (the strain is shown in the background coloured bars):
image

### DIFF
--- a/src/ANT/ANTChannel.cpp
+++ b/src/ANT/ANTChannel.cpp
@@ -1260,8 +1260,10 @@ void ANTChannel::broadcastEvent(unsigned char *ant_message)
 
             case CHANNEL_TYPE_TEMPE:
             {
-                if (antMessage.tempValid)
-                    parent->setTemp(antMessage.temp/100.0);
+                if (antMessage.tempValid) {
+                    value = value2 = antMessage.temp/100.0;
+                    parent->setTemp(value);
+                }
             }
             break;
 
--- a/src/ANT/ANTChannel.cpp
+++ b/src/ANT/ANTChannel.cpp
@@ -911,9 +911,9 @@ void ANTChannel::broadcastEvent(unsigned char *ant_message)
                 value2 = antMessage.skinTemp;
 
                 // Store in XDATA
-                emit tcoreData(antMessage.coreTemp, antMessage.skinTemp, antMessage.tempQual);
+                emit tcoreData(antMessage.coreTemp, antMessage.skinTemp, antMessage.heatStrain, antMessage.tempQual);
                 // and forward to telemetry
-                parent->setCoreTemp(antMessage.coreTemp, antMessage.skinTemp);
+                parent->setCoreTemp(antMessage.coreTemp, antMessage.skinTemp, antMessage.heatStrain);
             }
         }
         break;
--- a/src/ANT/ANT.cpp
+++ b/src/ANT/ANT.cpp
@@ -162,7 +162,7 @@ ANT::ANT(QObject *parent, DeviceConfiguration *devConf, QString athlete) : QThre
         connect(antChannel[i], SIGNAL(posData(uint8_t)), this, SIGNAL(posData(uint8_t)));
 
         // Core temp data
-        connect(antChannel[i], SIGNAL(tcoreData(float, float, int)), this, SIGNAL(tcoreData(float, float, int)));
+        connect(antChannel[i], SIGNAL(tcoreData(float, float, float, int)), this, SIGNAL(tcoreData(float, float, float, int)));
 
         // timer for master channel broadcasts
         connect(antChannel[i], SIGNAL(broadcastTimerStart(int)), this, SLOT(slotStartBroadcastTimer(int)));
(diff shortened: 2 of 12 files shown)


src/ANT/ANTChannel.cpp
```diff
@@ -1260,8 +1260,10 @@ void ANTChannel::broadcastEvent(unsigned char *ant_message)
 
             case CHANNEL_TYPE_TEMPE:
             {
-                if (antMessage.tempValid)
-                    parent->setTemp(antMessage.temp/100.0);
+                if (antMessage.tempValid) {
+                    value = value2 = antMessage.temp/100.0;
```
ihewitt, line 1264: Missed this before, sets "value" so the data displays in the search window.

src/Charts/OverviewItems.cpp
```diff
@@ -3187,6 +3187,8 @@ DataOverviewItem::itemPaint(QPainter *painter, const QStyleOptionGraphicsItem *,
                 // lets use the heat color, but with a little translucency
                 QColor brushcolor = Utils::heatcolor(heatvalue);
                 brushcolor.setAlpha(127);
+                int wid=100;
```
ihewitt, line 3190: This might not be needed, and might just be I had bad data, but I hit a crash locally because columnWidths was less than "i"

src/FileIO/FitRideFile.cpp
```diff
@@ -3280,6 +3292,12 @@ struct FitFileParser
         last_distance = km;
         last_altitude = alt;
 
+        if (core_index != -1) //did we get coretemp dev fields
+        {
+            //Parse into xdata here since this isn't in a separate record
+            decodeCoreTemp(def, secs, core_index, values);
+        }
```
ihewitt, line 3299: Is there a better place/way to do this? The coretemp isn't in a separate record so can't be handled in a switch like the other decodeX functions so I'm storing the "dev id" that the fields were found in then re-reading them in the decode function.

src/FileIO/FitRideFile.cpp
```diff
@@ -5397,8 +5484,35 @@ void write_record(QByteArray *array, const RideFile *ride, bool withAlt, bool wi
...
-            write_float32(ridePoint, point->tcore, true);
+            //Locate XData
+            XDataSeries* xdata_core = ride->xdata("TCORE");
+
+            double core = point->tcore;
+            double skin = 0.0, hsi = 0.0;
+            int    qual = 0;
+
+            if (xdata_core){
+                XDataPoint* xdp = nullptr;
```
ihewitt, line 5497: Is there a better way to find the XData correlating to a particular ride point? Here I'm just keeping a cursor into the xdata and moving it forwards to the next/nearest ride point since the XData could have missing/different data points.

src/FileIO/FitRideFile.cpp
liversedge: I think the Fit file updates should be a separate PR, there is too much going on at once in this one.

The approach doesn't look right, but would like to be able to test any Fit updates with example files, so best to do after the other changes are applied?

ihewitt: Thanks, will split out the fit update and I'll add a test file - roundtripping with a .fit file from a Forerunner looked good.

ihewitt:
Removed the fit file changes, this is just the additional ANT field parsing.

amtriathlon:
Thank you!

amtriathlon:
Hi Ivor, I noticed the following error logged when running current master, does the signature mismatch was introduced in this PR?
```
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.021] warning: QObject::connect: No such signal ANTChannel::tcoreData(float, float, float, int)
[9:29:01.033] warning: QObject::connect: No such signal ANT::tcoreData(float, float, float, int)
[9:29:01.033] warning: QObject::connect: No such signal ANTlocalController::tcoreData(float,float,float,int)
```

ihewitt:
Hell, current master? I'll check and fix that tonight. I think I can see the typo when I split the changes 🙁 (ANTChannel.h:261)
apologies..

Issue #2553, opened 2017-06-27
FIT export doesn't seem to work for XData


Nadelbaum:
There are no errors and clearly some part of the XData is available, but when importing a GoldenCheetah exported FIT file (with XData) back to Stryd Power Center all the values of XData fields will be 0.

havu1977:
Steps to reproduce:

1) Export a .FIT file from Stryd Power Center
2) Import that .FIT file into Golden Cheetah -> You will see "STANDARD" and "DEVELOPER" tabs with proper values
3) Immediately Export the imported data back to a .FIT file
4) Import the .FIT file exported from Golden Cheetah back to Stryd Power Center and you will see that all "DEVELOPER" (XData) values will be 0 even if the headers are there.

No workaround known at this time.

amtriathlon:
Effectively XData series are not exported in the FIT format, this is an unimplemented feature.

PR #4612, opened 2025-01-30
Read and write xdata developer fields to FIT files

ihewitt:
Read and writes CoreTemp xdata to FIT files - will discuss the approach on the forum to find a better way.

I've added a sample fit file, generated with a walk (still ill!) from a Forerunner wearing a CoreTemp sensor, I've then loaded the file into GoldenCheetah and resaved as a FIT file to test the round-trip, along with text dumps of the two files.

Code problems:
- The coretemp values are held as developer fields in the activity record in the FIT file so adding another "decodeXx()" handler in the "switch on record type" doesnt work for these.

[Edit, these should be resolved now]

> - The workaround I've implemented here is that when it detects the native "139" value (CoreTemp) it remembers the developer field id, then after the fields have been read it calls a second "decodeCore" to extract the additional developerfields which don't have native ids to populate the XData.
> - The second issue is the reverse. 🙂 When populating the FIT file to output I want to use the data from the xdata but these don't have a 1to1 relationship with the ride->point arrays so I keep a cursor into the xdatapoints and update it for the matching second.


ihewitt:
Cleaned up writing and reading from xdata.

src/FileIO/FitRideFile.cpp
ihewitt, line 1: I'm planning on changing write_dev_fields() line~5235 to handle generic developer fields from XData e.g. for #2553 but that's going to need quite a bit of a rework since I'll need to store the record 207 values.

amtriathlon:
Hi Ivor, your last comment makes me to think that is the right approach.
I imported your sample file to current master and I got:
image
Since those developer fields are already imported by the general logic we should not add yet another special case, but to adapt https://github.com/GoldenCheetah/GoldenCheetah/pull/4207 to use those standard XData fields instead, IMHO.

ihewitt:
Yeah I was actually wondering about doing the opposite initially and *excluding* the fields from the generic developer xdata if they're in a special case. But then thought, never mind it won't hurt to duplicate them. 😄
There's quite a bit of rejigging needed to get it working in a completely generic manner with the developer fields though, for a few reasons, I'll definitely make a start but might take a while to get working.

One problem is that the data needed to write out the valid "complete" developer fields is currently only hardcoded in the special case so the xdata would need the additional information added in somehow/somewhere for the native data types, ids, developer headers etc.
Also, the data table you have above for example doesn't include the actual "core temp" value, the QIC_ one you see there is actually a hack that the garmin data field uses to be able to display the Fahrenheit value, since the coretemp field has a "native num" assigned to it, then the GoldenCheetah reader excludes it from the developer fields on read. 

I've got a few ideas to try out anyway. Albeit cautiously, since the current FIT file code terrifies me and I don't want to break it.

amtriathlon:
Agree about the complexity of https://github.com/GoldenCheetah/GoldenCheetah/issues/2553 in general, and I don’t have problem with a special case for some dev fields s.t. these related to tcore as an intermediate step, which I don’t like is to add an special case for import when we already have a general mechanism, that sound like going in the wrong direction to me.

src/ANT/ANT.cpp
```diff
@@ -388,12 +388,12 @@ void ANT::requestFecCalibration(uint8_t type)
 
 void ANT::requestPwrCapabilities1(const uint8_t chan)
 {
-    sendMessage(ANTMessage::requestPwrCapabilities1(chan!=-1?chan:pwrChannel));
```
ihewitt, line 391: chan is uint8_t so gives a warning here.

ihewitt:
Experiment reading and writing generic FIT developer fields, needed metadata is held in a CIQ tag.
(not working for all generic data yet)

Is xData DEVELOPER *just* used for fit developer data? If so was thinking these could 1. be split into DEVEL1, DEVEL2 etc to collect per-device fields together... but also a lookup table for known values to then rename them if recognised from application id, so
18fb2cf0-1a4b-430d-ad66-988c847421f4->STRYD
6957fe68-83fe-4ed6-8613-413f70624bb5->CORE
etc
to make it easier to use/identify.

liversedge:
Please do not change how DEVELOPER fields are parsed- lots of user charts will break if you do

ihewitt:
gotcha @liversedge , makes sense and certainly a lot simpler to leave as is! thanks.

src/FileIO/CsvRideFile.cpp
```diff
@@ -1567,14 +1567,16 @@ RideFile *CsvFileReader::openRideFile(QFile &file, QStringList &errors, QList<Ri
...
-        // create the XDATA series
-        tcoreSeries = new XDataSeries();
-        tcoreSeries->name = "TCORE";
-        tcoreSeries->valuename << "Core" << "Skin" << "HSI" << "Quality";
-        tcoreSeries->unitname << "C" << "C" << "%" << "q";
+        // create the XDATA series        
+        develSeries = new XDataSeries();
+
+        develSeries->name = "DEVELOPER";
+        develSeries->valuename << "core_temperature" << "skin_temperature" << "heat_strain_index" << "core_data_quality";
```
ihewitt, line 1578: should move the coretemp specifics into a separate PR, but it's hard to get the generic change working without this to test with!

ihewitt:
Shall I close this PR and then can create two new/separate ones for the "fit generic field change" and the "core to generic developer fields" change

ihewitt:
Getting there:
image
This is from reading in, exporting, then reloading the Fr955v19.28andStryd.fit test file in test/runs.
Some fine-tuning still to-do.

amtriathlon:
Hi Ivor, have you merged an older version in https://github.com/GoldenCheetah/GoldenCheetah/pull/4612/commits/033d6967ea49b97ff8e930c89950d4f80d63d587?
It seems to have modifications to JSON ridefile for tcore.

ihewitt:
> Hi Ivor, have you merged an older version in [033d696](https://github.com/GoldenCheetah/GoldenCheetah/commit/033d6967ea49b97ff8e930c89950d4f80d63d587)? It seems to have modifications to JSON ridefile for tcore.

reverted. thanks.
Was trying to get the "Core Temperature" checkbox/graph enabled in the Ride/Curves, since the core is only in xdata/developer fields then the areDataPresent()->tcore never gets set. I'll figure out a different way to trigger/enable that.

src/FileIO/RideFile.cpp
```diff
@@ -1014,12 +1014,25 @@ RideFile *RideFileFactory::openRideFile(Context *context, QFile &file,
...
-            i->start -= timeOffset;
-            i->stop -= timeOffset;
+            // drag back intervals
+            foreach(RideFileInterval *i, result->intervals()) {
+                i->start -= timeOffset;
+                i->stop -= timeOffset;
+            }
+
+            // drag back xdata
+            QMapIterator<QString,XDataSeries*> it(result->xdata());
```
ihewitt, line 1025: XData datapoints weren't being pulled back so were out of step with the ride datapoints.

ihewitt:
Added the two examples for the stryd files from #2553 into roundtrip to show the working fit file export.

src/FileIO/RideFile.cpp
liversedge: XDATA does not have to share the same km / secs timestamps as the main data- generally XDATA likely comes from a separate source or device.

liversedge:
I will try and find some time to look at this before next weekend (22nd).

ihewitt:
> I will try and find some time to look at this before next weekend (22nd).

no rush... I took years to do my last change! :grinning:

src/FileIO/RideFile.cpp
ihewitt: Updated to just pull back the DEVELOPER XDATA.

liversedge:
@ihewitt so as I see it there were 2 key problems- 1) keeping the fit developer field metadata in the JSON file (to know how to write to a  FIT file) and then 2) writing Fit developer fields to the FIT file in a generic way re-using the metadata if it is available

I will look thru, as I would like to see this done generically - i.e. works for all developer fields not just TCORE.

Before I start reviewing, are there any design decisions you made that you want to highlight ?

ihewitt:
Thanks, just out at the moment! I'll write some notes here when I get back.

ihewitt:
Yes this is mainly about holding and using the developer fields generically now, the core temp implementation is minor and I could remove that from this PR, although it is only a very tiny part of this, and was quite useful to test/check while developing. 
This has removed the storage of TCORE as it's own XDATA field and now uses the generic DEVELOPER fields instead.

I initially tried to use additional XDATA to store the FIT metadata, but then I noticed there was an old piece of code that created a CIQ tag (was line 4715, QString ciqInfo), although this was being created and not used so I updated that and changed my implementation to use that instead. 
I've added a CIQinfo struct to the RideFile to hold the metadata about developer ids and the fields, this is then used to populate the CIQ tag, although uses the Qt JsonWriter/Reader rather than hardcoded string logic.

Notes: (wasn't sure if it's easier to have these here or annotated on the files) 

FitRideFile.cpp: 
- line 1597. It only creates a new name for a developer field if it doesn't already have one.
- line 3085. It always stores developer fields even if they have a native id.
- line 4750, this populates and creates the new CIQinfo/CIQfield structs to store the developer fields
- line 4765, this now uses the qt json writer to create the json for the CIQ tag and writes it into tags escaped.
added the "uint_" writers to support the additional fields in the sample fit files.
- "write_session()" -  removed the unneeded CIQ device headers that were added in the previous coretemp change, these are optional and will need some additional metadata to work properly, I intend to add this back in generically at some point.
- "write_dev_fields()" - changed to be generic using the CIQinfo list, at the moment not all of the fit metadata is used, but now the ciq info is in a json block I intend to add those (e.g. scaling/offset etc). again, at the moment it works fine for the test files and the files I have which don't use these fields.
- line 5440: doesn't support all .fit types yet.

RideFile.cpp:
- line 1022: XData wasn't being 'dragged back' so the data was out of sequence when written back to the FIT, this now drags back the DEVELOPER fields to match the ride datapoints.
- line 1064: reads back the CIQ into the CIQinfo struct list.
- line 2567 & 2824: _(core specific)_ ok since the core temp isn't stored as TCORE in the ride file, if it detects a core_temperature DEVELOPER field then sets the tcore, "data present" and fills the RideFile "tcore" value also.
- 3632 on: this is the new CIQinfo, CIQfield structs.

RideMetaData.cpp:
- Since the CIQ tag is a chunk of json it wasn't displaying nicely in the "Extra" tab text box, changed the box to have 4space tabs and not wrap.

CsvRideFile.cpp:
- _(core specific)_
when a coretemp file has been created during a training session, it now reads the .tcr file and populates the XData DEVELOPER fields as if it was read from a fit file and also adds to the CIQinfo header (line 1641) as if it had been read from a fit file.

On the whole I'm pleased with how this works now, only point I'm a little wary of is the logic to copy the core dev field over to the ridefile in RideFile:2567, I tried to do that adding TCORE to the Json but that was reverted: (https://github.com/GoldenCheetah/GoldenCheetah/pull/4612#issuecomment-2644183472) although I guess the problem is knowing if the data is real or the original derived, so perhaps it's ok.

liversedge:
Hey- thanks for the detailed notes- i will apply and test with other files that contain developer fields. I like the overall approach you describe and will look into it. 

Cheers
Mark

amtriathlon:
Hi, I did some tests and this seems to be working quite well, for example this Garmin 955 with Stryd sensor run file has a lot of developer fields which are correctly restored after FIT export and reimport: https://github.com/GoldenCheetah/GoldenCheetah/blob/master/test/runs/Fr955v19.28andStryd.fit.zip
When and if @liversedge is Ok with the code changes I would like to merge to enable testing in the wild.
Thank you, Ale.

amtriathlon:
I found [Travis CI will stop support for macOS starting March 31st, 2025](https://docs.travis-ci.com/user/reference/osx/), this explain why they didn't update versions from long time ago.
So I would like to generate the last build there this weekend, ideally including this PR.
The code looks fine IMHO and I plan to squash and merge for this purpose, if there are no objections.
Thanks, Ale.

ihewitt:
At the moment this only supports record developer fields, it would be nice
to extend this to lap and session fields also but I haven't had any spare
time to take a look at that. If everyone's happy with the way it stores
json in the extra tags data then I think it's good in the current state.

amtriathlon:
Thank you, Ivor, Record fields for LAP SWIM data would be nice too, perhaps for the next release.

ihewitt:
Brilliant, many thanks.
Yeah I'll definitely take a look as soon as I have some free time.

amtriathlon:
Hi Ivor, I think there is a a problem after this change with native fields other than GoldenCheetah standard ones: before these changes these fields were stored as EXTRA XData fields, but now it seems they are ignored.

The [case reported in the forum](https://groups.google.com/g/golden-cheetah-users/c/hIp8P3CnI-g) happens with Core Temp coming as a native field instead of as a developer field like in your sample file so it is ignores, before these changes it went to EXTRA>FIELD_139. This is an offending file:
[Karoo-Botley_BW_Swanmore_Wickham-2025-05-09-1238.zip](https://github.com/user-attachments/files/20151705/Karoo-Botley_BW_Swanmore_Wickham-2025-05-09-1238.zip)

ihewitt:
bother. I was sure I tested with a direct from device file too. ☹️ I'll
take a look tonight after work.

amtriathlon:
Hi Ivor, I was wrong, the problem is related to the previous PR and limited to Core Temp as native field, solution seems easy too: https://github.com/GoldenCheetah/GoldenCheetah/issues/4643

PR #4613, opened 2025-02-01
Show Distance and speed in mapchart, in low panel (Full Plot)

peret2000:
It adds Speed curve, and also distance along with time in X axis in the lower graph (Full plot check box in map chart).
Perhaps it is too much information, so you can consider if it is useful or useless. I have implemented in case it improves a little the chart, that is a very valuable one.

This is the current version:
Current Version

And with the PR:
PR


amtriathlon:
Hi, yes, I think it is too much and it is unnecessary, considering you can use the full ride chart -perhaps with a distance x-axis- parallel to the map chart, like we do in default Map perspective:

image

PR #4614, opened 2025-02-03
Added manager for named workout filters

thejockl:
* Including a tag-browser
* Changed behaviour of WorkoutFilterBox from editingFinished to returnPressed
  * achieve consistency with SearchFilterBox
  * prevent parallel execution of the update-slot (resulting in segv)


amtriathlon:
Hi Joachim, the hamburger menus looks larger than the one in Search/Filter box, can it be made more alike?
image
image

I don't understand what Obtain does in the manage filter dialog or what select tags does in the menu.

thejockl:
Hi Ale, added a burger-menu-icon with some padding:
image

Select tags give you a tag browser with the option to add them to your current query:
image

Obtain is meant to update a named filter from the WorkoutFilterBox. I use this for example when following a plan and want to update the week without need for adding a new filter. I couldn't find a better name, if you have an idea it would be highly welcome

amtriathlon:
> Hi Ale, added a burger-menu-icon with some padding: image

Much better, thank you!

> Select tags give you a tag browser with the option to add them to your current query: image

Ok, thanks for the explanation.

> Obtain is meant to update a named filter from the WorkoutFilterBox. I use this for example when following a plan and want to update the week without need for adding a new filter. I couldn't find a better name, if you have an idea it would be highly welcome

I think Update would be better, it is also a term for which we already have translations.

amtriathlon:
Thank you!

PR #4616, opened 2025-02-07
Correct incomplete coretemp ANT signal/slot

ihewitt:
Fixes https://github.com/GoldenCheetah/GoldenCheetah/pull/4611#issuecomment-2642897612, incorrect signal signature, sorry about that!


amtriathlon:
Thank you!

PR #4617, opened 2025-02-09
Separate the user and core data processors in the MainWindow menu

paulj49457:
This PR will list the user defined data processors first, and then the core data processors with a separator.

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -2639,10 +2639,26 @@ void MainWindow::onEditMenuAboutToShow()
     toolMapper = new QSignalMapper(this); // maps each option
     connect(toolMapper, &QSignalMapper::mappedString, this, &MainWindow::manualProcess);
 
+    bool userProcDefined = false;
+    // User defined data processors first
     for (QList<DataProcessor*>::iterator iter = processors.begin(); iter != processors.end(); ++iter) {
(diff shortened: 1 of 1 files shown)


amtriathlon:
Why?

paulj49457:
The list of data processors are listed in alphabetical order and depending the names I assign my data processor names they get mixed in with the core data processors, separating them out makes it easier to find them.

amtriathlon:
> @amtriathlon Sorry I missed the new option that has been added to the data processors, which is much better allowing individual selection of the data processors, but maybe the name for the option could be "Hide in Main Window menu" or similar, as initially it wasn't obvious to me:
> 
> image

There are thread on the forum you can use to ask questions and provide feeback: https://groups.google.com/g/golden-cheetah-users/c/U00RH76BJoI/m/4qdQw9adAwAJ and https://groups.google.com/g/golden-cheetah-users/c/DbAiOHgTAFI

BTW, we have been though this more times than I would like to remember:

1. You misunderstand a feature but refuse to use the forum as everyone else
2. Then go ahead and implement a better version according to your understanding
3. That would be fine in your fork, but you pretend it to be incorporated to base code, so you make a PR
4. That would not be that bad, except you almost never accept a rejection and then start to argue you were right in the first place.

No maintainer likes this behavior, as you probably noticed all others except me ignore you from a long time ago, I tried to rescue the value of your contributions until today, now I reached my limit too.

PR #4618, opened 2025-02-17
Zoom property to Live map chart settings

peret2000:
This PR adds an initial zoom for Live map chart in train view. 'Initial' because it can be modified on the fly with the +/- buttons.

It has been discussed previously in [user forum](https://groups.google.com/g/golden-cheetah-users/c/v19Jsjd1Vqk).

You can see the map with your desired zoom. If not set, it defaults to 15, the same as it is now hardcoded.

Settings

It allows a zoom equal to 0. In such case, the zoom is adapted to the boundaries of the route, and the map is not moved when the position changes to be centered on the position. It will show a static view of the route, and the marker will continue to move across the path.

As the zoom can be configured independently for each live mart, you can, for example, have two maps, one showing the general view, and other with a zoom to the current position:

View

### DIFF
--- a/src/Train/LiveMapWebPageWindow.cpp
+++ b/src/Train/LiveMapWebPageWindow.cpp
@@ -82,10 +82,20 @@ LiveMapWebPageWindow::LiveMapWebPageWindow(Context *context) : GcChartWindow(con
     customUrl = new QLineEdit(this);
     customUrl->setFixedWidth(600);
 
+    customZoomLabel = new QLabel(tr("Initial Zoom"));
+    customZoom = new QSpinBox(this);
+    customZoom->setFixedWidth(60);
+    customZoom->setRange(0, 20); // Set the range for zoom levels
+
+
     if (customUrl->text().trimmed().isEmpty()) customUrl->setText("https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png");
     commonLayout->addRow(customUrlLabel, customUrl);
 
     connect(customUrl, SIGNAL(returnPressed()), this, SLOT(userUrl()));
+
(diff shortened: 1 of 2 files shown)


src/Train/LiveMapWebPageWindow.cpp
```diff
@@ -230,13 +240,17 @@ void LiveMapWebPageWindow::telemetryUpdate(RealtimeData rtd)
...
+                code += QString("centerMap (" + sLat + ", " + sLon + ", " + sZoom + ");");
+            }
             code += QString("showMyMarker (" + sLat + ", " + sLon + ");");
             markerIsVisible = true;
         }
         else
         {
-            code += QString("moveMarker (" + sLat + ", " + sLon + ");");
+            (zoom() != 0) ? code += QString("moveMarker (" + sLat + ", " + sLon + ", true);") :
+                code += QString("moveMarker (" + sLat + ", " + sLon + ");");
```
amtriathlon, line 253: It looks weird to use a conditional expression only for its side effects, is better to assign the result of the conditional expression if you want to use one, of use an if statement otherwise.

amtriathlon:
> It allows a zoom equal to 0. In such case, the zoom is adapted to the boundaries of the route, and the map is not moved when the position changes to be centered on the position. It will show a static view of the route, and the marker will continue to move across the path.

I would prefer to remove this option, it is not clear for users and adds an unneeded complexity, better to keep it simple and just allow to configure zoom, IMHO.

peret2000:
Ok, I'll keep it just with the zoom property, with no special case. Tomorrow I will rearrange the code

peret2000:
Reopened

PR #4619, opened 2025-02-18
Add Zoom property to Live map chart

peret2000:
Accidentally closed #4618; this is a replica of it.

This PR adds an initial zoom for Live map chart in train view. 'Initial' because it can be modified on the fly with the +/- buttons.

It has been discussed previously in [user forum](https://groups.google.com/g/golden-cheetah-users/c/v19Jsjd1Vqk).

You can see the map with your desired zoom. If not set, it defaults to 15, the same as it is now hardcoded.

Settings


amtriathlon:
Thank you, does it make sense to set zoom = 0?
Otherwise minimum value should be higher.

amtriathlon:
Ok, thank you!

PR #4620, opened 2025-02-23
Elevation chart as a new widget for train window

peret2000:
This is a implementation of a chart for train view called 'Elevation Chart'.
It shows a short term view of the slope during a GPS workout.
It already exists as a video widget, developed by @human705 ([Zoomed Elevation widget #3793](https://github.com/GoldenCheetah/GoldenCheetah/pull/3793)). When video player is not used (no RLV video), video widgets cannot be used as tiles in train view.

This chart is very based on that one developed by Peter Kanatselis, as an independent chart, integrated with the standard design of charts in the UI.

VideoPullRequest

It shows:
* Graphical view of the altitude profile for the next meters
* Current slope, in a vertical line, indicating current position
* View of slope variation in next seconds through a bubble that changes its color and vertical position (as a floating bubble), according to its value (higher position means bigger change in positive slope)

Configuration data:
Profile Distance (m): meters ahead that are shown in the altitude graph (from 50 to 2000 meters). Additionally, a 20% of that distance is added current position
Seconds for delta slope: the bubble shows the incremet/decrement in slope for next seconds; it is computed as the difference in altitude between current position and position reached in those seconds, considering current speed. From 5 to 600 seconds
Elevation window size (%) relative to current altitude: Vertical size is a percentage of horizontal distance, so that graph can be adapted to the slope profile, and representative of the difficulty. From 5% to 20%

Maybe it shows much information and a more simplified version can be better: for example, just the slope, without the bubble with delta slope indication. Comments are welcome.

It adds two files (ElevationChartWindow.h and ElevationChartWindow.cpp) following the same design for other charts in train view.

This has been previously discussed in [user forums](https://groups.google.com/g/golden-cheetah-users/c/Nt9hXRENlQ0).

### DIFF
--- a/src/Gui/GcWindowRegistry.cpp
+++ b/src/Gui/GcWindowRegistry.cpp
@@ -43,6 +43,7 @@
 #include "MetadataWindow.h"
 #include "TreeMapWindow.h"
 #include "DialWindow.h"
+#include "ElevationChartWindow.h"
 #include "RealtimePlotWindow.h"
 #include "SpinScanPlotWindow.h"
 #include "WorkoutPlotWindow.h"
@@ -69,7 +70,7 @@ GcWindowRegistry* GcWindows;
 void
 GcWindowRegistry::initialize()
 {
-  static GcWindowRegistry GcWindowsInit[34] = {
+  static GcWindowRegistry GcWindowsInit[35] = {
     // name                     GcWinID
     { VIEW_TRENDS|VIEW_DIARY, tr("Season Overview"),GcWindowTypes::OverviewTrends },
     { VIEW_TRENDS|VIEW_DIARY, tr("Blank Overview "),GcWindowTypes::OverviewTrendsBlank },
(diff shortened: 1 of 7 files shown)


amtriathlon:
Thank you, @peret2000, it looks ok for me but I am not a user of Train, I would like to have feedback from @thejockl before merging.

thejockl:
Looks good to me - and I like the functionality. Thanks @peret2000!

peret2000:
Thanks to both for the feedback. Happy to have this chart added in the next release. Should you want any change, code refactorization, etc, I will do it.

peret2000:
Once it is merged, I should add some documentation to [train chart types](https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_ChartTypes_Train) in the wiki

amtriathlon:
There is a problem reported at the users forum (https://groups.google.com/g/golden-cheetah-users/c/qEB8ic5F1A8/m/OmjnuDwbBAAJ) indicating this chart is calling locationAt/altitudeAt even when there is no location/altitude data s.t. when using ERG workouts.

peret2000:
> There is a problem reported at the users forum (https://groups.google.com/g/golden-cheetah-users/c/qEB8ic5F1A8/m/OmjnuDwbBAAJ) indicating this chart is calling locationAt/altitudeAt even when there is no location/altitude data s.t. when using ERG workouts.

Sorry the mistake. Thanks for the quick detection and correction

PR #4621, opened 2025-02-24
Limit chart settings size

amtriathlon:
Currently it is too tall for lower resolution displays, as discussed in the users forum, the size limit in AthleteTab has no effect so it is moved to ChartSettings and adjusted. The tallest seems to be CP chart settings.
I tested on Windows (Qt 5.15.2 and 6.5.3), Linux (Qt 6.5.3), and macOS (Qt 6.7.3).

### DIFF
--- a/src/Charts/ChartSettings.cpp
+++ b/src/Charts/ChartSettings.cpp
@@ -18,6 +18,7 @@
  */
 
 #include "ChartSettings.h"
+#include "Colors.h"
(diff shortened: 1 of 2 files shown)


amtriathlon:
I am closing this since current height is necessary for CP chart settings, the blank space in Trends charts config is misleading.
If we want to make chart settings less tall we would need to change the layout or make the contents scrollable, IMHO.

PR #4622, opened 2025-03-06
Correction to 8eb96b2d6 (ElevationChartWidget): isHidden() is always 'false'

peret2000:
Using `!isVisible()` instead behaves as expected, to avoid updating elevation chart widget in training being updated when it is not in the selected layout


amtriathlon:
Thank you!

PR #4624, opened 2025-03-08
Optimize computation/draw of widgets and charts during training workouts, and Fix some train interface issues

peret2000:
It has been commented in [developers forum](https://groups.google.com/g/golden-cheetah-developers/c/A84_QVmUsKw/m/s6VromcWAgAJ). 

- When swicthing the layout to a one without video, video widgets remained visible, overlapping the layout. This is very annoying when starting a training session with no video: odd number of starts show videow widgets overlapped (in particular, the first start)
This is due to: all widgets of all train layouts are created when slecting train view, none of the widgets are hidden when they are not necessary, and the odd number of starts required is due to previous point
- Stops recomputing or redraw of video widgets or charts when their layout is not active. That avoids taking care of having as many layouts with many charts as desired, as they will not be updated continuously (only parts that compute averages, etc will continue to do their job)

I cannot test it with VideoPlayer, although changes do not mind video engine used.

### DIFF
--- a/src/Train/VideoWindow.cpp
+++ b/src/Train/VideoWindow.cpp
@@ -565,6 +565,9 @@ void VideoWindow::resumePlayback()
 
 void VideoWindow::telemetryUpdate(RealtimeData rtd)
 {
+    if (!isVisible())
+        return;
+
     Lock lock(stateLock);
     if (!hasActiveVideo())
         return;
--- a/src/Train/VideoWindow.cpp
+++ b/src/Train/VideoWindow.cpp
@@ -55,7 +55,7 @@ VideoWindow::VideoWindow(Context *context)  :
     videoSyncDistanceAdjustFactor = 1.;
     videoSyncTimeAdjustFactor = 1.;
 
-    curPosition = 1;
+    curPosition = 1; 
 
     state = PlaybackState::None;
 
@@ -303,7 +303,8 @@ void VideoWindow::showMeters()
         p_meterWidget->AdjustSizePos();
         p_meterWidget->update();
         p_meterWidget->raise();
-        p_meterWidget->show();
+        if (isVisible())
+            p_meterWidget->show();
         p_meterWidget->startPlayback(context);
     }
     prevPosition = mapToGlobal(pos());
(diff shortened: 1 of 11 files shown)


src/Train/VideoWindow.cpp
```diff
@@ -55,7 +55,7 @@ VideoWindow::VideoWindow(Context *context)  :
     videoSyncDistanceAdjustFactor = 1.;
     videoSyncTimeAdjustFactor = 1.;
 
-    curPosition = 1;
+    curPosition = 1;
```
amtriathlon, line 58: Don't do this, please

src/Train/WorkoutWidget.cpp
```diff
@@ -114,6 +114,12 @@ WorkoutWidget::WorkoutWidget(WorkoutWindow *parent, Context *context) :
     configChanged(CONFIG_APPEARANCE);
 }
 
+void WorkoutWidget::update()
+{
+    if (isVisible())
+        QWidget::update();
+}
+
```
amtriathlon, line 122: According to the docs QWidget::update does nothing when the widget is hidden: https://doc.qt.io/qt-5/qwidget.html#update, so is this change necessary?

src/Train/VideoWindow.cpp
peret2000: Sure! sorry, I will remove that blank along with next comment. Awaiting to your opinion, just to know if I make changes or not

src/Train/WorkoutWidget.cpp
peret2000: That's true. The thing is that isHidden() was returning 'false' always in the case of the elevation chart, and that caused to do a PR to substitute isHidden() by isVisible(). [https://doc.qt.io/qt-5/qwidget.html#isHidden](https://doc.qt.io/qt-5/qwidget.html#isHidden) does not solve my doubts, so I did it just in case. I cannot debug update(), as it belongs to Qt libraries.
As a workaround test, I have added temporary code 'bool checkhidden = IsHidden();' instead of the one you are commenting; the value is always 'false', no matter if the layout is visible or not.
Anyway, let me know your opinion, I can remove it if you want.

amtriathlon: IsHidden is different from !isVisible, as explained in https://doc.qt.io/qt-6/qwidget.html#isHidden, but I would not assume QWidget::update is wrong without proof, IMO this change should be removed.

src/Train/WorkoutWidget.cpp
```diff
@@ -311,6 +318,8 @@ WorkoutWidget::eventFilter(QObject *obj, QEvent *event)
 {
     // process as normal if not one of ours
     if (obj != this) return false;
+    if (!isVisible())
+        return false;
```
amtriathlon, line 316: Is this a fix for a real issue or a "just in case" change?
I mean, the events processed here are mouse and keyboard, so I don't see in which scenario these may arrive when the chart is not visible.

amtriathlon:
Thank you!

PR #4625, opened 2025-03-14
ActionButtons: Showing +- labels with Qt 6.8.2

thejockl:
* With Qt 6.8.2 button-labels are not shown if widget is scaled too small
* Using ActionButtonBox throughout
* Set padding to 0px for +- buttons in ActionButtonBox
* Generalized ActionButtonBox to work with QAbstractItemView

See also https://groups.google.com/g/golden-cheetah-users/c/R-58f1LvwsM


amtriathlon:
Thank you!

PR #4627, opened 2025-03-16
Italian translations

pierg75:
Added a bunch of missing Italian translations.


amtriathlon:
Than you very much, Pier!

PR #4628, opened 2025-03-27
Use speed data from imported GPX files

mrmekon:
When importing a GPX file, prefer embedded speed data (if present) over GPS coordinates for calculating speed and distance.

The GPX file format supports providing the measured speed at each trackpoint as a TrackPointExtension. Some programs use this to report readings from dedicated speed sensors.

If no speed data is embedded or if data is missing from any trackpoints, Golden Cheetah calculates speed from the GPS coordinates. When using GPS, weak signal or low-quality receivers can result in irregular speed measurements due to location jitter, whereas dedicated speed sensors should be consistent in any environment.

Distance is additionally calculated from speed and time if available, falling back to GPS otherwise.

### DIFF
--- a/src/FileIO/GpxParser.cpp
+++ b/src/FileIO/GpxParser.cpp
@@ -46,6 +46,7 @@ GpxParser::GpxParser (RideFile* rideFile)
     lon = 0;
     lat = 0;
     hr = 0;
+    speed = std::numeric_limits<double>::infinity();
     temp = RideFile::NA;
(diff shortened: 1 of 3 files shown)


mrmekon:
Justification and implementation choices discussed [here](https://groups.google.com/g/golden-cheetah-users/c/BclmCJNPzFw)

amtriathlon:
Thank you!

PR #4633, opened 2025-04-18
Ensure the Device Type Tile reflects changes on edit

paulj49457:
Editing the Device Type overview tile doesn't immediately update the tile or register a change in the ride file.


amtriathlon:
Thank you!

PR #4636, opened 2025-04-18
Prevent GC saving the perspectives twice

paulj49457:
The saveState() call results in the loaded perspectives being saved twice when GC exits, once in saveState() and again when the AbstractView's destructor is called.


amtriathlon:
Thank you.

PR #4645, opened 2025-05-21
Options->Fields page improvements

paulj49457:
I have been looking at the Options->Fields page, and have created a PR to:

1) Display whether a field is a User, Special or Metric field in a new Category column so that it is easy to confirm that overridden field names are correct, the Category column is auto populated depending on the text in the Field Column, 

2) The field type is now auto populated for Special & Metric fields, this ensures they are correct.

3) I have also grouped the display aspects (screen tab, summary, interval)
**Edit**: After searching the code, the "interval" checkbox means that there is a value for each of the activities intervals, rather than a single value for the entire activity, if correct probably something to add to the Wiki. Therefore this field is now probably in the wrong place.

4) Removed the unexplained expressions field, as it is displayed and set in the Options->Fields page, stored in the FieldDefinition class, but never used in the GC code.

Data Fields Wiki reference: https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Preferences_Data-Fields

New appearance below, I'm happy to change any aspect of this PR, the column ordering might not be correct, maybe "category" isn't the right title.

Data fields


amtriathlon:
> I have been looking at the Options->Fields page, and have created a PR to:
> 
> 1. Display whether a field is a User, Special or Metric field in a new Category column so that it is easy to confirm that overridden field names are correct, the Category column is auto populated depending on the text in the Field Column,

I don't like this change, I don't think it helps at all. If you want to make metric overrides easier the way to go would be to add the edit button to Metric tiles, as you did for Metadata tiles, IMHO.

> 2. The field type is now auto populated for Special & Metric fields, this ensures they are correct.

This makes sense.

> 3. I have also grouped the display aspects (screen tab, summary, interval)

Please don't do that, it will be rejected.

>    **Edit**: After searching the code, the "interval" checkbox means that there is a value for each of the activities intervals, rather than a single value for the entire activity, if correct probably something to add to the Wiki. Therefore this field is now probably in the wrong place.
> 4. Removed the unexplained expressions field, as it is displayed and set in the Options->Fields page, stored in the FieldDefinition class, but never used in the GC code.

Interval Metadata was partially implemented by https://github.com/GoldenCheetah/GoldenCheetah/commit/1dd7faa2762ca7edfc702f4f6c2491b00cd95ffa, we need to document what is working and what is pendind, and add to Release Notes.

paulj49457:
Thank you for your feedback, I have updated the PR as follows, and will look to create another PR for editable metric tiles:

1) The expression field has been reinstated, and its title has been changed to "Interval expression", as its only relevant when the interval checkbox is set.

3) I have restored the column order

4) I have retained the additional field type column (special, metadata & user) as I think this is useful for the user to understand why certain fields cannot be changed, the fields automatically corrected are:

              Special:
                  The Type column is auto populated
                  The following columns are automatically made blank or unset: Values, Interval, Interval expression.

              Metadata:
                  The Type column is auto populated
                  The following column is automatically made blank or unset: Values.

New appearance below, I'm happy to change any aspect of this PR, can you confirm the automatic corrections in 4 above are correct?

Data fields - 2

amtriathlon:
Hi Paul, I gave it a try and I don't like how it looks like, beside trying to add a new field give me crash. I will not merge this, sorry.

PR #4651, opened 2025-05-30
Calculating expected load also based on past stress

thejockl:
Additional:
* Allow charts with planned load
* Initialize RideItem::planned to false (was: uninitialized)


amtriathlon:
Hi Joachim, the changes seems ok to me, adding cogganPlannedXX is consistent with existing PMC metrics and we can include it for now, but those are old versions used before we have a general mechanism for creating PMC derivatives using any base metric. I have a proposal to extend that mechanism to include planned/expected as options and to get rid of all these special cases.

PR #4655, opened 2025-06-09
Remove Qt deprecation warnings

paulj49457:
Note: This PR has been initially pushed so that I can test the changes compile for other platforms (I only have a Windows platform).

All failing due to Qt versions <6.8.x
- Windows: Using Qt version 5.15.2 in C:/Qt/5.15/msvc2019_64/lib
- Mac: Using Qt version 6.5.3 in /Users/appveyor/Qt/6.5/macos/lib
- Linux: /home/appveyor/Qt/5.15/gcc_64/bin/qmake

The PR aims to remove some of the following warnings:
- Deprecation warnings now GC is using Qt6.8.x, access for points has changed.
- qAsConst should be replaced with std::as_const
- QScopedPointer should be replaced with std::unique_ptr
- addAction constructor has changed, earlier versions are deprecated
- 'QtPrivate::warnIfContainerIsNotShared': Do not use foreach/Q_FOREACH with containers which are not implicitly shared. Prefer using a range-based for loop with these containers: `for (const auto &it : container)`,
- Unused parameters, usually due to platform dependent compilation
- printf formatting type mismatches

There are also other Qt warnings which might be worth looking at in the future:

warning C4996: 'QDateTime::QDateTime': Pass QTimeZone instead
warning C4996: 'QDateTime::setTimeSpec': Use setTimeZone() instead
Charts\OverviewItems.cpp:
warning C4996: 'QChart::setAxisX': was declared deprecated
warning C4996: 'QChart::axisY': was declared deprecated
warning C4996: 'QChart::axisY': was declared deprecated

### DIFF
--- a/contrib/kmeans/kmeans.cpp
+++ b/contrib/kmeans/kmeans.cpp
@@ -30,7 +30,7 @@ void Kmeans::free() {
 
 
 
-void Kmeans::initialize(Dataset const *aX, unsigned short aK, unsigned short *initialAssignment, int aNumThreads) {
+void Kmeans::initialize(Dataset const *aX, unsigned short aK, unsigned short *initialAssignment, [[maybe_unused]] int aNumThreads) {
     free();
 
     converged = false;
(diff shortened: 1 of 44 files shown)


amtriathlon:
We still support Qt 5.15 so, if you want to remove deprecation warnings, you need to use supported alternatives or to use conditional compilation. Additionally you can enable AppVeyor on your fork, no need to make a PR just to test compilation...

PR #4656, opened 2025-06-14
Reworked the dialog for manually creating activities

thejockl:
* Transformed the dialog into a wizard
* Only allowing activities in the past
* Considering the sport for estimations and shown metrics
* Added TriScore and SwimScore
* Removed Daniel Points, Notes and Average Speed
* See https://groups.google.com/g/golden-cheetah-users/c/Yz8g2J1Ue6w/m/6u0UAfSRAgAJ

### DIFF
--- a/src/FileIO/LapsEditor.cpp
+++ b/src/FileIO/LapsEditor.cpp
@@ -24,111 +24,101 @@
 #include "HelpWhatsThis.h"
 #include "Context.h"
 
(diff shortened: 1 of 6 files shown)


amtriathlon:
Hi Joachim, since you renamed ManualRideDialog class I think the source file names should be renamed too.

amtriathlon:
Thank you!

PR #4660, opened 2025-06-17
Single instance of SpecialFields

paulj49457:
Whist working on the editable Metric Tile, I have noticed the following issues with SpecialFields:

Currently the SpecialFields object is created within GlobalContext, and it is updated when user metrics/user metadata items are created/removed, this is managed by GlobalContext & RideCache. GlobalContext is a singleton, and the specialFields parameter also needs to be a singleton (a single point of metric field truth), but SpecialFields class isn't a singleton and separate instances can and are created in the code.

Detected issues:

i) The RideMetadata & BatchProcessingDialog classes created their own instance of the SpecialFields objects and held on to them, this will prevent them getting updated by add/deleted user metrics/metadata items when the GlobalContext instance of SpecialFields is updated.
ii) Local instances of the SpecialFields were created and destroyed numerous times as local function variables.
ii) Each creation/deletion of the SpecialFields object would cause its model pointer to leak memory, even though the model is never used within GC, there doesn't appear to be a delete matching the new.
iii) The FieldType enum, and fieldType() accessor function (declared but never implemented) are not used, they seem to have been replaced by the isUser(), isSpecial() & isMetric() functions.

This PR has therefore aims to:

i) Ensure that SpecialFields (and SepcialTabs for consistency) are singletons, and GlobalContext & RideCache can reload its configuration.
ii) GlobalContext not longer holds a SpecialFields parameter, SpecialFields is a single globally available object.
ii) The unused model and accessor have been removed.
iii) The unused and unimplemented fieldType() and associated enum have been removed.
iv) All SpecialFields accessor functions now use const QString& parameter types.


amtriathlon:
Thank you!

PR #4637, opened 2025-04-19
Remove view specifics from Abstract View

paulj49457:
Remove view specifics from Abstract View, if this is PR accepted/merged then another small PR can follow to remove the remaining references to specific views.

PR #4661, opened 2025-06-25
Remove the remaining view specifics from Abstract View - pt2

paulj49457:
Remove the remaining view specifics from Abstract View, this PR follows on from #4637.


amtriathlon:
Thank you!

Issue #4665, opened 2025-07-02
Improve support for CORE sensor


fellrnr:
GC has support for displaying Core temperature from a CORE sensor in a dial window. I'd like to add support for displaying Skin Temperature and Heat Strain Index (HSI). I'd also like to add support for graphing those three values. 

I've found that HSI is far more useful than core temperature when heat training. Skin temperature is also useful as it will lead changes in HSI and core temperature. A graph is also valuable as it shows the rate of change more clearly. 

I've contacted CORE to ask how to calculate Heat Load, as that would be useful to see in real time. 

Here's a photo of how things look in my fork:

![Image](https://github.com/user-attachments/assets/270cdd4e-0e10-49dd-9e9e-638975aa535a)

I raised in the GC developers forum as https://groups.google.com/u/1/g/golden-cheetah-developers/c/vtYtC2J2ZfU

fellrnr:
Uploaded patch to improve core support. 

- Add dial window (telemetry chart) for skin temp, heat strain index, and estimated heat load
- Calculate an estimate of heat load. This is different from the propitiatory CORE calculation, but seems good enough to give a useful indication. Heat load resets at midnight local time. (An option to input a starting value for heat load would be a follow on update.)
- Add option to plot skin temperature and heat strain index
- Modify robot to give random values for core temp, skin temp, and HSI for testing. 

[0001-Improve-support-for-CORE-sensor-Fixes-4665.patch](https://github.com/user-attachments/files/21183993/0001-Improve-support-for-CORE-sensor-Fixes-4665.patch)

PR #4668, opened 2025-07-12
Improve support for CORE sensor

fellrnr:
This is to resolve the issue "Improve support for CORE sensor #4665"

- Add dial window (telemetry chart) for skin temp, heat strain index, and estimated heat load
- Calculate an estimate of heat load. This is different from the propitiatory CORE calculation, but seems good enough to give a useful indication. Heat load resets at midnight local time. (An option to input a starting value for heat load would be a follow on update.)
- Add option to plot skin temperature and heat strain index
- Modify robot to give random values for core temp, skin temp, and HSI for testing.
-

### DIFF
--- a/src/Train/NullController.cpp
+++ b/src/Train/NullController.cpp
@@ -24,15 +24,28 @@
 #include "RealtimeData.h"
 #include "PhysicsUtility.h"
 
+#define CORE_START 37.5
+#define CORE_UPPER_LIMIT 40.0
+#define CORE_LOWER_LIMIT 35.0
+
+#define SKIN_START 35.0
+#define SKIN_UPPER_LIMIT 40.0
+#define SKIN_LOWER_LIMIT 30.0
+
+#define HSI_START 2.0
+#define HSI_UPPER_LIMIT 5.0
+#define HSI_LOWER_LIMIT 0
+
+
 NullController::NullController(TrainSidebar *parent,
                                DeviceConfiguration *dc)
   : RealtimeController(parent, dc), parent(parent), load(100),
     bicycle(NULL)
 {
-  //set values before start as heat 
-  this->core = 37.5;
-  this->skin = 37.0;
-  this->heatStrain = 2.0;  
+  //set values before start as heat
+  core = 37.5;
+  skin = 37.0;
+  heatStrain = 2.0;
 }
 
 int NullController::start() {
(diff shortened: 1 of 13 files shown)


src/Train/NullController.cpp
```diff
@@ -74,6 +78,26 @@ void NullController::getRealtimeData(RealtimeData &rtData) {
     rtData.setCadence(85 + ((rand()%10)-5));
     rtData.setHr(145 + ((rand()%3)-2));
     rtData.setHb(35 + ((rand()%30)), 11 + (double(rand()%100) * 0.01f));
+
+    rtData.setCoreTemp(this->core, this->skin, this->heatStrain);
+
+    //randomize robot temp for testing
+    this->core += ((rand()%4) - 2) /10.0;
+    if(this->core < 37)
+      this->core = 37.5;
```
paulj49457, line 87: Given 37.5 is used on line 33, 87 & 92, these values should be #define DEFAULT_CORE_TEMP 37.5 to ensure they track, also the limits could be #defined

src/Train/NullController.cpp
```diff
@@ -29,6 +29,10 @@ NullController::NullController(TrainSidebar *parent,
   : RealtimeController(parent, dc), parent(parent), load(100),
     bicycle(NULL)
 {
+  //set values before start as heat 
+  this->core = 37.5;
+  this->skin = 37.0;
```
paulj49457, line 34: Why is this defaulted to 37.0, when the default on line 92 is 37.5 ?

src/Train/NullController.cpp
fellrnr: I was following what had been done for other test values of using raw numbers. E.g.:

    rtData.setCadence(85 + ((rand()%10)-5));
    rtData.setHr(145 + ((rand()%3)-2));
    rtData.setHb(35 + ((rand()%30)), 11 + (double(rand()%100) * 0.01f));

We could tidy up the robot, using defines and making the behaviour more realistic, but I'm not sure how much it's used.

src/Train/NullController.cpp
fellrnr: Line 92 (etc.) isn't so much of a default as pushing the random walk back towards reasonable values. We could tidy this all up if it's useful.

fellrnr:
I found an uninitialized variable bug in src/Train/WorkoutWidget.cpp - line 194 added 
ctempSum = stempSum = hsiSum = 0;

I've fixed this and added the tidy ups that @paulj49457 suggested above. 

I struggled with git for a couple of days before I got the commits into this pull request, so I'm not sure if I got this right, but it looks okay to me.

amtriathlon:
Thank you!

PR #4669, opened 2025-07-12
Updated gc-blank.png image

ZajtiM:
Updated gc-blank.png image, with the same desing as GC Icon and Splash. It is displayed in Live Map chart in Train view.

**OLD:**
Old_GC-Blank

**NEW:**
New_GC-Blank


amtriathlon:
Thank you, Mitja, the .png in images-src is different or we can remove it to avoid duplicates?

ZajtiM:
Can be removed.

amtriathlon:
Merged, thank you!

PR #4674, opened 2025-07-20
Add the override classifier for some overridden functions

dmaslenko:
To fix the `inconsistent-missing-override` compilation warning:
> "'xxx' overrides a member function but is not marked 'override'".


amtriathlon:
Thank you!

PR #4676, opened 2025-07-22
Fix warnings about override and fields initialization order

dmaslenko:
### What
- Fixes compiler warnings related to the order of member initialization in constructor initializer lists. The order of initialization is now aligned with the order those fields are declared in their corresponding header files, preventing warnings about fields being initialized out of order.
- A member function was explicitly marked with override to clarify that it overrides a base class method, which resolves another common compiler warning.
 
No logic or functional behavior of the application was changed; only warning cleanups were performed.


amtriathlon:
Thank you!

PR #4677, opened 2025-07-26
Fix compilation warnings

dmaslenko:
### Description
The PR contains few commits with fixed compilation warnings.

The fixes are made without business logic changing except of the commit where the abandoned `statusCode` variable was reused in the logs, as result the log message was changed correspondingly.


amtriathlon:
Thank you!

PR #4678, opened 2025-07-27
Fix warnings: unused variable and diff types for print function

dmaslenko:
### What
- Removed or commented the local variables that never used.
- Reused `%lld` to print long long integer.

### Why
To fix the compilation warnings like:
- `variable '***' set but not used`
- `unused variable '***'`
- `format specifies type 'int' but the argument has type 'qsizetype'`


amtriathlon:
Thank you!

PR #4679, opened 2025-07-27
New monthly calendar

thejockl:
* Created new Trends-Chart "Planning Calendar"
* Added supporting tools to Colors
* Added notification about changed season
* Updated ManualActivityWizard to optionally take the date as parameter
* Added some new icons for the calendar


src/Charts/PlanningCalendarWindow.cpp
```diff
@@ -0,0 +1,387 @@
...
+            || rideItem == nullptr) {
+            continue;
+        }
+        if (context->isfiltered && ! context->filters.contains(rideItem->fileName)) {
+            continue;
+        }
+
+        QString route = rideItem->getText("Route", "").trimmed();
+        QString workoutCode = rideItem->getText("Workout Code", "").trimmed();
+        QString sport = rideItem->getText("Sport", "").trimmed();
```
amtriathlon, line 214: It is better to use rideItem->sport since it will get the normalized sport name avoiding the need to deal with translations or alternative names s.t. Cycling.

thejockl: Fixed in https://github.com/GoldenCheetah/GoldenCheetah/pull/4679/commits/05ffc90fcdec5dae64206c9e33e78aaf4360cfc0

paulj49457:
I use the SubSport field to define the activity (road bike, MTB, turbo, etc) and I don't use the WorkoutCode field, so is it possible to have the SubSport available when a planned activity is created ?

thejockl:
> I use the SubSport field to define the activity (road bike, MTB, turbo, etc) and I don't use the WorkoutCode field, so is it possible to have the SubSport available when a planned activity is created ?

I like that idea - and it is only a minimal change: Just remove https://github.com/GoldenCheetah/GoldenCheetah/blob/81ba70ee3f06684d4839ad828d4a556782bc8d0d/src/Gui/ManualActivityWizard.cpp#L321 and the line after

Do you want to create the PR?

paulj49457:
> > I use the SubSport field to define the activity (road bike, MTB, turbo, etc) and I don't use the WorkoutCode field, so is it possible to have the SubSport available when a planned activity is created ?
> 
> I like that idea - and it is only a minimal change: Just remove
> 
> https://github.com/GoldenCheetah/GoldenCheetah/blob/81ba70ee3f06684d4839ad828d4a556782bc8d0d/src/Gui/ManualActivityWizard.cpp#L321
> 
> and the line after
> Do you want to create the PR?

Would you mind updating this PR with the change ?

thejockl:
> Would you mind updating this PR with the change ?

Done

amtriathlon:
Thank you!

PR #4681, opened 2025-07-28
Correction to configuration path override in AbstractView  pt2

paulj49457:
The getPathToPerspectiveFile() virtual function added in pt2 to the AbstractView class doesn't work as intended :(  but this doesn't impact GC operation.

But, for views that have a different path to the standard athlete one (https://github.com/paulj49457/GoldenCheetah/wiki/Equipment-Management-Feature) , getPathToPerspectiveFile() returns the overridden path upon loading, but when saving the overriding EquipmentView class has been destroyed when saveState() is called in AbstractView destructor and the standard athlete path is incorrectly returned.

This correction doesn't impact GC functionality, but provides the intended mechanism to extend the view for derived classes.

### DIFF
--- a/src/Gui/AbstractView.h
+++ b/src/Gui/AbstractView.h
@@ -54,7 +54,7 @@ class AbstractView : public QWidget
 
     public:
 
-        AbstractView(Context* context, int type, const QString& view, const QString& heading);
+        AbstractView(Context *context, int type, const QString& view, const QString& heading);
         virtual ~AbstractView();
         virtual void close() {};
 


amtriathlon:
When I merge your PR I thought getPathToPerspectiveFile() virtual function was overkill but harmless, now you want to use that for views to set it to different folders for reasons I don’t understand considering views should not store data, only configuration information about how to display data…

paulj49457:
The aim of getPathToPerspectiveFile() was to allow extension without modification of AbstractView. The equipment management feature has a single perspective shared by all athletes and is stored in gcroot, and the addition of getPathToPerspectiveFile was to support this extension, this correction PR achieves the same goal but works correctly, I'm sorry for the oversight on the destructor virtual call in saveState.

Regarding the equipment management features storage of data in the perspective, this was an expedient way to initially implement the functionality, then as the PR developed more data was being held in the tiles, I agree it needs to change. So I intend to refactor this in the Autumn, so the equipment-perspective.xml will just hold the tile's presentation information, and a separate xml file will hold the user entered data, history, notes, etc.

amtriathlon:
So you will not need this or virtual getPathToPerspectiveFile() and we can go back to the simpler version using context->athlete->home->config().canonicalPath() as the path for abstract view. If a view needs access to gcroot it can use GlobalContext.

paulj49457:
I have updated the PR to simplify it, whilst still allowing classes derived from AbstractView to modify the perspective's path.

This means I can override the viewCfgPath in my EquipmentView classes constructor, and load/store the equipment view's single shared perspective from the gcroot path location.

If this remains unacceptable, let me know, and I'll update the PR to restore the perspective path code to the original version.

amtriathlon:
Thank you!

PR #4684, opened 2025-08-04
Metric override icon incorrectly displayed in Trends Overview

paulj49457:
The Metric tile's override icon is incorrectly displayed in Trends Overview, clicking the pencil icon on a Trends Metric Tile doesn't display the metric override menu (correctly I think), so in this case the pencil edit icon should be hidden. It doesn't make sense to override metrics whose displayed value is based upon date ranges and/or activity filters.


amtriathlon:
Thank you!

PR #4686, opened 2025-08-17
Show workout for planned activity in TrainView

thejockl:
New option in context menu for planned activities in AnalysisSidebar to show matching workouts in TrainView


amtriathlon:
Hi Joachim, it seems FLOAT is a typedef in MSVC tool chain triggering: `error C2365: 'FLOAT': redefinition; previous definition was 'typedef'`, the same happens with WORD, so we would need to use different tokens.
DataFilter parser uses DF_FLOAT and Json parser JS_FLOAT, so WF_FLOAT/WF_WORD could follow the style.
Thank you, Ale.

amtriathlon:
Thank you!

PR #4689, opened 2025-08-19
Update FITmetadata.json

dconka:
- Changed Sigmasport to SIGMA, as the URL changed from sigmasport.com to sigma.bike
- Added new SIGMA ROX 4.0 SE devices


amtriathlon:
FITmetadata.json is a generated file as documented in the file itself, please don't manually update.
See https://github.com/GoldenCheetah/GoldenCheetah/tree/master/util/fit

PR #4692, opened 2025-08-24
New dialog to repeat planned workouts

thejockl:
* New Wizard to select planned workouts based on past timerange
* Implemented simple conflict resolution strategies (remove all preexisting, skip days with preexisting, fail)
* Create new planned activities based on the selected, past ones
* Additional: Added forgotten GPL-Header to Calendar.cpp|h

### DIFF
--- a/src/Gui/RepeatScheduleWizard.cpp
+++ b/src/Gui/RepeatScheduleWizard.cpp
@@ -274,7 +274,7 @@ RepeatSchedulePageActivities::initializePage
     QDate endDate = field("endDate").toDate();
     QLocale locale;
     numSelected = 0;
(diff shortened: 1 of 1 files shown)


src/Gui/RepeatScheduleWizard.h
```diff
@@ -0,0 +1,117 @@
...
+#include <QLabel>
+#include <QTreeWidget>
+
+
+class RepeatScheduleWizard : public QWizard
+{
+    Q_OBJECT
+
+    public:
+        enum {
```
paulj49457, line 36: Can this be moved outside of the class and made an enum class? This would provide a type (rather than int) for the overridden (int nextId() const override;) functions and require an explicit cast to set values other than those defined.

PR #4694, opened 2025-08-27
New "completerval" filter function

paulj49457:
Updated:

When changing auto completer values or checking the integrity of activities, I occasionally find values in metadata fields which are not one of the defined auto completer values for the field, finding these is a pain. So I have created a completervalues() function which takes a user metadata field name and returns the metadata auto completer values for that field.

For example to check all the Sport fields have one of the auto completer values set, the following filter will display all the activities with incorrect Sport metadata values 

`!(completervalues("Sport") contains Sport)`

If the metadata field has no auto completer values, or a field other than a user metadata field is provided the function returns an empty string.

As usual I'm happy to modify as required, and will update the wiki if this gets merged.

I wasn't sure whether the vector version would be useful so removed it in the second commit, as a vector of string values didn't seem to be useful or meaningful, see the DataFilter-safe.cpp file in the first commit for this functionality.


amtriathlon:
Hi Paul, this seems too specific for your use case, perhaps is better to keep in your fork.
Something more general could be to have a function to get the possible values of a field as a vector of strings and then use contains to check in your use case.

paulj49457:
Thanks for the feedback.

I have updated the PR in the latest commit so completervalues() returns the list of completer values, I have updated the PR description above. If it gets merged, I'll then update the Wiki.

PR #4698, opened 2025-09-05
Display perspective switch expression.

paulj49457:
To assist in checking perspective switch expressions and their position in the perspective auto switching search order, I thought it useful to display the switch expression next to the perspective name:

Updated:

**Activities view**:
image

**Trends view**:
image

**Train View**:
image


amtriathlon:
Beware "switch expression" applies only to Analysis view, for remaining views the expression has other name and meaning.

paulj49457:
I have updated the PR to support the trends (activities filter) and train (switch for) perspective variants, and the PR description above.

Also fixed an error when adding a train perspective, the switch for field was ignored, so the perspective "switch for" always had the value "Don't switch", although it could be corrected later by editing.

amtriathlon:
Thank you!

PR #4701, opened 2025-09-07
Running speed read from Footpod ANT+ device

peret2000:
It is read if the footpod does not send cadence or stride length is not configured.
For such, stridelength is not initially set to 115 cm, but to 0, so if it is not read from the device, it means the device is not able to use it.
Also, if cadence is not captured by the device, also the speed cannot be computed.
In any of those situations, speed can be read from the device


amtriathlon:
Thank you!
I think we should remove the following line to allow stridelength=0 as a valid configuration:
https://github.com/GoldenCheetah/GoldenCheetah/blob/ed929366aaee62f9a1c08b5afeccbadf30d8fdd0/src/Train/DeviceConfiguration.cpp#L112

PR #4704, opened 2025-09-15
Added a day view to PlanningCalendarWindow

thejockl:
* Reusing the existing configuration
* Linked from days in month view to the more detailed view
* Fixed color setting after construction


amtriathlon:
Thank you!

PR #4709, opened 2025-09-27
Additional line for entries in calendar day-view

thejockl:
* Calendar Day-View now supports showing any field as tertiary line (good use: Notes)
* Additional fix: Rewriting metric name (secondary line); now BikeScore^TM is shown instead of BikeScore&8482;


amtriathlon:
Thank you!

PR #4713, opened 2025-10-13
The initial analysis view's perspective is incorrect when no items are imported

paulj49457:
The initial analysis view's perspective is incorrect when no items are imported, this is caused by the following call sequence (this is summary of the main calls relevant to this problem):
Note: mainWindow::resetPerspective is re-entrant in the start-up sequence but protect by its guards
 

```
mainWindow::selectAnalysis
    mainWindow::resetPerspective sets pactive = true
    |     AnalysisView::setPerspectives
    |          on first load it calls AnalysisView::setViewSpecificPerspective()
    |                this then called AnalysisView::setRide()
    |                      this then called mainWindow::switchPerspective
    |                           but as pactive == true, it didn't apply the perspective change      :(
    |  pactive = false
```

the proposed change removes setRide from this initial startup sequence, as AnalysisView::setViewSpecificPerspective only gets called during initial view loading, so using perspectiveSelected with the correct perspective achieves the right result, as the ride has already been set at this point. 

```
mainWindow::selectAnalysis
    mainWindow::resetPerspective sets pactive = true
    |     AbstractView::setPerspectives
    |           on first load it calls AbstractView::perspectiveSelected() with perspective from
    |                  AnalysisView::getViewSpecificPerspective, this sets the ride's perspective correctly,
    |                                                                               and the charts & controls 
    |  pactive = false
```

### DIFF
--- a/src/Gui/AbstractView.cpp
+++ b/src/Gui/AbstractView.cpp
@@ -714,11 +714,9 @@ AbstractView::setPerspectives(QComboBox *perspectiveSelector, bool selectChart)
     if (!loaded || selectChart) {
         loaded = true;
 
-        // generally we just go to the first perspective
-        perspectiveSelected(0);
-
-        // allow views to override the default perspective (if required)
-        setViewSpecificPerspective();
+        // generally we just go to the first perspective, but we allow
+        // the views to override the default perspective (if required)
+        perspectiveSelected(getViewSpecificPerspective());
 
         // due to visibility optimisation we need to force the first tab to be selected in tab mode
         if (perspective_->currentStyle == 0 && perspective_->charts.count()) perspective_->tabSelected(0);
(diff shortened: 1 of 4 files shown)


amtriathlon:
I don't use auto-import so I am used to this behavior, to the point I consider it a "feature": the starting view is independent of the selected activity and there is where I have a Web chart for Garmin Connect download, but I recon it is more likely a bug.
I will test this for a few days.

amtriathlon:
I looked at the code and I didn't like the changes. Likely we will change the initial view before release since I expect we will have better options than the last activity when there are no new activities imported, so for now lets keep this without changes.
PS: I created [a new issue]( https://github.com/GoldenCheetah/GoldenCheetah/issues/4717) for tracking this.

PR #4714, opened 2025-10-13
Trends activity click thru doesn't set the correct perspective for the activity in Analysis View

paulj49457:
The trends view activity click thru doesn't set the correct perspective for the activity in Analysis View, this is because the view hasn't been updated before the ride is set in Analysis view and its guard in setRide() rejects it, as:
context->mainWindow->athleteTab()->currentView() == 1 is not true, the current view is still trends.

### DIFF
--- a/src/Gui/AthleteTab.cpp
+++ b/src/Gui/AthleteTab.cpp
@@ -187,6 +187,8 @@ AthleteTab::selectView(int index)
 void
 AthleteTab::rideSelected(RideItem*)
 {
+    emit rideItemSelected(context->ride);
+
     // if we selected a ride we should be on the analysis
     // view-- this is new with the overview and click thru
     // coming in other charts but when navigation model is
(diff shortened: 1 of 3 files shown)


amtriathlon:
It looks like this change fixes that issue but it creates a new one: to go back to Trends you need to click the back button twice.

paulj49457:
Unfortunately I didn't test the back buttons as I don't use them, I'll take a look at this PR to fix the back button issue.

paulj49457:
@amtriathlon I have fixed the back double click issue, it required a slight re-ordering of the calls in AthleteTab::rideSelected().

I have also added/changed whilst fixing this PR:
I have added an optionally complied stackDebug() function.
Protected the "recent" activities list which is not currently used, maybe this should/could be removed?
Made the action() function private to the NavigationModel.

amtriathlon:
> @amtriathlon I have fixed the back double click issue, it required a slight re-ordering of the calls in AthleteTab::rideSelected().

It seems to work, thank you!
 
> I have also added/changed whilst fixing this PR: I have added an optionally complied stackDebug() function. Protected the "recent" activities list which is not currently used, maybe this should/could be removed? Made the action() function private to the NavigationModel.

let’s keep the changes focused, I will not merge this part since it is unnecessary for the problem this PR intends to fix.

Issue #4717, opened 2025-10-16
Make the starting view configurable


amtriathlon:
Currently the program starts in Activities view with the last activity selected in default perspective ([switch expressions are not honored in this case](https://github.com/GoldenCheetah/GoldenCheetah/pull/4713)). I think we could make the starting view configurable: for daily issue starting in the new calendar chart or other Trends chart may be more useful than looking at the details of the last activity and installations dedicated to trainer control may choose to start in Train view.
This is independent of autoimport/autosync happening afterwards.

paulj49457:
See #4718 for a possible solution.

amtriathlon:
Fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/9f3b8bc101e23c2a1dc29d9cfcb40f1f967aa14d

PR #4718, opened 2025-10-20
Selectable initial startup window

paulj49457:
The following PR addresses issue #4717 providing a user option in Options->General->Application Behaviour:

startup view

The navigation model buttons work correctly, the only known minor issue is the first click thru from trends view will display the ride in the analysis view with the default perspective.

Also (which can be removed if required) I have hidden the default auto import complete dialog when the view is not analysis, as I don't see its relevance to train or trends view, but requires the use to acknowledge it.

### DIFF
--- a/src/Gui/AbstractView.h
+++ b/src/Gui/AbstractView.h
@@ -184,8 +184,6 @@ class AbstractView : public QWidget
 
         bool loaded;
 
-        void setViewsInitialRide();
-
         // Support view specific behaviour
         virtual void notifyViewStateRestored();
         virtual void notifyViewPerspectiveAdded(Perspective* page);


amtriathlon:
> Also (which can be removed if required) I have hidden the default auto import complete dialog when the view is not analysis, as I don't see its relevance to train or trends view, but requires the use to acknowledge it.

Please remove that, autoimport is independent of the starting view selected.

paulj49457:
Import dialog changes have been removed in the latest commit.

amtriathlon:
There is a weird behavior after merging this PR and selecting Trends as the initial view: when switching to Activities view a "*" first column is added to Activities list which was not present before:
image
After some use it goes back to normal:
image
but I couldn't identify a pattern yet.

ZajtiM:
Hi Ale.

Added column with * is some old behavior. I noticed it before, with official releases, not some local builds or snapshots. I can reproduce it when i open new / second athlete.

Here is gif which shows it. But if i click on it, it disappears.

Activities_list_sidepanel


You can even see it with default Athlete when you open GC. But that one disappears on its own. Here is another gif. At the beginning is just a black screen, because GC is starting (didn't show MainWindow yet).

Activities_list_sidepanel_2

paulj49457:
I've been away for a few days, I'll take a look at the "*" issue, I had not seen this on my existing test athlete configuration.

paulj49457:
See #4722 for a solution to the issue regarding the hidden asterisk group by field being displayed.

amtriathlon:
I found another issue similar to the above: when you start in Trends view no User Chart works until you change to Activities View and back, it also happens when you open a second athlete which also starts in Trends view, so it is not caused by this changes, but  we need to fix that before merging.

paulj49457:
@amtriathlon Thank you for the testing, I'll take a look at the issues you have found.

paulj49457:
@amtriathlon I have investigated the issue of the UserCharts being empty when Trends is the startup view... The problem is that in this scenario there is no selected rideItem (which always the case when Activities is the startup view, and then a rideItem is always selected for the remaining GC operation), the UserChart needs a selected ride to function correctly (the evaluation processing) even though its in rangemode.

I looked at refactoring the UserChart to remove this dependency but that's not straightforward.

**Latest commit:**

So, I have overridden AbstractViews notifyViewStateRestored() function in Trends view and set an initial ride using identical logic to activities, and the UserChart is now poplulated. Starting in Train and switching to Trends works, the UserCharts are populated.

I made a minor update to the navigation model to ensure the correct rideItem was updated in the transition between analysis and trends, and back again.

I have refactored the click through and initial analysis rideItem perspective logic, this works and selects the correct perspective for the rideItem activity.

Can you please test? I'll continue testing, but haven't seen anything wrong so far....

**Future discussions/work:**

It would be nice if the ride selected in Activities was the calendar day displayed when the user switches to the planning calendar in the trends view, at the moment it defaults to todays date, using the selected activities date would seem more consistent when switching between views?

amtriathlon:
> @amtriathlon I have investigated the issue of the UserCharts being empty when Trends is the startup view... The problem is that in this scenario there is no selected rideItem (which always the case when Activities is the startup view, and then a rideItem is always selected for the remaining GC operation), the UserChart needs a selected ride to function correctly (the evaluation processing) even though its in rangemode.
> 
> I looked at refactoring the UserChart to remove this dependency but that's not straightforward.

Fully agree, I added a minimalist fix for additional athletes in https://github.com/GoldenCheetah/GoldenCheetah/commit/f63773f0971886863c2a36f4448167f568763bed

> **Latest commit:**
> 
> So, I have overridden AbstractViews notifyViewStateRestored() function in Trends view and set an initial ride using identical logic to activities, and the UserChart is now poplulated. Starting in Train and switching to Trends works, the UserCharts are populated.
> 
> I made a minor update to the navigation model to ensure the correct rideItem was updated in the transition between analysis and trends, and back again.
> 
> I have refactored the click through and initial analysis rideItem perspective logic, this works and selects the correct perspective for the rideItem activity.
> 
> Can you please test? I'll continue testing, but haven't seen anything wrong so far....

I will test in the next few days
 
> **Future discussions/work:**
> 
> It would be nice if the ride selected in Activities was the calendar day displayed when the user switches to the planning calendar in the trends view, at the moment it defaults to todays date, using the selected activities date would seem more consistent when switching between views?

I am not sure, but I think it is something to discuss on the Calendar thread at the users forum.

amtriathlon:
It works for me now, I will merge this, thank you!

PR #4722, opened 2025-10-26
Fix ride navigator asterisk column problem for additional athletes

paulj49457:
The problem of the hidden asterisk column (used for groupby) appearing when additional athletes were created was due in part to there no being an ActiveWindow event being generated for the additional athletes, so this has been changed to LayoutRequest event, also the hidden asterisk column was not always set to zero width.

Test steps:

1. Open GC
2. Switch to athletes view
3. Either create or open an additional athlete
4. Switch to activities view
5. Asterisk displayed without this fix.

Switching between Athletes using the tabbar, opening and closing athletes all now work, without the hidden groupby column being displayed.


amtriathlon:
Thank you!

PR #4726, opened 2025-11-03
New Options Processor Tab

paulj49457:
I keep failing to remember that the core & python processors are a tab under Options->Data Fields->Processors & Automation, possibly as I don't see the relationship between data fields configuration and the core & python processors. Therefore I propose the Processors & Automation are moved to a new Options->Processors level tab to make them more prominent and easier to find, I'm sure someone more artist could improve the icon.

image


amtriathlon:
The rationale was to group all Data configuration and manipulation options in one place, like Data contains what before were separate charts (Details and Editor), but you can keep this in your fork if you prefer this way.

PR #4727, opened 2025-11-08
Calendar: Agenda View

thejockl:
* Read-only view to show
  * missed planned activities (configurable number of days to look back)
  * todays planned activities, phases and events
  * upcoming planned activities, phases and events (configurable number of days to look ahead)
* Color of planned activities is now user configurable (coming from global color theme)
* Added a hint to show whether a filter is active (all calendar views)
* Configuration: Reorganized in tabs
* Refactoring: Created enums for all user roles used in the calendar (all views, all delegates) to improve readability; removed data that was only written but never read


amtriathlon:
Thank you!

PR #4728, opened 2025-11-08
Using MultiMetricSelector in FavouriteMetricsPage

thejockl:
As the same type of metrics selector was needed for the settings of the PlanningCalendarView, a reusable widget MultiMetricSelector was created in #4679, duplicating code from FavouriteMetricsPage.

This change elminates the duplication and uses the MultiMetricSelector also in the global settings.

A user visible change for FavouriteMetricsPage is the new filter field for available metrics.


amtriathlon:
Thank you!

PR #4729, opened 2025-11-09
Bugfix (Calendar): Parallel activities ending after midnight

thejockl:
When having two parallel activities ending after midnight, they were not layouted correctly in day- and week view of the calendar. This change fixes that, showing them side by side and not overlapping any more.


amtriathlon:
Thank you!

PR #4730, opened 2025-11-09
Consistent ride navigator selected activity and chart display

paulj49457:
This PR addresses the following closely associated analysis view issues:

1) When a search/filter is applied which returns no results, charts are displayed (with contents) in the analysis view for the selected activity, even though the ride navigator is empty. **PR: will now display a blank chart with message "No results from search/filter"**

2) When a search/filter is applied which returns results but the currently selected activity is not in the results, charts are displayed (with contents) in the analysis view for the selected activity, even though the ride navigator cannot show or highlight it. **PR: will now display a blank chart with message "The selected activity is not in the search/filter results, Select a new activity OR cancel the search/filter"**

3) When a search/filter is applied which returns results but the currently selected activity is not in the results, if an activity is now selected in the ride navigator, the overview tiles still display information for the previous selected activity, not the new activity.
**PR: will now display the correct information for the selected activity**

4) When a search/filter is applied and the currently selected activity is within the results, its highlight is lost. **PR: will reapply the highlight for the selected activity**

**The changes are as follows:**

RideNavigator.[cpp|h] 
Updated the searchStrings & clearSearch functions to reapply the selected ride/activity highlight, fixes point 4 above.

GcWindowRegistry.[cpp|h]
Ensured windowsForType(int type) and idsForType(int type) are part of GcWindowRegistry class so they will link.
Removed the fixed size of GcWindowsInit[], as this will prevent future sizing issues, it has a sentinel as an iteration limit

ChartSpace.[h|cpp]
Updated the showEvent to update the space's tiles
Updated rideselected guard to remember the selected ride, as it may have changed
These fix point 3 above.

AnalysisSideBar.cpp
Created the RideNavigator using the GcWindowRegistry::newGcWindow function as this ensures the chart type (GcWindowTypes::ActivityNavigator) is set for the RideNavigator - this is required to prevent RideNavigator being blanked, for points 1 & 2.

GoldenCheetah.[h|cpp]
Added the concept of search/filter chart blanking which overrides the existing chart blanking mechanism when a search/filter is active, and reverts to the existing logic in the absence of a search/filter.
Added new default virtual function bool selectedRideInfo() to GcChartWindow class, so charts requiring the selected ride/activity can indicate this to support the search/filter chart blanking - for points 1 & 2 above.
Added new blank page for point 2 above to inform the user why no data is displayed on the chart.
The following unused items have been removed to prevent confusion:
QLayout*_blankLayout; attribute
setBlankLayout(QLayout *layout) function

**To test:**

1) Switch to analysis view, and to an overview chart.
2) Set a filter which returns no activities, Sport = "", the analysis charts are blanked, with a message that there are no results
3) Switch to trends, confirm the charts are displayed, although if related to activities they will be empty
4) Switch back to analysis
5) Set a filter for Running activities, Sport = "Run", Select any running activity
6) Set a filter for cycling activities, Sport = "Bike", the chart is now blanked with a message to select an activity in the results, or cancel the search/filter
7) Select a bike activity from the ride navigator, confirm the charts display the correct information.

I'm happy to adjust the PR if the behaviour I have implemented for points 1-4 isn't right, the search/filter chart blanking behaviour only applies to those charts overriding selectedRideInfo() and returning true i,e, those requiring a selected ride.


paulj49457:
I'm now thinking that rather then applying this logic to only those charts assigned to the VIEW_ANALYSIS, the base GCWindowClass should have a protected Boolean (and simple public accessor function) indicating whether the chart relies upon a selected activity. Each chart requiring the search/filter blanking logic could then set this flag true in its constructor, as it is the chart's implementation that governs whether the selected activity is required or not, rather than which views it can be assigned to.

paulj49457:
Latest commit implements the idea in the comment above, this allows more targeted blanking decided by each individual chart rather than on a per view basis.

amtriathlon:
Hi Paul, I really dislike this change, initially because is fixes a non-issue and clarifies a confusion nobody had until now, but more important, because it breaks click-trhu behavior from other views.

paulj49457:
Hi Ale, I'm confused, I've just tested the click thru from the Calendar view and it works, I then set a global filter, and again click thru works.  Would you mind telling me under which conditions it fails ?

amtriathlon:
I mean, this could make sense if the only way to set the activity for Activities view chart were to select from Activities list, but since it can also happens by click-thru from others view, this make no sense.
What happens if you click-thru from Trends or Plan view and the target activity is not one of the activities on the filtered activities list?

paulj49457:
When I tested it setting the filter in the activities view, it also applies to the plan & trends calendars, so I don't think it is possible to click thru on an activity that is not currently displayed in ride navigator with the current functionality.

But I see your point, the proposed PR https://github.com/GoldenCheetah/GoldenCheetah/pull/4811 is likely to break that assumption, assuming a Sport="Bike" filter was applied, and a ride navigator "completed" option selected, clicking through from the calendar on a planned bike ride (which would be displayed) would be an issue, I guess the only option is to update this PR after 4811 to then switch the ride navigator filter to "All" to ensure the selected activity can be displayed in the navigator.

Can we put this PR on hold until 4811 (or similar) is merged, I'll then take another look and update it.

amtriathlon:
Let’s put this to rest, I don’t see a significative advantage for such a complex change, better to focus on pending issues a new features.

PR #4731, opened 2025-11-10
Mini Calendar on non-analysis start-up view fix

paulj49457:
This PR fixes the mini calendar issue raise by @ZajtiM in the calendar discussions:

I noticed one problem with Calendar in Analysis view. It doesn't show month with latest activity. You can see on screenshot. You have to click on activity to show month with selected activity. I tested it with latest Snapshot and it works as it should. This is screenshot of GC (AppImage) from your latest PR.

image

This also happens when opening a second athlete, it looks like some mini calendar initialization is not happening when Activities is not the starting view.

### DIFF
--- a/src/Gui/Views.cpp
+++ b/src/Gui/Views.cpp
@@ -68,10 +68,10 @@ AnalysisView::~AnalysisView()
 void
 AnalysisView::setRide(RideItem *ride)
 {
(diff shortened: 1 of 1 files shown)


amtriathlon:
Thank you!

PR #4733, opened 2025-11-14
Calendar: Additionally respecting homefilters

thejockl:
When collecting the activities for the planning calendar, only the filter was used. Now the homefilter is considered as well.

See https://groups.google.com/g/golden-cheetah-developers/c/viSV80ze3r8/m/lbkJnG1jCAAJ


amtriathlon:
Thank you!

PR #4734, opened 2025-11-16
Default Plan view startup auto import issue

paulj49457:
**Issue raise by Poncho**: when the calendar is used as default view and is opened immediately after program start, new activities autoimported in the background only appear after switching to another view. For me it would be convenient if a successful autoimport would refresh the calendar automatically.

The auto import not updating the calendar is probably due to the import containing more than 20 activities  ( Poncho can you confirm this is the issue? or provide more information on how to repeat the issue? )  , as I have tested it with fewer activities and it updates fine, as each activity is added it is updated in the calendar

The issue is trying to import >20 activities, which RideImportWizard.cpp considers to be a batch import and then it doesn't signal in the RideCache that a ride has been added, see  https://github.com/GoldenCheetah/GoldenCheetah/blob/a7200a31ff54f1f8a9cc8bf765fe699a02654638/src/Gui/RideImportWizard.cpp#L1121  and   https://github.com/GoldenCheetah/GoldenCheetah/blob/a7200a31ff54f1f8a9cc8bf765fe699a02654638/src/Core/RideCache.cpp#L302

### DIFF
--- a/src/Charts/CalendarWindow.cpp
+++ b/src/Charts/CalendarWindow.cpp
@@ -62,6 +62,7 @@ CalendarWindow::CalendarWindow(Context *context)
     });
     connect(context, &Context::filterChanged, this, &CalendarWindow::updateActivities);
     connect(context, &Context::homeFilterChanged, this, &CalendarWindow::updateActivities);
+    connect(context, &Context::autoImportCompleted, this, &CalendarWindow::updateActivities);
     connect(context, &Context::rideAdded, this, &CalendarWindow::updateActivitiesIfInRange);
     connect(context, &Context::rideDeleted, this, &CalendarWindow::updateActivitiesIfInRange);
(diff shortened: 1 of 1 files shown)


paulj49457:
I have seen occurrences of the issue raised by Poncho, but cannot identify why, so maybe this PR is required.

Reducing the number of notified rides on batch import, or removing all notifications and just having a single batch import notification at the end will speed up import and reduce visual updates.

amtriathlon:
IIRC the user confirmed it was not this case, so I would wait to have a better diagnostic before to implement a solution, I will close this for now.

paulj49457:
This PR has been reopened now we understand why the calendar wasn't being updated, it occurs when the user has cleared their import folder, meaning that each auto import has >20 files, and therefore the individual ride update events are suppressed, and the calendar doesn't get an event to update its display.

amtriathlon:
Thank you!

PR #4739, opened 2025-12-01
New calendar chart color issue

paulj49457:
When manually creating a new calendar chart (with default Modern Dark color scheme, windows 11, qt 6.8.3) the new chart appears with a light background, while existing calendar charts loaded from the perspective's file appear correctly.

This appears to be caused (from google search) by the Perspective's style sheets interacting with the local calendar palette settings when the chart is added to the Perspectives QStackWidget:

[tabbed->addWidget(newone);](https://github.com/GoldenCheetah/GoldenCheetah/blob/9e8ed1717af5ced83cf787e80bfbb93db55cb1b3/src/Gui/Perspective.cpp#L742C13-L742C19)

causing the appearance to be lost. I confirmed this my calling the calendar's configChanged() before and after this line, before the chart has a light background, after the line the chart appearance is correct.

The fix is to reapply the palette settings to the calendar's widgets on the chart's showEvent.

I'm not sure why this is an issue for the calendar chart, as other charts use QPalette settings, but the calendar uses palette options (active, inactive & disabled) not used by any other chart:

        palette.setColor(QPalette::Active, QPalette::Window, activeWindow);
        palette.setColor(QPalette::Inactive, QPalette::Window, activeWindow);
        palette.setColor(QPalette::Disabled, QPalette::Window, alternateBg);

### DIFF
--- a/src/Charts/CalendarWindow.cpp
+++ b/src/Charts/CalendarWindow.cpp
@@ -161,6 +161,14 @@ CalendarWindow::CalendarWindow(Context *context)
     });
 }
 
+void
+CalendarWindow::showEvent(QShowEvent*)
+{
+    // When the chart is added to the Perspective's QStackedWidget, the Perspective's style sheets
+    // can conflict with the local calendar palette settings causing the appearance to be lost.
+    // Therefore re-apply the palette upon showEvent to ensure correct calendar appearance.
+    PaletteApplier::setPaletteRecursively(this, palette, true);
+}
 
 int
 CalendarWindow::getDefaultView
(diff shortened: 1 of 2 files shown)


amtriathlon:
I have this issue in my normal config when switching to Modern Dark in current master:

image

It looks ok after merging this PR:

image

@thejockl are you ok with this change?

thejockl:
I will have a look this weekend (currently off sick)

thejockl:
@paulj49457 thanks for the fix, this is the correct fix for that issue. I will apply a similar patch to the AgendaWindow (currently under development)

amtriathlon:
Thank you!

Issue #4705, opened 2025-09-17
Trends charts - isolate axis hides labels when using bars


amtriathlon:
When hovering the cursor over a y-axis isolates correctly the related curves but, when they are bars all labels disappear. This was [reported at the users forum](https://groups.google.com/g/golden-cheetah-users/c/zl1suWdXqXQ) and it is likely due to this hack not working in this case:

https://github.com/GoldenCheetah/GoldenCheetah/blob/3d7858b17f9bc4eeedcd02312e2467732a9ff3dd/src/Charts/AllPlot.h#L234-L238

paulj49457:
https://github.com/GoldenCheetah/GoldenCheetah/pull/4741 fixes this issue.

PR #4741, opened 2025-12-07
Fix for isolate axis hides labels when using bars

paulj49457:
Fixes https://github.com/GoldenCheetah/GoldenCheetah/issues/4705

The pen wasn't set for bar charts, and the hack relies upon the pen's color value to set the visibility of the labels.


amtriathlon:
Thank you!

Issue #4100, opened 2021-12-07
Re-implement Chart DB


liversedge:
Need to move away from GCP and reimplement on self-hosted basis.
Will move the release date to End of January to allow this to complete since its a big change.

liversedge:
Lets make this more extensible, so we can use it for more things in the future:
1. Sharing Charts
2. Sharing Perspectives (Sets of Charts)
3. Sharing Metrics
4. Sharing Workout Files (ERG et al)
5. Sharing Training Plans
6. Sharing Trainer Config (Virtual Curves)

liversedge:
If we create a backend API there is nothing to stop us creating a website to browse it for non-GC users. But that will likely come later.

liversedge:
Moving to 3.7, it will take too long to implement and better to do alongside v3.7. Moving the release date back to the end of the year.

gille:
I've sent a PR to upgrade ChartDB to most recent Go. The core idea is still VERY good, but could likely use some work & help.

https://github.com/GoldenCheetah/CloudDB/pull/2

PR #4742, opened 2025-12-08
Check for the latest version of GC

paulj49457:
The following PR checks the user's installed GC application is the latest version every 31 days, and reminds them to upgrade if they are using an old version. The new checking functionality downloads the GCUpgrade.h file from the GoldenCheetah Github project, and compares the versions with those within the users application. This means there is no additional overhead for the build release process.

An example of the dialog is below:

image

**Questions:**

- If you prefer accessing the version information from the goldencheetah.org website, let me know and I'll update the PR? I searched the internet for the supported file request rate on Github, and got the following response: "A free GitHub account is subject to a rate limit of 60 requests per hour per IP address when accessing public raw files without an authentication token. This limit applies to direct access via raw.githubusercontent.com." so the PRs Github use will not cause any issues with each active GC instance making a single request every 31 days.
- I think this now replaces the CloudDBVersionClient classes informUserAboutLatestVersions() functionality (I have commented out the function call), or is that to check the DB version?
- Is every 31 days ok?
- The check only retries every 31 days, if the connection fails when attempting to get GCUpgrade.h it will retry in 31 days, is this ok?

**To test**
set the date test in gcVersionCheck() to true, and set the following in GCUpgrade.h to old versions:
#define VERSION_LATEST 5001
#define VERSION_STRING "V3.6"

### DIFF
--- a/src/Core/GcUpgrade.h
+++ b/src/Core/GcUpgrade.h
@@ -131,6 +131,7 @@
 #define VERSION31_BUILD VERSION31_UPG
 
 // the next two will with each build/release
+// note: the GC version check functionality in mainWindow is dependent upon format of VERSION_STRING
 #define VERSION_LATEST 5006
 #define VERSION_STRING "V3.7.1"
 //#define GC_VERSION VERSION_STRING // To force version string on non-tagged ci builds
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -756,7 +756,7 @@ MainWindow::MainWindow(const QDir &home)
     }
 
     versionClient = new CloudDBVersionClient();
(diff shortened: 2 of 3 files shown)


amtriathlon:
Hi Paul, @gille is working to revamp and keep CloudDB up and going so we may not need this after all, see https://github.com/GoldenCheetah/GoldenCheetah/issues/4100, otherwise I think to preserve current UX including its translations would be better than start from scratch.

paulj49457:
Hi Ale, I'll watch the progress of https://github.com/GoldenCheetah/GoldenCheetah/issues/4100, if this PR is required I will update it to use the existing UX including its translations.

amtriathlon:
Ok, let’s close this for now.

PR #4743, opened 2025-12-09
#else constexpr doesn't do anything it's just a compiler warning and doesn't look like a useful comment either.

gille:


amtriathlon:
Thank you!

PR #4746, opened 2025-12-09
Improve the build speed by 2-3X by using PCH. This avoids rebuilding …

gille:
…large parts of QT and C++ std headers over and over. It's possible to make more things faster by better modularizing of GC but that's quite a bit more work and will fracture the internals so I don't think we should explore that first.

Old build: https://ci.appveyor.com/project/Joern-R/goldencheetah-knhd8/builds/53087382
New build: https://ci.appveyor.com/project/Joern-R/goldencheetah-knhd8/builds/53213525

Old ones time out, new ones finish in ~48 min

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -221,14 +221,13 @@ build_script:
 
 # Windows
 - cmd: qmake.exe build.pro -r -spec win32-msvc
-- cmd: if not exist qwt\lib\qwt.lib jom -j4 sub-qwt
-# QWT has issues with parallell builds.
-- cmd: jom sub-src
+- cmd: if not exist qwt\lib\qwt.lib jom -j1 sub-qwt
+- cmd: jom -j2 sub-src
 
 # Linux / macOS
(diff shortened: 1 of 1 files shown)


amtriathlon:
Hi Magnus, this is great, building Linux and Windows binaries in less than an hour on AppVeyor is great!
On Linux we were on the rivet of the extended 1h30 and on Windows it never finished after upgrading to Qt6, even when caching QWT, that's the reason the opengl32sw.dll error was not detected.

gille:
The slow builds were driving me crazy while playing with a few features so, unfortunately to fix MacOS we need to port to CMake which.. while doable (I'm happy to do it) is going to take a bit more work and I don't think we should embark on it before we drop QT5.

amtriathlon:
Currently macOS builds on AppVeyor are well within the 1h30 limit, so we would be ok for now.
When or after merging this PR I will revert https://github.com/GoldenCheetah/GoldenCheetah/commit/fa84a37a5384a295ca64305ff9f7becef1dd1ab7 and remove the `& exit /b 1` from https://github.com/GoldenCheetah/GoldenCheetah/blob/06ed87db09d5714835d145c8033be166e65d9ecc/appveyor.yml#L224 since it is annoying to have to manually retry the jobs after changing appveyor.yml, it was just a workaround to ensure QWT was cached before the 1h30 limit on Linux and Windows.

gille:
I don't know why but this last run was slooooooooow on Mac, I'll see faster builds are IMO always better, but it's also maybe not worth hunting too much. I've got bigger fish that I intend to fry.

amtriathlon:
Thank you!

Issue #4541, opened 2024-09-12
Upgrade Python-SIP embedding


amtriathlon:
Currently we are using Python 3.7 with SIP 4.19.8 in our builds and both are deprecated.
SIP v4 module is not available for current Python versions, so we will need to upgrade, ideally to v6, but this will likely involve extensive source code changes and testing, since they are not compatible, I am not sure if v5 could be an useful intermediate step.
In the meantime we could build using Python 3.10 including the built from source SIP 4.19.8 module in embedded Python, but users willing to use a separate Python install would need to build SIP from source themselves.

amtriathlon:
Nobody is currently working on this, AFAIK, but we accept patches. See [Developers Guide](https://github.com/GoldenCheetah/GoldenCheetah/wiki/Developers-guide) if you plan to contribute.

tirons:
Can I ask why no one is working on it? Is the Python integration not used much? I'm new to GC and trying to absorb as much as I can. I certainly don't want to work on an unimportant but labour intensive port.

amtriathlon:
> Can I ask why no one is working on it? Is the Python integration not used much? 

We have a users forum and a developers forum, if you have questions please use them.

> I'm new to GC and trying to absorb as much as I can.

I suggest you to start reading the wiki, as explained in https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Special-Topics_Working-with-Python#working-with-python you can use embedded Python from our binary releases.

PR #4747, opened 2025-12-11
Upgrade Python, Pandas, and SIP versions. Make the builds future proo…

gille:
…f(er).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -146,19 +146,16 @@ install:
 - cmd: echo DEFINES+=GC_WANT_PYTHON >> src\gcconfig.pri
 - cmd: echo PYTHONINCLUDES=Core c:\Python311-x64\include >> src\gcconfig.pri
 - cmd: echo PYTHONLIBS=-L\"c:\Python311-x64\libs\" -lpython311 >> src\gcconfig.pri
+# Upgrade pip to ensure you have the latest version
+- cmd: python -m pip install --upgrade pip
+# Install your project's dependencies from a requirements.txt file
+- cmd: "python -m pip install --only-binary :all: -r src/Python/requirements.txt -t C:/Python/lib/site-packages"
 # GSL
 - cmd: echo GSL_INCLUDES=c:\tools\vcpkg\installed\x64-windows\include >> src\gcconfig.pri
 - cmd: echo GSL_LIBS=-Lc:\tools\vcpkg\installed\x64-windows\lib -lgsl -lgslcblas >> src\gcconfig.pri
 
 # Linux / macOS - install dependencies first (including Python 3.11)
 - sh: bash appveyor/$OS_NAME/install.sh
-
-# Upgrade pip to ensure you have the latest version
-- cmd: python -m pip install --upgrade pip
-- sh: if ! $CI_LINUX; then export PATH="/usr/local/opt/python@${PYTHON_VERSION}/bin:$PATH" && python3 -m pip install --upgrade pip; fi
-# Install your project's dependencies from a requirements.txt file for Windows and macOS, for Linux it happens in after_build script.
-- cmd: "python -m pip install --only-binary :all: -r src/Python/requirements.txt -t C:/Python/lib/site-packages"
-- sh: if ! $CI_LINUX; then export PATH="/usr/local/opt/python@${PYTHON_VERSION}/bin:$PATH" && python3 -m pip install -r src/Python/requirements.txt; fi
 
 before_build:
 
--- a/appveyor/macos/after_build.sh
+++ b/appveyor/macos/after_build.sh
@@ -17,7 +17,7 @@ cp `brew --prefix icu4c`/lib/libicudata.*.dylib GoldenCheetah.app/Contents/Frame
 
 # Copy python framework and change permissions to fix paths
 # Note: The framework already contains site-packages with numpy, sip, etc.
-# because appveyor.yml runs "pip install -r requirements.txt" which installs
+# because install.sh runs "pip install -r requirements.txt" which installs
 # packages into the brew Python's framework site-packages.
 echo "Copying Python Framework from ${BREW_PYTHON_FRAMEWORK}"
 # Remove any old attempts to avoid link confusion
--- a/appveyor/macos/before_build.sh
+++ b/appveyor/macos/before_build.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ev
 
+export PATH="/usr/local/opt/python@${PYTHON_VERSION}/bin:$PATH"
+
 # Get config
 cp qwt/qwtconfig.pri.in qwt/qwtconfig.pri
 cp src/gcconfig.pri.in src/gcconfig.pri
--- a/appveyor/macos/install.sh
+++ b/appveyor/macos/install.sh
@@ -48,8 +48,9 @@ sudo cp D2XX/libftd2xx.1.4.24.dylib /usr/local/lib
 brew install python@${PYTHON_VERSION}
 export PATH="/usr/local/opt/python@${PYTHON_VERSION}/bin:$PATH"
 python3 --version
-
-# Note: Python packages are installed in appveyor.yml via pip install -r requirements.txt
-# They go into the brew Python framework's site-packages, which is copied in after_build.sh
+# Upgrade pip to ensure you have the latest version
+python3 -m pip install --upgrade pip
+# Install your project's dependencies from a requirements.txt file
+python3 -m pip install -r src/Python/requirements.txt
 
 exit


peret2000:
I am testing this PR and works perfectly, I am not having any issues. The development environment is much simpler to arrange, and there are no restrictions in the versions of python and its packages.
Tested on Ubuntu 24.04 for developent and running on ubuntu 25.10.
Using python 3.12, and last pandas (2.3.3), sip (6.15.0) and numpy (2.2.6) packages.
Correct behavior with data processors on importing, and all the python charts in use. I have only had to change some few calls to methods of pandas related with its change of version, as expected.
Grat job, Thanks a lot! this removes the stuck in python, sip, pandas, and all that python stuff

gille:
Thank you! Yes I tried to clean up and make it easier to change things for the next time we need to lock/move things.

The only remaining things is CI which unfortunately takes quite a while to loop on :(

amtriathlon:
Hi Magnus, I really appreciate your work on this it was a long time delayed pending to fix.
AppVeyor turnaround can be improved if each developer enables a free account for his fork, time will be limited to 1h instead of 1h30 so it may not be enough in this case, but it is better than to share a single instance in PRs when you just want to do some checks without clogging the shared instance.

gille:
I did create my own account but whenever I clicked to build it just says "contact support" (which I finally got around to). To speed the builds up I tried to run one type until it worked. I'll see if/when they get back to me.

gille:
The build succeeded on Mac! The previous ones had already passed. I unfortunately do not have a Mac to test on but the Linux build worked fine when I tried it.

MeisterP:
Maybe a minor issue... but the automatic configuration for gsl and python you're introducing in `src/gcconfig.pri` doesn't work in conjunction with `PKGCONFIG`.
Basically, using `PKGCONFIG = gsl python3-embed` only works if I remove all the new automatic configuration from this pull request beforehand.

gille:
:/ do you have the error? Otherwise I'll have to bring up yet-another-vm to test this on (worst case I'll follow up with the fix ASAP)

MeisterP:
My mistake when merging your pull request locally. Sorry for the noise. It's working correctly 👍

Thanks for your work. Updating python is much appreciated.

gille:
:/ it did build nicely on Windows the time before. I'm putting this failure into the Appveyor is flaky bucket.

amtriathlon:
AppVeyor performance sometimes varies wildly and I have to manually re-run incomplete jobs, but this doesn't seem to be the problem in this case since I did that more than 5 times in different hour, last time tonight, and it never finished while master build completed in about 1hr for the Windows jobs, likely there is something else more systematic, see https://ci.appveyor.com/project/Joern-R/goldencheetah-knhd8/history

image

gille:
Hmmmm... let me dig into why Windows is so much slower. This build definitely adds some things but this is more than I expect.
I have ideas for finding build performance but this is a little more than I expect to find easily.

I don't think I can even see the "re-run incomplete"...

amtriathlon:
> Hmmmm... let me dig into why Windows is so much slower. This build definitely adds some things but this is more than I expect. I have ideas for finding build performance but this is a little more than I expect to find easily.

Your last change enabling PCH is likely the solution, performance regressed to how it was before that change, and Windows builds start with a gcconfig.pri bundled with precompiled libraries instead gcconfig.pri.in like Linux/macOS.

BTW, originally we have 2 ci environments: AppVeyor for Windows created by Joern and Travis for macOS by Gareth.
When Gareth leave I added Linux builds to Travis following the modular approach already in place with common parts on https://github.com/GoldenCheetah/GoldenCheetah/blob/master/deprecated/.travis.yml and separate scripts for each OS starting from a simple compile to generate release ready AppImages, then changed macOS builds to generate release ready binaries too since to do that locally was a PITA.
Windows config was monolithic on AppVeyor.yml and I extended from basic compilation to produce full release binaries too.
That worked until this year Travis deprecated macOS support, then I migrated Travis scrips to AppVeyor, initially for macOS then Linux, always trying to preserve Windows builds working since we use them for releases… and that is how end up with what you found, I prefer to make changes incrementally and ensure the binaries keep working, is not that easy to test all possibilities and we have not much help from users either except may be for Linux, to say Windows and specially macOS users are not that collaborative is quite un understatement…
In particular I would prefer this PR keep focused on Python, changes to QWT, enabling MP and anything else should be separate affairs unless strictly necessary.

> I don't think I can even see the "re-run incomplete"...

I think Cancel/Rebuild/Re-run incomplete depend on permissions.

gille:
Exactly, this was because when I had to change how PCH worked internally it broke on Windows (having the log line turned useful). Figuring this out though took more time than I would have liked (the pain of hours+ to wait on a test...).

I'm going to roll back all unrelated changes to make this be "just Python".

QWT being flaky is.. annoying but we can make a decision about that elsewhere.

On a separate note, I did make some work on the "unify appveyor scripts" to have them be one style rather than the current mixed with some in appveyor.yml, some in scripts, etc. But that's not happening for a while.

"ensure the binaries keep working, is not that easy to test all possibilities" -- A thing I want to add in the year is to as much as possible add automated tests in the code so that we have high confidence in releases without having to manually test everything. Having to manually test things is error prone, tedious, and we end up with people just testing what they care about and others whining which is annoying.

For some reason I can't even build my own branches, others have raised this with AppVeyor but it's a blackhole unfortunately.

The frustrations with Mac builds made me look into buying a Mac just to test... (I might still if I can justify it)

gille:
I finally have a Mac build that actually works! I'm going to clean this up (and fix the define to just be one keeping GC_WANTS).

Any other objections to or opinions about this change?

I integrated SIP directly because it let me patch the code (I did this online for a while, but it broke on Windows and while I eventually will succeed in the fight with the build system I'm not sure it adds much). Moving to contrib is possible but makes the build a bit more fragile as we're fighting the order of include visibility.

amtriathlon:
Hi Magnus, this looks ready to merge, great job!

We need to check for https://github.com/GoldenCheetah/GoldenCheetah/issues/4541#issuecomment-2358789257 due to the change in Pandas version.
The problem in Python Charts is described in https://groups.google.com/g/golden-cheetah-developers/c/lpHNWKdNICY including a workaround if it is still present.

gille:
The groups link is broken. Does golden-cheetah-developers exist? (it used to. maybe you need to add me? I can't even see the group as existing)

The Pandas problem is interesting, it seems like they aren't too spooked by it, but I also think I might have a fix for it.

amtriathlon:
Yes, may be you need to suscribe, I am not a forum admin in that one.

This is the problem:

```
Hey there 

I'm having issues with pandas after updating to pandas 1.3 

I'm using the following code: 

> import pandas as pd 
> 
> rd = GC.activity() 
> ix = range(0, len(rd['seconds'])) 
> df = pd.DataFrame(rd, columns=rd.keys(), index=ix) 
> 
> print(df) 


With pandas 1.2, I get: 

> Python Console (3.9.5 (default, Jul 4 2021, 10:35:01) [GCC 10.3.0]) 
>>>> seconds cadence cadenced ... POWER_Kickr POWER_Target TRAIN_TARGET 
> 0 0.0 70.0 0.0 ... 100.0 92.0 NaN 
> 1 1.0 70.0 0.0 ... 100.0 92.0 92.0 
> 2 2.0 70.0 0.0 ... 95.0 92.0 92.0 
> 3 3.0 69.0 0.0 ... 95.0 92.0 92.0 
> 4 4.0 71.0 2.0 ... 99.0 92.0 92.0 
> ... ... ... ... ... ... ... ... 
> 5398 5398.0 76.0 1.0 ... 91.0 92.0 92.0 
> 5399 5399.0 75.0 0.0 ... 91.0 92.0 92.0 
> 5400 5400.0 77.0 2.0 ... 0.0 92.0 92.0 
> 5401 5401.0 77.0 0.0 ... 0.0 92.0 92.0 
> 5402 5402.0 78.0 1.0 ... 0.0 92.0 92.0 
> 
> [5403 rows x 26 columns] 

But with pandas 1.3: 

> Python Console (3.9.5 (default, Jul 4 2021, 10:35:01) [GCC 10.3.0]) 
>>>> seconds ... TRAIN_TARGET 
> 0 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 1 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 2 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 3 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 4 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> ... ... ... ... 
> 5398 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 5399 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 5400 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 5401 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 5402 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, ... ... [nan, 92.0, 92.0, 92.0, 92.0, 92.0, 92.0, 93.0... 
> 
> [5403 rows x 26 columns] 
```
and the workaround is to use list() over the array-like objects before to create the DataFrame:

> See https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Special-Topics_Working-with-Python#gcactivityactivitynone, it’s a dictionary with series names as keys and an array like object as values.
> The array like objects are iterable and they can be easily converted to standard types using list(), np.asarray(), etc.

Eventually we could do that in a Python wrapper in https://github.com/GoldenCheetah/GoldenCheetah/blob/1a4dd1d10f346bfea25b7086522146703288325a/src/Resources/python/library.py if we cannot find a better solution.

gille:
The workaround should have hinted at what the problem actually is. It's a bug in the Python wrapping in Golden Cheetah.
I'll send a PR for it (I have a fix).

I've also submitted a fix to Pandas but that's going to take a while to get merged & released.

gille:
This is the upstream Pandas PR if anyone is interested: https://github.com/pandas-dev/pandas/pull/63781

amtriathlon:
Hi Magnus, I installed the Windows version, embedded Python and SIP integration seems to be Ok, but none of the required dependencies for Python charts (numpy, pandas, plotly, etc.) is available:
image

lit looks like site-packages folder was not included in the installer.

amtriathlon:
I downloaded the AppImage and it reports:

Python version mismatch: GoldenCheetah was built with Python 3.13, but found Python 3.10 at /usr/bin/python3
Python embedding failed. GoldenCheetah requires Python 3.13 installed and in PATH.

image

while according to the configuration:

image

it should use the Python version embedded in the AppImage.

BTW, I understood the idea was to standarize on Python 3.11 not 3.13.

MeisterP:
I had a quick look at the AppImage on linux

It is built with `python3.13` but bundles `python3.11`

About GoldenCheetah: `Python: 3.13.5 [3.13.5]`

```
$ ./GoldenCheetah_v3.8_x64.AppImage --appimage-extract
$ ls squashfs-root/opt/
_internal  python3.11
```

looks like python is installed twice: https://github.com/gille/GoldenCheetah/blob/899c8fe7dfa0cdc6ff1020e23636063a59dd63cc/appveyor/linux/install.sh#L41 and https://github.com/gille/GoldenCheetah/blob/899c8fe7dfa0cdc6ff1020e23636063a59dd63cc/appveyor/linux/install.sh#L50


According to the appveyor log:

first install:
```
[00:03:17] python3 --version
[00:03:18] Python 3.13.5
```

and later
```
[00:03:58] python3.11 --version
[00:03:59] Python 3.11.14
```

it then picks up the wrong version when building:
```
[00:05:05] Project MESSAGE: Python configuration:
[00:05:05] Project MESSAGE:   Includes: -I/home/appveyor/.pyenv/versions/3.13.5/include/python3.13 -I/home/appveyor/.pyenv/versions/3.13.5/include/python3.13
[00:05:05] Project MESSAGE:   Libs: -L/home/appveyor/.pyenv/versions/3.13.5/lib -lpython3.13 -ldl -L/home/appveyor/.pyenv/versions/3.13.5/lib -Wl,-rpath,/home/appveyor/.pyenv/versions/3.13.5/lib -lm
[00:05:05] Project MESSAGE: Enabling Python support
```

and installs the 3.11 python appimage
```
[00:47:13] 2026-01-18 04:40:43 URL:https://release-assets.githubusercontent.com/github-production-release-asset/251001753/df3ad70a-449e-498c-9c5a-8c732adb7048?sp=r&sv=2018-11-09&sr=b&spr=https&se=2026-01-18T05%3A27%3A55Z&rscd=attachment%3B+filename%3Dpython3.11.14-cp311-cp311-manylinux_2_28_x86_64.AppImage&rsct=application%2Foctet-stream&skoid=96c2d410-5711-43a1-aedd-ab1947aa7ab0&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skt=2026-01-18T04%3A26%3A59Z&ske=2026-01-18T05%3A27%3A55Z&sks=b&skv=2018-11-09&sig=syIP5avIuh5B3gxhXj6UPxxvroIEz4BmeGc3l0Il1Zg%3D&jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmVsZWFzZS1hc3NldHMuZ2l0aHVidXNlcmNvbnRlbnQuY29tIiwia2V5Ijoia2V5MSIsImV4cCI6MTc2ODcxMzA0MywibmJmIjoxNzY4NzExMjQzLCJwYXRoIjoicmVsZWFzZWFzc2V0cHJvZHVjdGlvbi5ibG9iLmNvcmUud2luZG93cy5uZXQifQ.toJmkjA9_17wrhNPreyOZTqlkXGhFVQ6se4zrwX9qYg&response-content-disposition=attachment%3B%20filename%3Dpython3.11.14-cp311-cp311-manylinux_2_28_x86_64.AppImage&response-content-type=application%2Foctet-stream [17570296/17570296] -> "python3.11.14-cp311-cp311-manylinux_2_28_x86_64.AppImage" [1]
[00:47:13] chmod +x "${PYTHON_APPIMAGE_FILE}"
[00:47:13] ./"${PYTHON_APPIMAGE_FILE}" --appimage-extract
[00:47:13] rm -f "${PYTHON_APPIMAGE_FILE}"
[00:47:13] export PATH="$(pwd)/squashfs-root/usr/bin:$PATH"
[00:47:13] pip install --upgrade pip
[00:47:15] Requirement already satisfied: pip in ./squashfs-root/opt/python3.11/lib/python3.11/site-packages (25.2)
```

amtriathlon:
I was surprised included Python is not used and likely unusable from the AppImage it looks like it was not even minimally tested.
Does this impact macOS too, internal Python is used from the DMG package?

gille:
That's weird... I did test it but I bet it worked for me because of the versions aligning on my system and it just grabbed the system Python by accident.

MacOS is fine i triple checked it and Mac is really picky about this. 

Quite a bit of this pain is coming from the NOT adopting what I did for Mac, let me do that everywhere (oh how I wish we had brew but we don't).

amtriathlon:
Hi Magnus, missing required packages on windows installer was due to missing target folder when installing them, they ended up in system Python instead of on the embeddable one. 
It is working with my usual charts now, I will continue testing. Thank you! 
PS: assuming macOS is working I will try to understand the problem on Linux to see if we can merge this.

amtriathlon:
The AppImage from last build detects and uses embedded Python 3.11 correctly, required modules (numpy, pandas, etc.) cannot be imported so I will look at this next.
PS: also need to tidy up the changes, likely using a PYTHON_VERSION environment variable set in appveyor.xml to be used Linux scripts.

MeisterP:
It's working when called directly:
```
./squashfs-root/opt/python3.11/bin/python3.11 -c "import sys; print(sys.path); import pandas as pd; print(pd.__version__)"                                                             1
['', '/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python311.zip', '/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python3.11', '/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python3.11/lib-dynload', '/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python3.11/site-packages']
3.0.0
```
(my system pandas is 2.3.3, so I'm sure it's picking up the correct one)

`print(sys.path)` from within GC is `['/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python311.zip', '/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python3.11', '/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python3.11/lib-dynload', '.']`

it's missing the `opt/python3.11/lib/python3.11/site-packages` directory for some reason.
adding `sys.path.append("/home/poncho/Downloads/Firefox/squashfs-root/opt/python3.11/lib/python3.11/site-packages")` gets things working.

amtriathlon:
I tried .pth and ._pth files and none seems to work, if we can not find a better solution we could to add site-packages path to [Library.py](https://github.com/GoldenCheetah/GoldenCheetah/blob/master/src/Resources/python/library.py) as:
`if sys.platform=='linux': sys.path.append(sys.prefix+'/lib/python3.'+str(sys.version_info.minor)+'/site-packages')`

MeisterP:
I have no idea why it isn't working from within GC.

There is also some linux specific code and path manipulation in `PythonEmbed.cpp`. Might be another place to add a workaround instead of `Library.py`

https://github.com/GoldenCheetah/GoldenCheetah/blob/d6c6830b92be3ab6d1e1ed467341231ae3e71816/src/Python/PythonEmbed.cpp#L266 and https://github.com/GoldenCheetah/GoldenCheetah/blob/d6c6830b92be3ab6d1e1ed467341231ae3e71816/src/Python/PythonEmbed.cpp#L278

amtriathlon:
Yes, it may be better in PythonEmbed since we can apply it only to the deployed Python case.

MeisterP:
> Yes, it may be better in PythonEmbed since we can apply it only to the deployed Python case.


Compiled locally on linux with system python and the path is correct.
So my guess is that it's either an issue with the appveyor scripts, or with the ubuntu image used to build. Either way, difficult to debug.

```
import sys
print(sys.prefix)
print(sys.path)

Python Console (3.13.11 (main, Jan 21 2026, 17:52:36) [GCC 15.2.1 20251122])
>>> /usr
['/usr/lib/python313.zip', '/usr/lib/python3.13', '/usr/lib/python3.13/lib-dynload', '/usr/lib/python3.13/site-packages', '.']
```

amtriathlon:
The AppImage is working with my Python charts now, if macOS dmg is Ok too, we could tidy up and merge.

MeisterP:
> The AppImage is working with my Python charts now, if macOS dmg is Ok too, we could tidy up and merge.

I can't test macOS.

The "compiled locally on linux" use case is still working with site-packages in path twice.

`['/usr/lib/python313.zip', '/usr/lib/python3.13', '/usr/lib/python3.13/lib-dynload', '/usr/lib/python3.13/site-packages', '.', '/usr/lib/python3.13/site-packages']`

amtriathlon:
I tried the macOS package on a M1 MacBook Air with Sequoia and it works as expected with my test Python charts, so far so good!

amtriathlon:
Thanks Magnus, great job!

gille:
No problem! It was a fun challenge to tackle.

amtriathlon:
Hi Magnus, testing by users is going well, there were some missing DLLs on Windows but after fixing that all pre-installed modules work as expected and it is possible to use pip to install new ones on Windows and Linux after extracting the AppImage, I detailed that in https://github.com/GoldenCheetah/GoldenCheetah/releases/tag/snapshot
On macOS preinstalled packages work as expected too and I can use pip on the included Python framework to list installed packages, but trying to install a new one doesn't work due to the lack of SSL support, do you have an idea if/how can we fix this?

peret2000:
Probably neither related nor will be the solution, but just in case:
I had to declare an environment variable before launching GC:
export SSL_CERT_DIR=/etc/ssl/certs
I did that a long time ago, unfortunately I do not remember the problem, but if it was not declared, some python scripts did not work.
I have found references to that variable in:
opt/python3.12/lib/python3.12/site-packages/pip/_vendor/truststore/_openssl.py
(where my gc embedded python is installed)
Maybe a clue, if the embedded python does not have a reference to the host certificates path

amtriathlon:
> I had to declare an environment variable before launching GC:
> export SSL_CERT_DIR=/etc/ssl/certs

Thank you, I added that to release page for Linux users having that issue.

On macOS here is the error I get, I am not sure if it is a missing module/library, certificates access or anything else:
```
$ pwd
/Applications/GoldenCheetah.app/Contents/Frameworks/Python.framework/Versions/Current/bin
$ ./python3.11 --version
Python 3.11.7
$ ./python3.11 -m pip list
Package            Version
------------------ -----------
asteval            1.0.8
dill               0.4.1
importlib_metadata 8.7.1
lmfit              1.3.4
narwhals           2.15.0
numpy              2.4.1
packaging          26.0
pandas             3.0.0
pip                25.3
plotly             6.5.2
python-dateutil    2.9.0.post0
scipy              1.17.0
setuptools         80.10.2
sip                6.15.1
six                1.17.0
uncertainties      3.2.3
wheel              0.42.0
zipp               3.23.0
$ ./python3.11 -m pip install requests
WARNING: Disabling truststore since ssl support is missing
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/requests/
WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/requests/
WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/requests/
WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/requests/
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/requests/
Could not fetch URL https://pypi.org/simple/requests/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/requests/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
ERROR: Could not find a version that satisfies the requirement requests (from versions: none)
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Could not fetch URL https://pypi.org/simple/pip/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/pip/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
ERROR: No matching distribution found for requests
```

gille:
Hmm interesting. Let me dig into it and see what I can come up with.

gille:
I know what's wrong, I've fixed it locally but fixing it exposing that there's a fundamental problem in the Python integration so fixing that is going to take just a bit longer (no more than a day or so assuming I have time to work on it).

The best part of the fix is that I think it'll get around most of the reported issues.

PR #4749, opened 2025-12-11
Parallelize builds and upgrade build environments

gille:


amtriathlon:
Hi Magnus, AppVeyor workers are 2 core for the free tier, to use -j4 may slow the builds due to additional context switches, AFAIK.

gille:
Yeah lets drop it down to j2, anyway this should let us build qwt with j2 which would be nice

gille:
Can we pull this one first? I'd like to rebase the other Python work on top of this.

amtriathlon:
Hi Magnus, there is still a  `# TMP will revert!` comment, I thought it was not ready yet.
Have you downloaded and tested the generated installers on Windows and on which macOS versions?

amtriathlon:
I did a minimum test on Sequoia and it seems to be ok, WebEngine is working and Python charts too. That’s a good thing since Monterey is out of maintenance more than a year ago.

image

PS: the Windows installer also works on Windows 11, I updated the description for the VC runtime in the installer script.

amtriathlon:
Thank you!

gille:
np! Sorry I got super busy and didn't get time to remove that comment or respond to you :(

Let me tackle and finish up the Python stuff ASAP.

Issue #4750, opened 2025-12-12
Deprecate QT5


gille:
QT6 is stable and widely available, there's no reason to keep supporting two frameworks that do the same.

amtriathlon:
Now the ci builds are running smoothly on all 3 supported platforms using Qt6, I agree.

PS: Travis-ci using Qt5 for Linux builds was disabled by https://github.com/GoldenCheetah/GoldenCheetah/commit/84b26c5977bc3cc4c5be102b300a2d725692cda1, so we don't need Qt5 support anymore and it can be removed to simplify the code and then to remove deprecated warnings.

paulj49457:
If the aim is to deprecate old versions of QT, are you meaning that all the QT_VERSION compilation logic is removed? For example:

#if QT_VERSION >= 0x060000
    QMultiHashIterator<QString, QwtPlotCurve*> c(curves);
#else
    QHashIterator<QString, QwtPlotCurve*> c(curves);
#endif

gille:
My goal (and opinion) is to force a minimum version of QT6.8 which is the current LTS. 

If this means that it builds with all current QT that's great but 6.8 is my main target (although I think I usually build against 6.10 at home).

amtriathlon:
> If the aim is to deprecate old versions of QT, are you meaning that all the QT_VERSION compilation logic is removed?

Yes, there is no point to keep this code, IMHO.

> My goal (and opinion) is to force a minimum version of QT6.8 which is the current LTS.

Currently this would be a problem, we are using Qt6.5 for macOS builds and Qt6.6 is the latest available on AppVeyor images: https://www.appveyor.com/docs/macos-images-software/#qt

gille:
Good point..
6.5.3 seems to be the one common version across all systems on AppVeyor so I guess that's a better target then.
https://www.appveyor.com/docs/macos-images-software/#qt
https://www.appveyor.com/docs/windows-images-software/#qt
https://www.appveyor.com/docs/linux-images-software/#qt

paulj49457:
I can work through the conditional compiler Qt logic in the PR #4751, removing anything older than 6.5.3

Maybe someone can look at the QDateTime::setTimeSpec': Use setTimeZone() warnings, and add that to #4751 ?

It would be great to have a clean build with no Qt warnings.

amtriathlon:
> It would be great to have a clean build with no Qt warnings.

[...]

PR #4751, opened 2025-12-13
Remove QT5 compatibility and deprecation warnings

paulj49457:
This PR fixes some of the issues in https://github.com/GoldenCheetah/GoldenCheetah/issues/4750, and removes deprecation warnings now GC is using Qt6.5.3 as minimum version.

- Access for points has changed.
- qAsConst should be replaced with std::as_const
- QScopedPointer should be replaced with std::unique_ptr
- addAction constructor has changed, earlier versions are deprecated
- 'QtPrivate::warnIfContainerIsNotShared': Do not use foreach/Q_FOREACH with containers which are not implicitly shared, prefer using a range-based for loop with these containers: for (const auto &it : container),

There are also other Qt warnings which might be worth looking at in the future:

warning C4996: 'QDateTime::QDateTime': Pass QTimeZone instead
warning C4996: 'QDateTime::setTimeSpec': Use setTimeZone() instead

warning C4996: 'QChart::setAxisX': was declared deprecated
warning C4996: 'QChart::axisY': was declared deprecated
warning C4996: 'QChart::axisY': was declared deprecated

### DIFF
--- a/src/Gui/Pages.cpp
+++ b/src/Gui/Pages.cpp
@@ -135,11 +135,7 @@ GeneralPage::GeneralPage(Context *context) : context(context)
     garminHWMarkedit->setValue(garminHWMark.toInt());
 
     connect(garminSmartRecord, 
-#if QT_VERSION < 0x060700
             &QCheckBox::stateChanged,
-#else
-            QOverload<Qt::CheckState>::of(&QCheckBox::checkStateChanged),
-#endif
             this, [this](int state) { garminHWMarkedit->setEnabled(state); });
 
     garminSmartRecord->setCheckState(! (isGarminSmartRecording.toInt() > 0) ? Qt::Checked : Qt::Unchecked);
@@ -201,11 +197,7 @@ GeneralPage::GeneralPage(Context *context) : context(context)
 
     connect(rBrowseButton, SIGNAL(clicked()), this, SLOT(browseRDir()));
     connect(embedR,
-#if QT_VERSION < 0x060700
             &QCheckBox::stateChanged,
-#else
-            QOverload<Qt::CheckState>::of(&QCheckBox::checkStateChanged),
-#endif
             this, [this](int state) { rDirectorySel->setEnabled(state); });
 
     embedR->setChecked(! appsettings->value(NULL, GC_EMBED_R, true).toBool());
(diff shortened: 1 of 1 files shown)


paulj49457:
@amtriathlon  I have removed all the QT compiler directives <6.5.3 as per discussion in #4750, and I have set the minimum QT version check within src.pro to QT6.5.3

I wasn't sure whether you wanted qwt folder updated, if you do, then [5647af8](https://github.com/GoldenCheetah/GoldenCheetah/pull/4751/commits/5647af8a2a0eee7a6e6377ac41fa8bc0ab2c129f)  can be reinstated.

I see that in the utils/rpi folder the raspberry pi build scripts reference QT5, do these need to be updated?

I think the removal of core5compat library and addressing the QRegExp & QXmlDefaultHandler issues should be separate PRs.

paulj49457:
Builds:

Windows, Qt6.8.3, using c++17, the recommend c++ version for Qt6
Linux: Qt6.8.3, using c++11 ( has -Wno-c++11-narrowing compiler flag)
Mac: Qt6.5.3, using c++11 ( has -Wno-c++11-narrowing compiler flag)

The remaining warnings should be separate PRs:

- Update Linux & Mac to use c++17
- The removal of core5compat library and addressing the QRegExp & QXmlDefaultHandler updates .
- Update Flex so it doesn't generate the "register" keyword which has been removed from the c++17 standard
- warning C4996: 'QDateTime::QDateTime': Pass QTimeZone instead
- warning C4996: 'QChart::setAxisX': was declared deprecated

amtriathlon:
Hi Paul, yes, better leave QWT without changes as discussed at the issue, Raspberry PI scripts are intended to build v3.5 IIRC, so  don’t change either.
BTW it would be very helpful if you enabled PpVeyor in your fork to check your work before to create a PR to avoid clogging the shared instance now that several developers are working concurrently.

paulj49457:
@amtriathlon I created an Appveyor account and was using it until Qt6 prevented GC building within the 1 hour limit, but @gille has fixed that (and it compiles much quicker locally as well, its a great update ) so I can now use my local Appveyor. I'll work on the remaining items in separate local PRs until they are ready, but I think this one is ready for review.

amtriathlon:
Thank you!

paulj49457:
@amtriathlon I thought I'd followed the guidance to replace QScopedPointer using the qt warning information and an internet search, but obviously not.   I'm sorry and thank you for fixing it.

PR #4761, opened 2025-12-15
Fix crash safety issues: Unsafe signal connections and tree child access

gille:
Systematically resolved 30 identified potential crash vectors and established automated regression testing to prevent strict reoccurrence.

Key Changes:
- Fixed 11 instances of unsafe `QObject::connect` calls (missing context object) in src/Charts/AgendaWindow.cpp, src/FileIO/FixSpikes.cpp, src/Gui/Agenda.cpp, src/Gui/BatchProcessingDialog.cpp, and src/Gui/IconManager.cpp. This prevents crashes caused by signals firing after the receiver has been destroyed.
- Fixed 19 instances of unsafe `QTreeWidgetItem` child access in src/Charts/LTMChartParser.cpp, src/Gui/ColorButton.cpp, src/Gui/AthletePages.cpp, and src/Gui/Pages.cpp by adding defensive `nullptr` checks before dereferencing.
- Added Python detection scripts util/check_unsafe_connects.py and util/check_unsafe_tree_child.py to statically analyze the codebase for these specific unsafe patterns.
- Integrated detection scripts into the regression test suite under `unittests/Core/signalSafety`, verifying the fixes and enforcing a strict zero-tolerance policy for future regressions.
- Added `testSplineCrash` to cover edge cases with empty spline lookups.


amtriathlon:
Thank you!

PR #4762, opened 2025-12-16
QFileDialog::getOpenFileName returns an empty string when the user pr…

gille:
…esses cancel, there's no need to throw a message box up


amtriathlon:
That was annoying, thank you!

Issue #4760, opened 2025-12-14
Workout Info Widget don't use Zones names


amtriathlon:
Short names are fixed to Zx which is different from what the rest of the program does and some? users expect, reported at the users forum: https://groups.google.com/g/golden-cheetah-users/c/JZb-mp70xiU/m/SdYxLuJ-AQAJ

thejockl:
Will pick this up

PR #4763, opened 2025-12-16
Workout Infowidget: Showing real zone name

thejockl:
See #4760
Display the zone names instead of a fixed Z# in the Workout Infowidget. Since configChanged is emitted only when zones are added or modified - but not when their names or descriptions are edited - changes to a zone’s name or description requires selecting a different workout to reflect the change


amtriathlon:
Thank you!

PR #4766, opened 2025-12-21
Qt6 checkstatechanged warnings

paulj49457:
Remove Qt warning for builds with Qt version >= 6.7.0

warning C4996: 'QCheckBox::stateChanged': Use checkStateChanged() instead

From QT6.7.0 QCheckBox::stateChanged() is being replaced with QCheckBox::checkStateChanged(), and stateChanged() is deprecated in Qt6.9.0


amtriathlon:
Thank you!

PR #4767, opened 2025-12-21
Unused variables and parameter build warnings

paulj49457:
Remove the following types of build warnings:

warning C4189: local variable is initialized but not referenced
warning C4100: unreferenced parameter
warning C4101: unreferenced local variable

I have only commented out unused variables and used [[maybe_unused]] for unused parameters.


amtriathlon:
Thank you!

PR #4768, opened 2025-12-22
QZip improvements

gille:
Removed unused constructors
Replaced raw pointer with a unique_ptr to clearly pass ownership.
Fixed a bug where we compared if a signed integer is negative (it never is).
Fixed a case where we take the pointer to a pointer we de-reference!


amtriathlon:
Thank you!

PR #4769, opened 2025-12-22
Nodiscard

gille:


amtriathlon:
Hi Magnus, for the connect case wouldn't be better to change to functor-based syntax (https://doc.qt.io/qt-6/signalsandslots-syntaxes.html#) so the verification happens at compile time? 
Checking the result code is ugly and to call qFatal is a no go for us.

PR #4775, opened 2025-12-25
linux/51-garmin-usb.rules: use mode 0660 and tag uaccess

MeisterP:
Use a more modern approach for device access.

The uaccess tag makes udev apply a dynamic user ACL to the device node, which makes the device usable to logged-in users.

See https://wiki.archlinux.org/title/Udev#Allowing_regular_users_to_use_devices


amtriathlon:
Merged, thank you!

I think this should be linked from https://github.com/GoldenCheetah/GoldenCheetah/wiki/ANT--and-BTLE-Troubleshooting#accessing-the-garmin-usb2-stick--tacx-usb-ant-antenna-t2028-on-linux, otherwise it is buried deep in source code where regular users don't look.

MeisterP:
Thanks. I've [updated](https://github.com/GoldenCheetah/GoldenCheetah/wiki/ANT--and-BTLE-Troubleshooting/_compare/85370037b00af6ce887481ead66b48d58f528d7d...12b9458cd1149889a6e65e0ae9a83648ecebc8a1) the wiki accordingly

PR #4776, opened 2025-12-27
Upgrade FIT SDK to 21.188

MeisterP:
- use raw strings for all regexes in fitprod.py
- Upgrade FIT SDK to 21.188


amtriathlon:
Thank you!

PR #4784, opened 2026-01-02
Standardize how the config is created for releases. This caused more …

gille:
…confusion than warranted with the Python work so it's best we have one instead of two ways of doing the same thing.


amtriathlon:
Thank you!

PR #4785, opened 2026-01-02
httpsessionstore timerEvent() hides QObject timerEvent() virtual function

paulj49457:
The MacOs compiler detects that httpsessionstore timerEvent() hides QObject's  timerEvent() virtual function:

../contrib/httpserver/httpsessionstore.h:102:10: warning: 'HttpSessionStore::timerEvent' hides overloaded virtual function [-Woverloaded-virtual]

So I have renamed the httpsessionstore timerEvent() to cleanUpTimerEvent().

### DIFF
--- a/contrib/httpserver/httpsessionstore.cpp
+++ b/contrib/httpserver/httpsessionstore.cpp
@@ -11,7 +11,7 @@ HttpSessionStore::HttpSessionStore(QSettings* settings, QObject* parent)
     :QObject(parent)
 {
     this->settings=settings;
-    connect(&cleanupTimer,&QTimer::timeout,this,&HttpSessionStore::cleanupTimerEvent);
+    connect(&cleanupTimer,SIGNAL(timeout()),this,SLOT(timerEvent()));
     cleanupTimer.start(60000);
     cookieName=settings->value("cookieName","sessionid").toByteArray();
     expirationTime=settings->value("expirationTime",3600000).toInt();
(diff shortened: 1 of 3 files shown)


amtriathlon:
Hi Paul, I didn’t use server mode, have you tested this?

That way of connect may fail at runtime, if you want to change this is would be better to switch to functor-based syntax (https://doc.qt.io/qt-6/signalsandslots-syntaxes.html#) so the verification happens at compile time.

paulj49457:
Hi Ale, I've updated to the functor approach which I had not appreciated provides compile time detection.

I have tried to test the latest update, running GC in server mode, and making rest API calls, and also using the WebPage chart within GC but the httpsession & httpsessionstore classes never get constructed. So I commented out their headers and source files from src.pro to confirm they aren't used and it builds fine on all GC platforms.

So I guess GC doesn't support http sessions, and the fix might as well be to just remove the (httpsession & httpsessionstore) files from the src.pro build file?

amtriathlon:
> So I guess GC doesn't support http sessions, and the fix might as well be to just remove the (httpsession & httpsessionstore) files from the src.pro build file?

You are right, server mode has no session concept, likely those modules were added thinking in future expansions, but that was before R/Python embedding which mostly superseded the http interface. I would just remove them from src.pro

paulj49457:
@amtriathlon PR updated to remove httpsession & httpsessionstore from the build.

amtriathlon:
Thank you!

PR #4786, opened 2026-01-02
Update HTTP TLS version

paulj49457:
The MCVS compiler detects:

..\contrib\httpserver\httpconnectionhandlerpool.cpp(126): warning C4996: 'QSsl::TlsV1_0': Use TlsV1_2OrLater instead.

AI says:

The use of QSsl::TlsV1_0 is discouraged in favor of QSsl::TlsV1_2OrLater for improved security and compatibility. This recommendation stems from the fact that QSsl::TlsV1_0OrLater is not a standard protocol option in Qt, and the intended behavior of allowing TLS 1.0 or higher is better achieved using QSsl::TlsV1_2OrLater, which ensures support for TLS 1.2 and later versions, aligning with modern security standards.


amtriathlon:
Hi Paul, what kind of testing have you done and on what platforms?

paulj49457:
Needs further testing, maybe this could be an issue to discuss updating the version.

PR #4788, opened 2026-01-03
Fixes so that lupdate doesn't throw warnings (and break on Windows).

gille:
This mistake also likely contributes to internationalization problems.

src/Gui/CalendarItemDelegates.h:
  ColumnDelegatingItemDelegate
    - uses tr() so should call Q_OBJECT CalendarDetailedDayDelegate
    - uses tr() so should call Q_OBJECT
    - Moved Q_OBJECT to the top of the class for consistency and this also guarantees it's private (which it has to be).

src/Gui/ColorButton.h:
  - Moved the comment from Colors.h here.

src/Gui/Colors.h:
  - Removed #if 0 code that confused lupdate since it doesn't understand preprocessor macros.

### DIFF
--- a/src/Gui/CalendarItemDelegates.h
+++ b/src/Gui/CalendarItemDelegates.h
@@ -115,6 +115,7 @@ private:
 
 
 class CalendarHeadlineDelegate : public QStyledItemDelegate {
+    Q_OBJECT
 public:
     enum Roles {
         DayRole = Qt::UserRole + 1 // [CalendarDay] Calendar day
@@ -153,6 +154,7 @@ private:
 
 
 class CalendarCompactDayDelegate : public QStyledItemDelegate {
+    Q_OBJECT
 public:
     enum Roles {
         DayRole = Qt::UserRole + 1, // [CalendarDay] Calendar day
@@ -191,6 +193,7 @@ private:
 
 
 class AgendaMultiDelegate : public QStyledItemDelegate {
+    Q_OBJECT
 public:
     enum Roles {
         // Qt::DisplayRole            // [QString] The default text to display
@@ -229,6 +232,7 @@ private:
 
 
 class AgendaEntryDelegate : public QStyledItemDelegate {
+    Q_OBJECT
 public:
     enum Roles {
         EntryRole = Qt::UserRole, // [CalendarEntry] Entry to be displayed
--- a/src/Gui/ColorButton.h
+++ b/src/Gui/ColorButton.h
@@ -25,6 +25,9 @@
 #include <QDialog>
 #include <QColor>
 
+// color chooser that also supports the standard colors (CPLOTMARKER, CPOWER)
+// and returns them as a QColor(1,1,1,<int>) where <int> is the color number
+// .e.g CPOWER is 18, see below for full list
 class ColorButton : public QPushButton
 {
     Q_OBJECT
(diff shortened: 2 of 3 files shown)


amtriathlon:
Hi Magnus, good catch on Colors.h that warning always bothered me but I could not find the cause.
WRT enabling tr context on lightweight objects we use  `Q_DECLARE_TR_FUNCTIONS` instead of making them full QOjects like in:

https://github.com/GoldenCheetah/GoldenCheetah/blob/1b21c6a4fc03819324de6d27eb635ba4bc5688d1/src/Metrics/AerobicDecoupling.cpp#L43

gille:
So that might work in some cases, but often doesn't and here, if I revert, the warning returns.

You should only use Q_DECLARE_TR_FUNCTIONS if your class isn't a Qbject (which means it never inherits QObject).
All of these inherit QObject.

https://doc.qt.io/qt-6/i18n-source-translation.html#c-use-tr
https://doc.qt.io/qt-6/qcoreapplication.html#Q_DECLARE_TR_FUNCTIONS

Some of the existing usages or TR_FUNCTIONS is likely wrong as they inherit Q_Object and probably should be updated (but we should do that later in the hunt for better translations). I'm not sure why lupdate isn't complaining about them.

Side note: the translations are really far behind, do you want me to drop a PR that updates them (technically this should be done with every PR but that's not been happening and our CI doesn't catch it). I went down this rabbit hole while fixing the Windows build for Python and noticed they don't update there).

amtriathlon:
> So that might work in some cases, but often doesn't and here, if I revert, the warning returns.
> 
> You should only use Q_DECLARE_TR_FUNCTIONS if your class isn't a Qbject (which means it never inherits QObject). All of these inherit QObject.
> 
> https://doc.qt.io/qt-6/i18n-source-translation.html#c-use-tr https://doc.qt.io/qt-6/qcoreapplication.html#Q_DECLARE_TR_FUNCTIONS

You are right, I have not noticed the classes you were talking about inherit from QObject, so your solution is the right one for them, my fault, sorry.

> Some of the existing usages or TR_FUNCTIONS is likely wrong as they inherit Q_Object and probably should be updated (but we should do that later in the hunt for better translations). I'm not sure why lupdate isn't complaining about them.

Supposedly we are using Q_DECLARE_TR_FUNCTIONS only for classes which don't inherit form QObject s.t. RideMetric subclasses.
 
> Side note: the translations are really far behind, do you want me to drop a PR that updates them (technically this should be done with every PR but that's not been happening and our CI doesn't catch it). I went down this rabbit hole while fixing the Windows build for Python and noticed they don't update there).

Both macOS and Linux builds run lupdate in before_build script, it is missing on Windows builds so we should add it to appveyor.yml.

OTOH after major changes and before release we run lupdate and commit the changes so translators can work with the .ts files directly.

gille:
For RideMetric that's fine, but src/FileIO/PowerTapDevice.h uses it and inherits Device which inherits QObject (but I can't repro the errors there)

I'm adding lupdate to Appveyor in my branch (https://github.com/GoldenCheetah/GoldenCheetah/pull/4784) (that's what prompted this fix)

amtriathlon:
> For RideMetric that's fine, but src/FileIO/PowerTapDevice.h uses it and inherits Device which inherits QObject (but I can't repro the errors there)

PowerTap translation is working AFAIK, I still have a LYC for testing.

> I'm adding lupdate to Appveyor in my branch (#4784) (that's what prompted this fix)

👍

gille:
Any objection to taking this one as-is? I need it to move the config part forward, which I need to wrap up Python support.

amtriathlon:
Done, thank you!

PR #4790, opened 2026-01-05
Gc Qwt time zone warnings

paulj49457:
The following files/classes in Qwt:

qwt_date.h
qwt_date_scale_draw.h
qwt_date_scale_engine.h

are not used by GC, but are built and generate unnecessary QTimeZone deprecation warnings.


amtriathlon:
Remember https://github.com/GoldenCheetah/GoldenCheetah/issues/4750#issuecomment-3655722037, also Qwt is cached on AppVeyor builds so this is pointless.

PR #4791, opened 2026-01-05
Qwt Type Truncation Warnings

paulj49457:
Removes the type truncation warnings, I'll see if I can push this to the Qwt project at some point (https://sourceforge.net/projects/qwt/)


amtriathlon:
Remember https://github.com/GoldenCheetah/GoldenCheetah/issues/4750#issuecomment-3655722037, also Qwt is cached on AppVeyor builds so this is pointless.

PR #4792, opened 2026-01-06
Athlete folder change confirmation

paulj49457:
Looking at the recent discussions on changing the Athlete's home directory, I think I might have found an issue with the confirmation dialog. After the user clicks save on the general page changes, if the athlete folder has changed, the config dialog asks for confirmation, but when the use clicks "Yes, Apply and Restart" the msgBox.exec() function returns 3 rather than the expected 1, so it always reverts the change.

image

I have updated the code to use the QAbstractButton and clickButton() approach which is widely used within GC. Testing confimrs this correctly commits the change and restarts GC with the new athlete folder. 

Possibly the return value has changed between Qt6 & 5, it must have worked in the past, but why not now?

### DIFF
--- a/src/Gui/ConfigDialog.cpp
+++ b/src/Gui/ConfigDialog.cpp
@@ -268,10 +268,11 @@ void ConfigDialog::saveClicked()
 
         // we want our own buttons...
         msgBox.addButton(tr("No, Keep current"), QMessageBox::RejectRole);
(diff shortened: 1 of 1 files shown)


amtriathlon:
Good catch, may be since that code has not changed in 13 years, your change seems more clear and robust anyway, thank you.

PR #4794, opened 2026-01-07
Key the CloudDB cache by hostname and add a button to clear the cache

gille:
This would have saved me a lot confusion today as the old system pulls from any host meaning you can easily get one chart from one host another chart from another host and then a state that doesn't actually exist anywhere!


gille:
MacOS failed with no warning, Windows tried to package 3.8 after building 3.7? (this should really be a top level setting, but I don't see how I changed any of that).

amtriathlon:
Windows was my fault, I forgot to update the installer script, it is fixed in https://github.com/GoldenCheetah/GoldenCheetah/commit/8d4996c905557e0be36fb5c704cad57aa2575dd1, merge master and it will be Ok. [macOS seems an AppVeyor problem](https://help.appveyor.com/discussions/problems/38732-macos-tests-dont-seems-to-start), hope they fix it soon so I can complete v3.8-Dev2601 release.

PR #4795, opened 2026-01-08
Combine view indexes and view relevance definitions into single enumerated type

paulj49457:
Note: before trying this PR please backup your configglobal-general.ini file (as the startup view's values will change), otherwise when reverting you will just have to re-select your default startup view.

Athlete views are determined by the defines in AthleteTab.cpp, based on the order they are added to the QStackedWidget:
```
89:    views->addWidget(homeView);     // trends, view index 0
90:    views->addWidget(analysisView); // analysis, view index 1
91:    views->addWidget(planView);     // plan, view index 2
92:    views->addWidget(trainView);    // train, view index 3
```
Additionally a window's view relevance is determined by the defines in GcWindowRegistry.h:
```
#define VIEW_TRAIN       0x01
#define VIEW_ANALYSIS 0x02
#define VIEW_PLAN        0x04
#define VIEW_TRENDS   0x08
```
The view's values 0..3 are used with AthleteTab.cpp, MainWindow.cpp, and a few places elsewhere in the GC code. This leads to the following examples, and a comment to fix this issue:

Screenshot a

Screenshot b

Therefore, this PR aims to use a single GcViewType to represent both the view type and a window's view relevance, ensuring the view and window relevance are always aligned:

```
// the GcViewType represents both the view type in GC code and due its unique
// bit values it can be used as for defining a window's view relevance mask.
enum class GcViewType : unsigned int {
    NO_VIEW_SET =    0x00,
    VIEW_TRAIN =     0x01,
    VIEW_ANALYSIS =  0x02,
    VIEW_PLAN =      0x04,
    VIEW_TRENDS =    0x08
};
```
Additionally hiding the view stack indexes within AthleteTab, and providing mapping and accessor functions to the GcViewType simplifies the code and readability within MainWindow.cpp, removing many of the view based switch statements:

Screenshot 2026-01-07 230958

I appreciate refactoring PRs are not usually approved, but I think this one provides future benefit and simplifies checking the code. It also completes the set of changes started by the GcViewStackIdx & GcSideBarBtnId updates:

Screenshot d

### DIFF
--- a/src/R/RTool.cpp
+++ b/src/R/RTool.cpp
@@ -2567,7 +2567,7 @@ RTool::dfForDateRangeMeanmax(bool all, DateRange range, SEXP filter)
     UNPROTECT(1);
 
     // apply perspective filter if trends view and filtered
-    if (rtool->perspective && rtool->perspective->type() == GcViewType::VIEW_TRENDS && rtool->perspective->isFiltered()) {
+    if (rtool->perspective && rtool->perspective->viewType() == GcViewType::VIEW_TRENDS && rtool->perspective->isFiltered()) {
         filt = true;
         filelist << rtool->perspective->filterlist(DateRange(range));
     }


amtriathlon:
According to feedback at the users forum it looks like this PR is provoking crashes at startup on macOS when the initial view is Train, apparently related to live maps, see https://groups.google.com/g/golden-cheetah-users/c/iHO4SV3fn3A

paulj49457:
Thank you for the additional information, I'm looking at now...

liversedge:
I noticed that when running the initial view is defaulted to train- this should be defaulted to analysis imho

paulj49457:
Hi Mark,

The startupViewType (if it exists in config) is used otherwise it is set to NO_VIEW_SET, the code then will pick up the default in the case statement which sets the Analysis view for next time and starts GC in Analysis, see MainWindow.

Originally the start-up view values were (0..3) but with the use of the new viewType (1,2,4,8) I needed to "convert" any old configuration to a default new value, Analysis. Using the no view default in the appsettings call means I didn't have to duplicate the default code:

    // switch to the startup view, default is analysis.
    GcViewType startupViewType = static_cast<GcViewType>(appsettings->value(NULL, GC_STARTUP_VIEW, QString::number(static_cast<int>(GcViewType::NO_VIEW_SET))).toInt());
    switch (startupViewType) {
        case GcViewType::VIEW_TRENDS: selectTrends(); break;
        case GcViewType::VIEW_ANALYSIS: selectAnalysis(); break;
        case GcViewType::VIEW_PLAN: selectPlan(); break;
        case GcViewType::VIEW_TRAIN: selectTrain(); break;
        default: {
            qDebug() << "Startup view not specified or unknown value, defaulting to analysis view";
            appsettings->setValue(GC_STARTUP_VIEW, static_cast<std::underlying_type_t<GcViewType>>(GcViewType::VIEW_ANALYSIS));
            selectAnalysis();
        } break;
    }

amtriathlon:
I think that’s ok for new and older version users, but for users coming from a version which already has a starting view configured value using the old codification, Activities likely becomes Teain with the new one.

paulj49457:
I appreciate that its not perfect, but it just requires the reselection of the default start-up, and it saves having another mapping just for the configuration. It should only affect those building from source or those using one of the development snapshots.

amtriathlon:
Having to deal with users that is a PITA for me, most user don’t even know the initial view is configurable now, they just see default behavior has changed for no good reason and -off course- they complain, see the users group thread, and this affects v3.7-sp1 users which should be all of them since it is the latest release.
Anyway it was my fault, I should never have merged this PR.

paulj49457:
I'm sorry, I had not appreciated the user impact, given the start-up view functionality is in v3.7-sp1.

Fixed now, please see https://github.com/GoldenCheetah/GoldenCheetah/pull/4830

PR #4796, opened 2026-01-08
RideDB.y warning: shift reduce conflict

paulj49457:
The following warning occurs in the build, which I have investigated as I wanted to understand the Bison/Flex processing in GC, I'm not a Bison/Flex expert so there maybe a better way to fix this issue:

```
../Core/RideDB.y:45.1-12: warning: deprecated directive: '%pure-parser', use '%define api.pure' [-Wdeprecated]
   45 | %pure-parser
      | ^~~~~~~~~~~~
      | %define api.pure
../Core/RideDB.y: warning: 1 shift/reduce conflict [-Wconflicts-sr]
../Core/RideDB.y: note: rerun with option '-Wcounterexamples' to generate conflict counterexamples
```

The %pure-parser to %define api.pure seems straightforward and the change removes the warning with no apparent parsing issues.

So to address the 1 shift/reduce conflict  in RideDB.y, I ran the counter examples and got the following output:

```
../Core/RideDB.y: warning: 1 shift/reduce conflict [-Wconflicts-sr]
../Core/RideDB.y: warning: shift/reduce conflict on token ':' [-Wcounterexamples]
  First example: '{' RIDES ':' '[' '{' string . ':' string '}' ']' '}' $end
  Shift derivation
    $accept
    `-> document                                                                          $end
        `-> '{' elements                                                              '}'
                `-> element
                    `-> RIDES ':' '[' ride                                        ']'
                                      `-> '{' rideelement_list                '}'
                                              `-> rideelement
                                                  `-> ride_tuple
                                                      `-> string . ':' string
													  
	
  Second example: '{' RIDES ':' '[' '{' string . ':' '[' xdata_values ']' '}' ']' '}' $end
  Reduce derivation
    $accept
    `-> document                                                                                                $end
        `-> '{' elements                                                                                    '}'
                `-> element
                    `-> RIDES ':' '[' ride_list                                                         ']'
                                      `-> ride
                                          `-> '{' rideelement_list                                  '}'
                                                  `-> rideelement
                                                      `-> xdata
                                                          `-> xdata_name   ':' '[' xdata_values ']'
                                                              `-> string .
```

So comparing:

```
 First example: '{' RIDES ':' '[' '{' string . ':' string '}' ']' '}' $end
Second example: '{' RIDES ':' '[' '{' string . ':' '[' xdata_values ']' '}' ']' '}' $end
```

So the issue is that the parser cannot detect the difference between the ride_tuple and the xdata fields, so to fix this I have added an INFO tag to encapsulate the ride_tuple fields in the rideDB.json file, as show below.

This fixes the issue, although this will mean the first time GC is run with this PR (and when the PR is reverted) the RideDB.json file is rebuilt, but this only a cache file, so its just a one off computation time.

```
  "VERSION":"2.0",
  "RIDES":[
	{
		"INFO":{
			"date":"2020/01/01 11:58:30 UTC",
			"filename":"2020_01_01_11_58_30.json",
			"fingerprint":"198636",
...
....
			"samples":"1"
		},
		"METRICS":{
			"activity_date":"43829.00000",
			"ride_count":"1.00000",
			"workout_time":"3510.00000",
...
...
```

### DIFF
--- a/src/Core/RideDB.y
+++ b/src/Core/RideDB.y
@@ -819,17 +819,17 @@ void RideCache::save(bool opendata, QString filename)
 
                                 if (interval->stdmeans().value(index, 0.0f) || interval->stdvariances().value(index, 0.0f)) {
 
-                                    stream << "\t\t\t\t\"" << name << "\": [ \"" << QString("%1").arg(interval->metrics()[index], 0, 'f', 5) <<"\",\""
+                                    stream << "\t\t\t\t\t\"" << name << "\": [ \"" << QString("%1").arg(interval->metrics()[index], 0, 'f', 5) <<"\",\""
                                                                                << QString("%1").arg(interval->counts()[index], 0, 'f', 5) << "\",\""
                                                                                << QString("%1").arg(interval->stdmeans().value(index, 0.0f), 0, 'f', 5) << "\",\""
                                                                                << QString("%1").arg(interval->stdvariances().value(index, 0.0f), 0, 'f', 5) <<"\"]";
 
                                 // if count is 0 don't write it
                                 } else if (interval->counts()[index] == 0) {
-                                    stream << ConstructNameNumberString(QString("\t\t\t\""), name,
+                                    stream << ConstructNameNumberString(QString("\t\t\t\t\""), name,
                                         QString("\":\""), interval->metrics()[index], QString("\""));
                                 } else {
-                                    stream << ConstructNameNumberNumberString(QString("\t\t\t\""), name,
+                                    stream << ConstructNameNumberNumberString(QString("\t\t\t\t\""), name,
                                         QString("\":[\""), interval->metrics()[index], QString("\",\""), interval->counts()[index], QString("\"]"));
                                 }
                             }


gille:
Is it possible to add unit testing around this? Getting rid of these warnings is great but if we have the ability to encode what we actually expect it'll make future maintenance much easier.

paulj49457:
I'm not sure the parsing needs to be that robust (error free yes) but given that the file format is defined and written by RibeDB.y in void RideCache::save(bool opendata, QString filename) , and the RideDB.json is not expected to be altered by the user, I'm not sure unit tests would provide much benefit in this case.

liversedge:
Any changes to parsing need to be very very carefully handled as errors will be catastrophic. 
I'd rather keep the shift/reduce conflict warning than mess about with the grammar without being 100% sure it was robust.

paulj49457:
When I said "robust", I mean that GC is not getting this file from many sources whose formats might test the robustness of the parsing grammar, this file is generated and read by the RideDB.y, so it parses correctly in the format that is written, without the current warning. My only issue of living with warnings is that sometimes they rely upon the behaviour of a third party library, which might change.

Issue #4647, opened 2025-05-24
Edit menu is missing on macOS builds when using translations


amtriathlon:
This was reported at the users forum: https://groups.google.com/g/golden-cheetah-users/c/yhJ1Cp5V960/m/izSqsktfAgAJ
It seems to happen with all languages other than English and macOS versions even when translations are correct as verified before menu creation.
I is likely related to menu injection/manipulation by macOS desktop manager but we don't have a precise diagnostic or a solution yet, contributions are welcome.

amtriathlon:
Fixed in v3.8 builds.

PR #4797, opened 2026-01-09
Fix the Edit menu disappearing when changing languages.

gille:
The edit menu attempted to populate itself with aboutToShow() but this is fragile in this case.

Fixes: #4647

### DIFF
--- a/src/Gui/ConfigDialog.cpp
+++ b/src/Gui/ConfigDialog.cpp
@@ -236,7 +236,7 @@ void ConfigDialog::closeClicked()
 //   ! new mode: change the CP associated with the present mode
 void ConfigDialog::saveClicked()
 {
-    // if a refresh is happenning stop it, whilst we 
+    // if a refresh is happening stop it, whilst we
     // update all the configuration settings!
     context->athlete->rideCache->cancel();
 
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -626,7 +626,9 @@ MainWindow::MainWindow(const QDir &home)
 
 
     editMenu = menuBar()->addMenu(tr("&Edit"));
-    connect(editMenu, SIGNAL(aboutToShow()), this, SLOT(onEditMenuAboutToShow()));
+    // Force the signal to emit, without this on MacOS the edit menu disappears if we have translation enabled.
+    onEditMenuAboutToShow();
+    connect(editMenu, &QMenu::aboutToShow, this, &MainWindow::onEditMenuAboutToShow);
 
     HelpWhatsThis *editMenuHelp = new HelpWhatsThis(editMenu);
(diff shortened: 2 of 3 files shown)


amtriathlon:
Good catch, the fact it only happens on macOS when using translations disoriented me.
This was introduce by https://github.com/GoldenCheetah/GoldenCheetah/commit/000a76983dfe5dc038c225f8ad1408adacfb15d8#diff-a858c8d2d1dce20a3b475fd7013ed3348e86e11528b167843c72d304f4de12c0 likely to dynamically update the Edit menu according to configuration, after this change it looks like the user has to restart GC to get the updated Edit menu, may be populateEditMenu still needs to clear the existing menu and be called from configChanged?

gille:
Ok, I think I've fixed it (it worked just fine while testing).

For translations to kick in you need to restart GC but that's a problem outside of this patch.

thejockl:
@gille: Just out of curiosity, do you know why aboutToShow didn't work here?

gille:
The problem is that aboutToShow() doesn't fire the way you hope it does (the way it's supposed to!). MacOS (Cocoa?) is very opinionated about the Edit menu and if it is empty MacOS might just decide to not show it at all. For a while it seemed related to the ->clear() but then I noticed the signal didn't even trigger the way we hoped it did. Forcing a population of the contents avoids this and returns your original idea which lets us use signals the way they're intended.

I was even  able to reproduce this (with some struggle) with English language (no translation) but only on a native ARM64 binary.

amtriathlon:
Retrospectively Joachim's change likely worked since it involved renaming the top level menu from Edit to Process, it was me who introduced this issue renaming it back in https://github.com/GoldenCheetah/GoldenCheetah/commit/5ef6080ef5378dd3ee197a3518a07a5f88afa705, considering the way we use that menu is unrelated to typical Edit system menus perhaps to rename it was a good idea after all, to change the wiki is not that a big issue and abandoned translations seems to be a lost cause.

gille:
@amtriathlon I can reproduce the bug regardless of menu name and translation (it's just less deterministic with English). The core problem is the empty menu.

Currently all this does is a force menu build which is near free and will be safe regardless of what we call the menu.

amtriathlon:
Ok, so the real fix is to add this before connecting the `aboutToShow` signal:
```
    // Force the signal to emit, without this on MacOS the edit menu disappears if we have translation enabled.
    onEditMenuAboutToShow();
```
and all the remaining changes are non-functional refactoring, am I right?

gille:
Yes that's the actual bug fix.

This also fixes a small memory leak, without this the QActions pile up until you exit GC.
`auto* action = new QAction(QString("%1...").arg(iter->name()), toolMapper.get());
`
The rest is just modernizing the code (moving the codebase to actual C++17/20 would make it generally much more readable).

amtriathlon:
Thank you.

PR #4800, opened 2026-01-10
Remove goldencheetah_plugin_import.cpp as it is autogenerated [skip ci]

gille:


gille:
Can we pull this? This file keeps cluttering up my workspace.

amtriathlon:
Done. You could do this by yourself too, see https://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file

PR #4805, opened 2026-01-17
Dialog to build filter queries for similar activities

thejockl:
* list all (non-zero) fields and metrics
* filter list by field / metric name
* select an operator per field (ignore, equals, contains, larger than, ...)
* available in the context menus of activity- and calendar view


amtriathlon:
Thank you!

PR #4806, opened 2026-01-17
Train mode: Improved readability for ErgFilePlot

thejockl:
* Any curve in ErgFilePlot can be hidden (W'bal, Power, Speed, ...)
* Configurable plot width for curves
* When using colored power zones: Zone color can be dimmed to enhance visibility of the curves


amtriathlon:
Thank you!

PR #4808, opened 2026-01-18
Updated Time Range context menu

thejockl:
* Centralized the menu creation
* Grouping of actions
* Showing that a phase is always directly assigned to a season
* Harmonized case
* Added ellipsis to all actions opening new dialogs
* Keeping ellipsis out of translatable string to keep as many current translations as possible


amtriathlon:
Thank you!

PR #4810, opened 2026-01-20
Add iterator support to GoldenCheetah containers so they're not affec…

gille:
…ted by the Pandas regression


amtriathlon:
Hi Magnus, perhaps would be better to integrate this as a separate commit on your Python PR so we can avoid conflicts, test only once and I can rebase and merge keeping commis separated.

gille:
Sure. The only "code" that matters is the .sip file, everything else is autogenerated anyway.

Issue #4809, opened 2026-01-20
Planned Activities in Activities View sidebar


amtriathlon:
Currently Planned activities are displayed in Activities view sidebar and this may be undesirable for some users, while it is possible to filter tem using !Planned in the toolbar filter box this has shortcomings:

- It is global for all views which may be undesirable
- It is not persistent across restarts  

The alternative could be to add a context menu "Show Planned" or "Hide Planned" persistent across restarts, although other more creative UX may be better if one is proposed.

paulj49457:
An alternative might be to put a button in the banner next to "Activities" (if this is possible) with a ("Show Planned" or ""Hide Planned" label, this would make it more obvious what local display filtering is in place.  I'm happy to investigate this option or your proposal and implement either.

amtriathlon:
Hi Paul, give it a try if you like.

thejockl:
This sounds like a really useful feature. What Do you guys think about two separate trees? One for completed, one for planned, separated by tabs?

amtriathlon:
For me the advantage to keep them in a single list, besides simplicity, is to make easier to select a pair of planned/actual to compare since users typically keep the list ordered by date

paulj49457:
See PR https://github.com/GoldenCheetah/GoldenCheetah/pull/4811 for a possible solution

PR #4811, opened 2026-01-21
Selectable show/hide planned activities

paulj49457:
This PR provides a ride navigator filter to display:

- **all** activities
- only those **completed**, or
- just those **planned**

If a global filter has been set and the ride nav display filter, the ride navigator activities list will display the intersection of the two filters. The default setting is "All", and its last setting will be stored on a per athlete basis.
 
It aims to resolve issue: https://github.com/GoldenCheetah/GoldenCheetah/issues/4809

ride nav filter

### DIFF
--- a/src/Gui/AnalysisSidebar.cpp
+++ b/src/Gui/AnalysisSidebar.cpp
@@ -154,7 +154,7 @@ AnalysisSidebar::AnalysisSidebar(Context *context) : QWidget(context->mainWindow
     // create the ride navs display filter
     activityFilter = new QComboBox(this);
     activityFilter->addItem(tr("All"), static_cast<int>(RideNavFilter::ALL));
-    activityFilter->addItem(tr("Completed"), static_cast<int>(RideNavFilter::COMPLETED));
+    activityFilter->addItem(tr("Actual"), static_cast<int>(RideNavFilter::COMPLETED));
     activityFilter->addItem(tr("Planned"), static_cast<int>(RideNavFilter::PLANNED));
     int index = appsettings->cvalue(context->athlete->cyclist, GC_NAVDISPLAYFILTER, "0").toInt();
     activityFilter->setCurrentIndex(index);


thejockl:
Nice approach, I like it much better than a separate tree. The following caught my eye:
- The combobox doesn't open cleanly when using the Breeze QStyle on Linux (the burger menu next to it has similar issues: https://bugs.kde.org/show_bug.cgi?id=473735). Maybe you can hardcode Fusion for Linux?
- When using fusion, the highlight color matches the splitters background color, this is a little bit confusing
- The text looks deactivated (both Breeze and Fusion) even when it is not
- When jumping from calendar to activities view and the current type (planned, completed) is hidden, the activity is shown without being visible in the tree. Since this matches the behaviour when using the filter, I would leave it as is.

Fusion:
Bildschirmfoto_20260121_195101
Bildschirmfoto_20260121_195132

Breeze:
Bildschirmfoto_20260121_195322

amtriathlon:
Hi Paul, I like how this looks like and behave on Windows!

paulj49457:
@thejockl Unfortunately I only have a Windows 11 development environment, so if you have some time please update this PR with any corrections required for MacOs or Linux.

amtriathlon:
I tried the ci build on macOS and it works as expected the text is dimmed but perhaps it is better that way, "Completed" is too long and it doesn't look good chopped, may we use "Actual" instead?

thejockl:
"Actual" would work for me - but I am not a native speaker (far from). If we rename it, I should follow that term in the calendar.

paulj49457:
Fine with me, I'll update the PR

amtriathlon:
I will merge this, thanks Paul.
Joachim, if you want to restyle for Linux and to change Completed by Actual, feel free to create other PR.
Once we are done I would trigger a new snapshot since I think this will encourage users to Plan activities without the hassle to deal with them in Activities view.

PR #4812, opened 2026-01-22
Add reconnect attempt timer after BT40Device unexpected disconnect

AnzeBlaBla:
I've implemented retry-reconnection to Bluetooth sensors.

Old behavior after an unexpected disconnect of a sensor was to just try to reconnect once. I've made it so it tries to reconnect every 5s, as this makes more sense - if something happened with the sensor, it probably won't be available immediately (happened during my training rides multiple times, for example having to walk away and losing HR for the rest of the ride).

The reconnecting stops once user presses the "Disconnect" button, which I think is what the user would expect.

### DIFF
--- a/src/Train/BT40Device.h
+++ b/src/Train/BT40Device.h
@@ -23,6 +23,7 @@
 #include <QLowEnergyController>
 #include <QLowEnergyService>
 #include <QQueue>
+#include <QTimer>
 
 #include "CalibrationData.h"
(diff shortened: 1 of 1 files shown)


amtriathlon:
Hi Anže, thanks for your contribution, macOS ci build is failing due to missing `#include <QTimer>`, clang is less forgiving than MSVC and gcc.

PR #4813, opened 2026-01-24
Using "actual activity" instead of "completed activity"

thejockl:
See #4811


thejockl:
In this PR the translation files are unchanged as I am unsure what the preferred way is:

1. Manually update the source strings to keep the translations
2. Recreate the translation files to fully match the source code

amtriathlon:
Thanks Joachim, I think 2) is Ok in this case I will do before generating a new snapshot

PR #4823, opened 2026-02-03
Cloud Service creates unnecessary GlobalContext instances

paulj49457:
Whilst looking at athlete contexts related to the AthleteView PR, I noticed the following code which creates temporary global context instances, when GlobalContext is a singleton.
```
double distanceFactor = GlobalContext().useMetricUnits ? 1.0 : MILES_PER_KM;
QString distanceUnits = GlobalContext().useMetricUnits ? tr("km") : tr("mi");
```
I have updated GlobalContext to hide the constructor so this doesn't occur again, and updated the above code:
```
double distanceFactor = GlobalContext::context()->useMetricUnits ? 1.0 : MILES_PER_KM;
QString distanceUnits = GlobalContext::context()->useMetricUnits ? tr("km") : tr("mi");
```

### DIFF
--- a/src/Core/Context.cpp
+++ b/src/Core/Context.cpp
@@ -33,7 +33,7 @@
 #include <QMutex>
 #include <QWebEngineProfile>
 
-// singleton
+
 static QList<Context*> _contexts;
 
 GlobalContext::GlobalContext()


amtriathlon:
Thank you!

PR #4824, opened 2026-02-10
Fix the WebEngine data path warning

paulj49457:
PR to fix the warning discussed in:  https://groups.google.com/g/golden-cheetah-developers/c/MefPbOsAUns/m/2c_rofoHBgAJ

I have ensured the WebEngine paths are initialised for new Athletes in MainWindow::loadCompleted() , previously they were only initialised for the bootstrap athlete, this has removed the data path warning.

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -1981,6 +1981,11 @@ MainWindow::loadCompleted(QString name, Context *context)
     // clear splash - progress whilst loading tab
     //clearSplash();
 
+    // setup the WebEngine paths
+    QString temp = const_cast<AthleteDirectoryStructure*>(context->athlete->directoryStructure())->temp().absolutePath();
+    context->webEngineProfile->setCachePath(temp);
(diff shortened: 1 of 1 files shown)


amtriathlon:
Hi Paul, thanks your investigating, I introduced this in https://github.com/GoldenCheetah/GoldenCheetah/commit/d18f24c03fe9e9e80c3722f9a4b34f1ed30804a8.
The code for the bootstrap athlete is still necessary or is it covered by your addition to MainWindow::loadCompleted()?
Ale.

paulj49457:
Hi Ale, yes its still needed for the bootstrap athlete, as the bootstrap is loaded differently to all subsequent athletes and doesn't use the openAthleteTab() and then loadCompleted() mechanism.

Once the Athlete Tile PR is complete, I'm thinking of looking at the bootstrap athlete to see whether I can change its loading so it loads the same as all other athletes and not coupled in with MainWindow widget construction, also this would remove the need to put any changes in two places.

amtriathlon:
Thank you!

PR #4815, opened 2026-01-26
Athlete View's tiles functionality

paulj49457:
The following PR updates the Athlete View tiles to provide additional functions and provide summary information on the athletes activities:

- When athletes are created/deleted the tiles often end up overlapped, they required dragging to reposition them, the overlapping has been removed.
- The placement of new athlete tiles (after initial load) given they can be dragged, has been improved.
- A summary of the actual and planned workouts is now included
- The anchor athlete (the one first loaded) now has a padlock to indicate it cannot be closed
- The current athlete (when more than one is open) is now highlighted
- Backup, Delete and saving unsaved files buttons are now available (depending on whether the athlete is loaded, or not)
- The unsaved activities button displays the number of unsaved activities with green/red background
- When more than one athlete has unsaved changes and GC is closed, the save dialog is displayed to warn the user to save the changes, a dialog is displayed for each athlete with unsaved changes, the dialog now displays the athletes name.

Athlete View functionality

Can you please test and provide any feedback?

PR #4827, opened 2026-02-13
Athlete View's tiles functionality, plus removal of bootstrap athlete code

paulj49457:
The following PR builds upon (and includes PR https://github.com/GoldenCheetah/GoldenCheetah/pull/4815) to remove the specific code for the initial bootstrap athlete loaded at start-up. This PR changes the start-up sequence to load the initial athlete in the same way as any other subsequently loaded athlete.

### DIFF
--- a/src/Charts/AgendaWindow.cpp
+++ b/src/Charts/AgendaWindow.cpp
@@ -519,6 +519,14 @@ AgendaWindow::getActivities
 (const QDate &firstDay, const QDate &today, const QDate &lastDay) const
 {
     QHash<QDate, QList<AgendaEntry>> activities;
+    if (! context || ! context->athlete || ! context->athlete->rideCache) {
+        return activities;
+    }
+    const QList<RideItem*> rides = context->athlete->rideCache->rides();
+    if (rides.isEmpty()) {
+        return activities;
+    }
+
     const RideMetricFactory &factory = RideMetricFactory::instance();
     QList<RideMetric const *> rideMetrics;
     QStringList rideMetricNames;
@@ -538,11 +546,15 @@ AgendaWindow::getActivities
     }
 
     int showTertiaryFor = getShowTertiaryFor();
-    for (RideItem *rideItem : context->athlete->rideCache->rides()) {
+    for (RideItem *rideItem : rides) {
         if (   rideItem == nullptr
-            || ! rideItem->planned
-            || rideItem->dateTime.date() < firstDay
-            || rideItem->dateTime.date() > lastDay
+            || ! rideItem->dateTime.isValid()) {
+            continue;
+        }
+        QDate rideDate = rideItem->dateTime.date();
+        if (   ! rideItem->planned
+            || rideDate < firstDay
+            || rideDate > lastDay
             || rideItem->hasLinkedActivity()) {
             continue;
         }
@@ -586,7 +598,7 @@ AgendaWindow::getActivities
             activity.secondaryValues << tr("N/A");
             activity.secondaryLabels << "";
         }
-        if (showTertiaryFor == 0 || (showTertiaryFor == 1 && rideItem->dateTime.date() == today)) {
+        if (showTertiaryFor == 0 || (showTertiaryFor == 1 && rideDate == today)) {
             activity.tertiary = rideItem->getText(getTertiaryField(), "").trimmed();
             activity.tertiary = Utils::unprotect(activity.tertiary);
         }
@@ -598,7 +610,7 @@ AgendaWindow::getActivities
         activity.start = rideItem->dateTime.time();
         activity.type = ENTRY_TYPE_PLANNED_ACTIVITY;
         activity.hasTrainMode = sport == "Bike" && ! buildWorkoutFilter(rideItem).isEmpty();
-        activities[rideItem->dateTime.date()] << activity;
+        activities[rideDate] << activity;
     }
     for (auto dayIt = activities.begin(); dayIt != activities.end(); ++dayIt) {
         std::sort(dayIt.value().begin(), dayIt.value().end(), [](const AgendaEntry &a, const AgendaEntry &b) {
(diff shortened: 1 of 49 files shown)


liversedge:
Woof !

Thats a lot of hard work, quite a scary change too. I will let Ale review etc but thank you !!
This is not easy to do and its the hard yards most folks don't notice.

paulj49457:
@liversedge thank you, if no one notices then mission achieved :) 

As to the scariness of the change, there are many minor bits & pieces to remove context from classes that don't need them, changes to ensure that if context is updated in classes (Search, DataFilter, WorkoutFilter, etc) then any context specific signals are reconnected. The key changes are in AthleteView (see PR https://github.com/GoldenCheetah/GoldenCheetah/pull/4815) and in MainWindow (this PR). I think for these two classes it is better to review the changes by looking at the final changed files, using the diffs as a guide, I've added plenty of comments to explain the changed code.

paulj49457:
Updated to latest master

thejockl:
After performing manual tests (no code review), this PR works mostly as expected for me. No instability or data loss was seen. What was tested:

- Startup behaviour: It seems the last selected open athlete is selected by default. This matches my expectation
- Tile
  - Open: Works as expected
  - Close: Works as expected
  - Show unsaved: Works as expected
  - Save unsaved: Works as expected
  - Settings: Works as expected
  - Shutdown: IMHO not related to a specific athlete, should be removed from the tile
- Toolbar: Could we remove some actions from the Athlete Menu that are now accessible from the tile? I am not a friend of submenus as they often get lost when moving the mouse over to them. Candidates are:
  - Athlete::Open
  - Athlete::Backup
  - Athlete::Delete
  - Maybe they could be replaced by one action: "Manage athletes" which switches to the athletes view
- Chaos didn't show any unexpected behaviour (kill with SIGTERM, SIGKILL with 1 or 2 open users, saved and unsaved activities) - startup always showed the select athlete dialog; no data was lost (despite the unsaved changes)
- Possible extensions for athlete view (maybe later)
  - Additional tile to add a new user
  - Selecting a tile should switch the athlete-tabs, similar to how switching athlete-tabs highlights the corresponding tile

amtriathlon:
My experience is similar for now, no detected issues in my normal workflow, except for the sidebar in Trends view which always starts collapsed: it can be restored but it is collapsed again on the next start.

WRT to shutdown button for the las athlete I think it is overkill, I would keep the close name like for any other athlete, the program terminating when the last tab is closed is common enough (word processors, spreadsheets, bowsers, etc.) not requiring a special warning, IMHO.

PS: I would like to (re)enable the close option in tabs, I see no reason to force the user to switch to Athletes view for that action.

paulj49457:
@amtriathlon & @thejockl  Thank you for the testing and feedback:

I'll update the PR to remove the "shutdown" text for the last open athlete and replace it with "Close" but if it is the last athlete GC will shutdown. 

I don't remember changing the close option in tabs, but I'll take a look to see how it can be re-enable.

I had thought about "_Selecting a tile should switch the athlete-tabs, similar to how switching athlete-tabs highlights the corresponding tile_", but which which region would action this? The left & right mouse buttons are used for dragging the tile on any part of the tile at the moment.

As for the sidebar in Trends view always starting collapsed, I expect this is related to the earlier plan view & trends view sharing the LTM sidebar, I'll investigate.

amtriathlon:
> I don't remember changing the close option in tabs, but I'll take a look to see how it can be re-enable.

It was part of this commit: https://github.com/GoldenCheetah/GoldenCheetah/commit/8150d5b7ce5f4558fb5b3194cbe7d986896aa9a8, but it is an unnecessary restriction now.

paulj49457:
@amtriathlon Thank for the link to the tabbar close commit, and so the latest commit implements:

- Remove "shutdown text"
- Re-enable the tabbar close buttons
- Allow athlete switching via a click on their avatar

Any modification to the athlete main menu can go in a separate PR, as this has been discussed in https://groups.google.com/g/golden-cheetah-users/c/tn5TirOOVdk/m/nmfYkffdBQAJ

paulj49457:
@amtriathlon The last couple of commits are the last, there will be no more (other than bug fixes):

The earlier commit removes duplicate code in the athlete closure and GC quit/close paths,

Closing athlete's via the tabbar & via the Athlete view card both use the same mechanism, then calls closeWindow()

```
CloseAthleteTab()
    checkImportAndUnsaved()
        removeAthleteTab() 
            switchAthleteTab() // update the last open athlete
if (last athlete) closeWindow()
    closeWindow()
         closeEvent()
             // there are no athletes to close here via this route
```

Close via the main menu quit (closeWindow, then CloseEvent) & close via the window decoration X button (CloseEvent) use the path:


```
closeWindow()
    closeEvent()
         // there are athletes to close here via this route, loop round:
        checkImportAndUnsaved()
            removeAthleteTab()  
                switchAthleteTab() // don't update the last open athlete
```


The last commit ensures that when GC is Quit or the window X button is used, the last open (current athlete) is retained for use on the next startup (see comments above), previously it depended on the open athlete's position in the tabbar.

paulj49457:
@amtriathlon The final commit fixes the Trends Sidebar (and other view items) start-up configuration issue.

The problem was restoring the sidebar value from the context, and not the view, possibly a side effect of the restructuring of loading the view & context at different times, this was also an issue for showLowBar & the windows style.

I have removed the pass through interface within AthleteTab, so that code in MainWindow that uses the view(s) is more obvious, as this usually needs to be protected when the AthleteView is displayed (viewStack->currentIndex() = GcViewStackIdx::SELECT_ATHLETE_VIEW). I have added protection to prevent view menu actions when the Athlete Window is displayed, as most of these actions cause changes to the hidden view specific windows.

Only bug fixes now...

paulj49457:
@amtriathlon The latest commit fixes the issue the athlete being closed whilst an activity import is in progress. This protects window closure, athlete menu quit, tabbar athlete closure and close from the athlete card. It modifies the original auto import protection to cover manual imports & cloud imports, basically anything that uses the RideImportWizard.

I'm away now for 3 weeks, so will pickup any other bugs when I return.

amtriathlon:
I am closing this since the functional components were merged by https://github.com/GoldenCheetah/GoldenCheetah/commit/ad84a589206b111f35e83ebebd4f2fe51c963989 and the refactoring part should be a separate PR.

PR #4828, opened 2026-02-16
LTM Sidebar filtering

paulj49457:
Using the planning view with the new calendar I noticed and have fixed the following minor issues:

1) The LTMS sidebar filter, once selected cannot be unselected, so I have changed its selection model from ExtendedSelection to MultiSelection, which supports selecting numerous filters by toggling each on/off with a mouse click, allowing the removal of all filters.

2) The Calendar wasn't combining the LTM sidebar and global filter (filter & home filter), I have updated the code so both can be applied, and the result is the intersection of the filters.

3) When the filter splitter is hidden, the filters remain applied, which I think is confusing, given that if the auto filters are hidden then any auto filter selections are removed, I think the same should apply to the main filter splitter, and I have implemented this.

Please let me know if you have comments/feedback & testing, happy to modify as required.

### DIFF
--- a/src/Charts/CalendarWindow.cpp
+++ b/src/Charts/CalendarWindow.cpp
@@ -1044,10 +1044,28 @@ CalendarWindow::getSummaries
     int numTimeBuckets = firstDay.daysTo(lastDay) / timeBucketSize + 1;
     bool useMetricUnits = GlobalContext::context()->useMetricUnits;
 
-    const RideMetricFactory &factory = RideMetricFactory::instance();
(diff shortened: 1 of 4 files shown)


amtriathlon:
ExtendedSelection is used for intervals too and in several other places, to unselect you have to hold <Ctrl> while clicking, Calendar filtering was fixed by https://github.com/GoldenCheetah/GoldenCheetah/commit/f21808d97c82bc814e570e496192d60ad1e26a2c and the use of hidden filters doesn't seem a problem.

PR #4837, opened 2026-03-07
Using QObject as baseclass for ICalendar

thejockl:
As ICalendar is not a real UI class and is never added to any layout, it overlays the menubar, preventing opening the first two entries. Using QObject instead of QWidget as superclass fixes this issue


amtriathlon:
Great, thank you!

PR #4839, opened 2026-03-13
Addressing crashes reported on the users list

thejockl:
See https://groups.google.com/g/golden-cheetah-users/c/tn5TirOOVdk/m/MKYMpsdMAQAJ
Plus a crash in PlanAdherenceWindow


amtriathlon:
Thanks!

PR #4842, opened 2026-03-14
Plan activity: Allow to enter Workout Code

thejockl:
Similar to "Log activity"
See https://groups.google.com/g/golden-cheetah-users/c/_ljv8-9VnbY/m/NE8mL-qaDAAJ


amtriathlon:
Thank you, I also missed this.

Issue #4831, opened 2026-02-25
Add colors settings in Activity's Overview Tiles


gillesF78:
It would be nice to add settings to customize Tiles's colors in Activity's overview (OverviewItemType.cpp and OverviewItemConfig in OverviewItems.cpp):

- Zone Histogram : colorize bars of the QBarSet with colors (GColorDialog dialog) (e.g. using [this trick](https://stackoverflow.com/questions/58741911/how-to-set-the-color-of-a-bar-in-a-qbarset) ). The result would look like this -->

<img width="235" height="172" alt="Image" src="https://github.com/user-attachments/assets/7f35f7f6-96ff-4c91-8583-26d599740276" />

- PMC Status summary : extend the single indicator to CTL, ATL, & TSB and enable colorization... The result could look like this -->

<img width="292" height="69" alt="Image" src="https://github.com/user-attachments/assets/142edec3-2c39-452e-bd25-2e2dfcce2e52" />

- Metric, KPI, Route, Workout Code : colorize font of the with colors (GColorDialog dialog) (currently, only the background's color is customizable in tile's settings)

<img width="233" height="111" alt="Image" src="https://github.com/user-attachments/assets/e09ef68b-7479-418d-9f62-a0b410421ac9" />

liversedge:
This FR is flawed-- themes are used to configure color choices-- this is important because users change theme e.g. dark to light. So named colors were introduced to enable customisation.

This FR would mean that color choices are embedded in an overview and would therefore have to make assumptions about the type of theme they were rendered with.

PR #4844, opened 2026-03-14
Adding color settings to the activity overview tiles

Rohi0109:
image
image

Took a stab at fixing #4831 by adding both font and bar colors. Let me know if you have any questions!

### DIFF
--- a/src/Charts/Overview.cpp
+++ b/src/Charts/Overview.cpp
@@ -242,27 +242,39 @@ OverviewWindow::getConfiguration() const
             break;
 
         case OverviewItemType::TOPN:
+            {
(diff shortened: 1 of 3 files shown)


liversedge:
The colors you have added are configurable in appearances as named colors (PMC, Power, HR et al)

This is so the UI is consistent across charts, I really don't like this PR or feature request-- it adds unnecessary complexity that goes against the UX design.

liversedge:
I would add that the biggest issue is this PR embeds color choices as config for the overview chart-- these color choices might look good to the user in one theme e.g. dark, but then look terrible when using a light theme.

Named colors were introduced to solve this problem

liversedge:
I appreciate you took the time to code this up, but the original FR is flawed. Sorry !!

PR #4848, opened 2026-03-21
Fix disappearing userchart content when changing the type

thejockl:
See https://groups.google.com/g/golden-cheetah-users/c/_ljv8-9VnbY/m/XAYLlGJJAwAJ

Removing the explicit copy-constructor and operator= from GenericSeriesInfo. Instead setting user1-4 explicitly to nullptr when duplicating a series.

### DIFF
--- a/src/Charts/GenericChart.h
+++ b/src/Charts/GenericChart.h
@@ -83,43 +83,6 @@ class GenericSeriesInfo {
             color("red"), opacity(100.0), opengl(true), legend(true), datalabels(false), fill(false)
         {}
 
-        GenericSeriesInfo(const GenericSeriesInfo &other) {
-            *this = other;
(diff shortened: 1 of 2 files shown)


liversedge:
Very uncomfortable about this-- you added a new feature but removed something foundational because of an issue with your new feature. Have you tested this robustly ?

thejockl:
The copy constructor and operator= was added by me in https://github.com/GoldenCheetah/GoldenCheetah/pull/4841, causing an issue reported on the list. This PR restores the proven status quo in GenericSeriesInfo

liversedge:
Yeah- apologies- I was just checking that.

thejockl:
Highly appreciated that you care so much!

PR #4850, opened 2026-03-28
Reworked the repeat schedule wizard

thejockl:
* Simplified design and concept
* Sticking closer to the original schedule
* Added option to keep gap days (front and back)
* Added option to use originally planned dates
* Removed options to insert rest days or spread same day activities


amtriathlon:
Merged by https://github.com/GoldenCheetah/GoldenCheetah/commit/30a9155b20278d91c42bff4c804b4e7b49a55c69

Issue #4816, opened 2026-01-26
Schedule Export and Import


amtriathlon:
May be as an extension of #4692 it would be useful to be able to export and import training schedules.

A file format candidate would be [iCalendar](https://en.wikipedia.org/wiki/ICalendar) for which we have some support in code via libical, but other formats may be useful.

Since we already have some integration with https://trainerday.com/ an interesting extension would be to be able to import complete training plans.

Also since coaches tend to use Excel or similar for planning some text format (CSV, JSON) could be desirable.

I opened this issue mostly for discussion and contribution.

thejockl:
iCalendar seems to be a very basic format, mostly built upon key value pairs with some standard and many custom keys. My proposal is to use an approach similar to OpenDocument: A zip file with well defined content. This will allow full flexibility and control over everything we want to add to the plan files:
* planned activities
* workout files
* phases and seasons
* description, license, ...

Possible structure within the zip (.gcplan?):
```
/plan.json              # Manifest: metadata & mapping (no payload)
/README.txt             # Human-readable summary (informational only)
/activities/            # RideItem files (one file per planned activity); reuse existing json-format
    day0001-activity01.json
    day0003-activity01.json   # Date relative to anchor date, time as defined in the json
    day0003-activity02.json   # Eventually multiple activities per day
    ...
/workouts/              # Repository of embedded WorkoutFiles; sync with existing TrainDB on import
    5eb63b...abc.mrc
    a3f12c...def.erg
    ...
```

iCalendar could be implemented in parallel but only to sync basic information like upcoming activities with external calendar applications.

thejockl:
Here is a first proposal, generated directly out of GoldenCheetah using a wizard very similar to the repeat schedule functionality. Feedback welcome. Only manual change I made: rename from .gcplan to .zip to allow attaching to this issue

[test1.zip](https://github.com/user-attachments/files/26486529/test1.zip)

amtriathlon:
I think there are parts of #4857 related to this issue which may be interesting to consider.

PR #4857, opened 2026-04-05
Feature/local ai workout api

casmacdo:


liversedge:
I would like to see this added to the cloud api integration rather than as specific to a single provider.

it will need a new type of integration and i can probably help with that.

for now i think developing it this way as a proof of concept is fine.

Is this solely for trainer day?

@amtriathlon  the architecture and intrusive changes in this patch will need looking at- e.g using Python as server endpoints and foundational updates to context look wrong and likely just due to a lack of awareness of the codebase?

src/Core/DataFilter.y
```diff
@@ -49,6 +49,8 @@ extern Leaf *DataFilterroot; // root node for parsed statement
 
 %}
 
+%define api.header.include {"DataFilter_yacc.h"}
```
liversedge: This should not be needed- suspect your tool chain needs fixing

amtriathlon: Agree, that changes should be removed, the solution is to uncomment the following line in gcconfig.pri:
https://github.com/GoldenCheetah/GoldenCheetah/blob/77aea55daf6814dfd18082f026c3477b25b07b1d/src/gcconfig.pri.in#L129-L130

amtriathlon:
> @amtriathlon the architecture and intrusive changes in this patch will need looking at- e.g using Python as server endpoints and foundational updates to context look wrong and likely just due to a lack of awareness of the codebase?

Hi Mark, In general terms I agree it is relevant to provide ways for AI agents to access athlete data and, now we have most of the basic infrastructure to represent detailed plans in place, to generate whole plans or interactive updates seem the killer application. For the whole plan we have #4816 open for discussion.

Model Context Protocol (MCP) seems a fair choice, even when created by Anthropic it seems reasonably open and standardized, I don't know if it is better or it will prevail vs alternatives s.t. A2A (Google), ACP (Linux Foundation), etc.

Using the REST API has the advantage the MCP bridge can be developed independently but the current implementation doesn't allow to access the current context and I don't know if the extension to do that in this PR is the way to go, I would like to know your opinion on this.

An alternative would be to provide a hook to run a Python server from GoldenCheetah using the Python API which is far richer providing access to current context and some modification features. A native server too, but until there is a clear "winner" perhaps is to early for this.

OTOH this PR includes some unrelated features which should be removed and made part of separate PRs, notably auto-pause for Train and GitHub actions for Linux builds.

Ale.

PR #4864, opened 2026-04-23
Fix crash when updating power zones

thejockl:
* Deleting RideCacheRefreshThreads after they finish
* Added option to cancel / interrupt RideCacheRefreshThreads
* Introduced QPointers to prevent use after free
* Canceling all RideCacheRefreshThreads before updating activities from Workout

```
#0  0x000055656d54dbf0 in RideCache::updateFromWorkoutAfter(QDate const&, bool) ()
[Current thread is 1 (Thread 0x7f6c0c0a62c0 (LWP 113812))]
(gdb) backtrace 
#0  0x000055656d54dbf0 in RideCache::updateFromWorkoutAfter(QDate const&, bool) ()
#1  0x000055656d765908 in LTMSidebar::configChanged(int) ()
#2  0x00007f6c12a04d99 in ??? () at /usr/lib/x86_64-linux-gnu/libQt6Core.so.6
#3  0x000055656db757c5 in Context::configChanged(int) ()
#4  0x000055656d4daa2c in Context::notifyConfigChanged(int) ()
#5  0x000055656d83dc2d in AthleteConfigDialog::saveClicked() ()
#6  0x00007f6c12a04ea1 in ??? () at /usr/lib/x86_64-linux-gnu/libQt6Core.so.6
#7  0x00007f6c209cac88 in ??? () at /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6
#8  0x00007f6c12a04ea1 in ??? () at /usr/lib/x86_64-linux-gnu/libQt6Core.so.6
#9  0x00007f6c208f482b in QAbstractButton::clicked(bool) () at /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6
```

### DIFF
--- a/src/Core/RideCache.cpp
+++ b/src/Core/RideCache.cpp
@@ -534,14 +534,35 @@ RideCache::nextRefresh()
     return(returning);
 }
 
+
+RideCacheRefreshThread::RideCacheRefreshThread(RideCache *cache)
+: cache(cache)
+{
+    QPointer<RideCacheRefreshThread> weakSelf(this);
+    connect(this, &QThread::finished, cache, [weakSelf, c = QPointer<RideCache>(cache)]() {
+        if (weakSelf && c) {
+            c->cleanupThread(weakSelf.data());
+        }
+    }, Qt::QueuedConnection);
+}
+
+void
+RideCache::cleanupThread(RideCacheRefreshThread *thread)
+{
+    thread->wait();
+    delete thread;
+}
+
 void
 RideCache::threadCompleted(RideCacheRefreshThread*thread)
 {
     updateMutex.lock();
     refreshThreads.removeOne(thread);
+    bool isLast = refreshThreads.isEmpty();
+    bool cancelled = isCancelled;
     updateMutex.unlock();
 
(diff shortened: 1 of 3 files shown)


thejockl:
This PR runs stable for me for a week now. Will remove the draft-state if it runs stable for one more

amtriathlon:
Hi Joachim, I merged this on my local build and it has been stable too, but I haven’t been experiencing crashes before either.

thejockl:
Hi Ale, also no crash on my machine yet. The crashes occured when updating the zones required adapting the planned indoor workouts values - and this probably could conflict with currently running RideCacheRefreshThreads (->updateFromWorkoutAfter). Here the call to cancel was too late.
IIRC you do most of your workouts outdoor, so this might affect you less than me in cold Germany...
While looking at this issue I could not find any call that deleted the RideCacheRefreshThreads after they finished. This PR tries to add this step to avoid memory leaks and too many threads.

amtriathlon:
Hi Joachim, thanks for the explanation, when you mark this as ready for review I will merge it and trigger new snapshot builds for better testing.

PR #4874, opened 2026-05-17
Support chart import to any relevant view

paulj49457:
The imported chart(s) are not checked against the chart relevance defined in GcWindowRegistry.cpp, the code only checks whether the type is Plan & Trends from the imported chart:

https://github.com/GoldenCheetah/GoldenCheetah/blob/980ef45541b19c36017931d2d07243e0dc6956af/src/Gui/Perspective.cpp#L1787

Therefore, it is possible to import the Agenda chart to the Trends view even though its relevance is set for only the Plan view.

This PR will display in the Import chart Combox all possible views relevant for a chart, and default to the view from which the chart was original saved (the type filed in the chart's file).

I have also added the Plan View to the Web Charts relevance, and ensured the windowsForType() & idsForType() implementations are members of the GcWindowRegistry class.

I'm not sure why there is a limitation when importing charts when GC_HAVE_ICAL is not defined, I have removed this restriction, but it could be wrong?
https://github.com/GoldenCheetah/GoldenCheetah/blob/980ef45541b19c36017931d2d07243e0dc6956af/src/Gui/Perspective.cpp#L1778

### DIFF
--- a/src/Gui/GcWindowRegistry.cpp
+++ b/src/Gui/GcWindowRegistry.cpp
@@ -110,7 +110,7 @@ GcWindowRegistry::initialize()
     { VIEW_TRAIN, tr("Workout Editor"),GcWindowTypes::WorkoutWindow },
     { VIEW_TRAIN, tr("Live Map"),GcWindowTypes::LiveMapWebPageWindow },
     { VIEW_TRAIN, tr("Elevation Chart"),GcWindowTypes::ElevationChart },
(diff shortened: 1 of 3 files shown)


amtriathlon:
> I'm not sure why there is a limitation when importing charts when GC_HAVE_ICAL is not defined, I have removed this restriction, but it could be wrong?

No, that is a remora from the time when Diary View was enabled only win libical was included in the build, it is Ok to remove.

PR #4877, opened 2026-05-25
Updated the RideNavigators styling (analysis view)

thejockl:
* Updated design with gray top-bar for completed, orange for planned
* Improved painting (no more drawing from delegates outside their rectangle)
* Configurable number of calendartext lines (0-5) in Options / Data Fields / Fields / Summary rows in activities list
* Added icons similar to calendar (same coloring mechanics)
* Added tooltip
* Showing dirty activities in italics
* Added isDirty-DataFilter


paulj49457:
I think we may need to reconsider the use of italics within the overview tiles (its used to show user override of the values), if its now widely used for the planning calendar and ride navigator dirty activity indication.

thejockl:
@paulj49457 , I decided to use italics because it doesn't occupy more space but is distinct from other text. Probably the same reason why you did the same. What do you think about using a lighter weight for the overrides? BTW, I love your editing-feature on the tiles, this made me use notes for every ride :)

paulj49457:
> @paulj49457 , I decided to use italics because it doesn't occupy more space but is distinct from other text. Probably the same reason why you did the same. What do you think about using a lighter weight for the overrides? BTW, I love your editing-feature on the tiles, this made me use notes for every ride :)

@thejockl I'll have think, as I remember there some other proposals to indicate the metric overridden values on the tiles, I'll find the thread and see what other alternatives were suggested.

amtriathlon:
Hi Joachim, I merged your PR in my local build and started to use, I like it so far but I have found some inconsistencies between the Navigator and the Calendar:

image
image

As you can see in Calendar the colors are used for the background and on the Navigator for the foreground, and the sport code is normalized via RideFile::sportTag in Calendar but not in Navigator (I use a non-standard name for Gym workouts).

BTW, the original coloring for the navigator allowed to choose between foreground/background for the coloring as explained in https://github.com/GoldenCheetah/GoldenCheetah/wiki/UG_Preferences_Data-Fields#Colour-keywords, looking at the code the checkbox is still there but likely in some reformatting it was not displayed anymore.

Perhaps we could re-add it and use the setting consistently for both Calendar and Navigator.

Currently I use the colors for Sports which is redundant (or a reinforcement if you like), but I am thinking to use it for RPE to get a view of easy/moderate/hard sessions at a glance so keeping the coloring independent of Sport can be useful.

That's all for now, when you think this is ready we could merge and generate a new snapshot to get more feedback from users.

Cheers, Ale.

PS1: this was the commit adding the fg/bg checkbox: https://github.com/GoldenCheetah/GoldenCheetah/commit/a2b128519b9bc85589e3f4cc55b829b2bc600141, I don’t know why it is not displayed anymore.
PS2: now I saw you removed the setting in this PR.

thejockl:
Hi @amtriathlon , I restyled the list so it matches the calendar better, especially day and week view:
Bildschirmfoto_20260602_154209

The calendars month view currently doesn't use the color for activities background as I wanted to avoid visual overload as this can become quite dense. But let me check how it would work visually.

On the sport: Unfortunately - other as in the calendar - I don't have the RideItem available in this view, but only the RideCacheModel. Maybe we could add a dedicated role similar to planned, dirty, ... to make the normalized sport available?

The idea with the coloring according to RPE sounds interesting, lets start a discussion on the list what needs to be done and what it means, both for the RideNavigator and the calendar.

Cheers, Joachim

PS.: I changed to draft PR last week as I had forgotten to activate drag&drop from the list, but this is now back. Additionally I scaled down some paddings as it was to "empty" for me when really using it.

amtriathlon:
It looks much better now, Joachim, thank you!

There is a static method to normalize sport, this change works for me:
```
diff --git a/src/Gui/RideNavigator.cpp b/src/Gui/RideNavigator.cpp
index 9a7aebd09..910a1daea 100644
--- a/src/Gui/RideNavigator.cpp
+++ b/src/Gui/RideNavigator.cpp
@@ -1548,7 +1548,7 @@ ActivityTreeView::paintRow
     const bool selected = selectionModel()->isSelected(index);
     const bool isHeader = useIndex.data(GroupByModel::HeaderRole).toBool();
     const bool planned = useIndex.data(GroupByModel::PlannedRole).toBool();
-    const QString sport = useIndex.data(GroupByModel::SportRole).toString();
+    const QString sport = RideFile::sportTag(useIndex.data(GroupByModel::SportRole).toString());
     const QString subSport = useIndex.data(GroupByModel::SubSportRole).toString();
     const int lineHeight = fm.lineSpacing() + 1;
     const bool dark = GCColor::isPaletteDark(palette());
```

Cheers, Ale.

thejockl:
From my perspective this PR is complete. Only remaining topic is the default for visible lines for calendartext. Current default is 3, I settled however with a more compact 2-line setup

amtriathlon:
Two lines works better for me too, I will merge and trigger snapshot update, thank you!

thejockl:
Perfect, the PR is updated

PR #4879, opened 2026-05-26
Calendar: Set the minimum number of visible minutes per activity

thejockl:
In day and week view, short activities might not show any data in their box. This change adds an option to define the minimum number of visible minutes per activity.


amtriathlon:
Thank you!

PR #4883, opened 2026-06-09
Fix wrong icons in RideNavigator

thejockl:
See https://groups.google.com/g/golden-cheetah-users/c/RVUs2FbTdPI/m/FAhSjqzGAQAJ
Source of the issue: Sport and SubSport are read from RideCacheModel
(via GroupByModel). To find the correct column in those models, the
metric name is used - but the model can have either translated or
original name as column title (this is the rootcause of the sometimes it
works, sometimes not).
RideNavigatorProxy doesn't have the translations of the column names
available, therefore using an empty string instead of SubSport. Sport
can have the same issue if translated.
    
This PR adds the untranslated, technical names (e.g. time_in_zone_L1)
using the role Qt::UserRole + 1 to the headerData and uses these values
for the icon lookup

### DIFF
--- a/src/Core/RideCacheModel.cpp
+++ b/src/Core/RideCacheModel.cpp
@@ -166,12 +166,21 @@ RideCacheModel::endRemove(int)
 bool 
 RideCacheModel::setHeaderData (int section, Qt::Orientation orientation, const QVariant &value, int role)
 {
-    if (orientation == Qt::Horizontal && role == Qt::EditRole) {
-        headings_[section] = value.toString();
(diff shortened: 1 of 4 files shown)


thejockl:
This PR currently only fixes the new issue on RideNavigator. Should we use the same pattern (Qt::UserRole + 1) for all column-finding in RideNavigatorProxy.h (see https://github.com/thejockl/GoldenCheetah/blob/ac06c5666d7449a2fd55b0ec94c562056d994b6b/src/Gui/RideNavigatorProxy.h#L147)?

PR #4888, opened 2026-06-22
HrZones implements getZoneDescriptions() as Zones already does

peret2000:
Needed by HTML chart, as well as for completeness.
It is a small change used by a next PR for HTML charts, but I think it is better to isolate it, as it does not belong to HTML but to old stable code. It keeps consistency with power Zones class, and avoids to repeat code if you are interested in these HR zones, as it is the case for HTML harts


peret2000:
Thank you!

PR #4897, opened 2026-07-10
Refactor RealtimeData derived series

amtriathlon:
Started with W'bal and VAM moved from TrainSideBar to a new RealtimeDataSession class derived from RealtimeData with an updateDerived method inteded to do the computation before telemetry update is generated and start/stop/newLap methods for the necessary resets of accumulation members.
This is still WIP, the next step would be to move (copy then remove after testing) Coggan/Skiba metrics, Joules, HeatLoad and averages/lap averages from DialWindow so they become available for the new HTML Chart and Overlay Widgets too.

### DIFF
--- a/src/Train/RealtimeData.cpp
+++ b/src/Train/RealtimeData.cpp
@@ -19,6 +19,7 @@
 
 #include "RealtimeData.h"
 #include "RideFile.h"
+#include "Context.h"
 
 #include <QtDebug>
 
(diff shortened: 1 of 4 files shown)


amtriathlon:
Both start/stop seem redundant, I think it is better just to re-create the session object for a full reset on session start/stop, I will change that.

amtriathlon:
This PR was merged accidentally when it was not ready, rebased and replaced by https://github.com/GoldenCheetah/GoldenCheetah/pull/4899, sorry.

PR #4903, opened 2026-07-24
Calendar: Added time configuration for new measures

thejockl:
See https://groups.google.com/g/golden-cheetah-users/c/cB27wvapGJk

Additional: Added german translation for the dialog label

### DIFF
--- a/src/Charts/CalendarWindow.cpp
+++ b/src/Charts/CalendarWindow.cpp
@@ -294,6 +294,7 @@ CalendarWindow::CalendarWindow(Context *context)
 
     calendar = new Calendar(QDate::currentDate(), static_cast<Qt::DayOfWeek>(getFirstDayOfWeek()), context->athlete->measures);
 
+    setMeasureTime(QTime(6, 30, 0));
     setStartHour(8);
     setEndHour(21);
     setMinVisibleMins(0);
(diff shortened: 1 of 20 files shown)


amtriathlon:
Hi Joachim, I built this on Windows and Measures Times doesn't seem to be used or saved.
I changed 6:30 to 8:00 in chart config, the change is preserved when re-opening chart config, but Add Measure still shows 6:30 and after restart config goes back to 6:30

thejockl:
WIll check later today (after a short ride)

amtriathlon:
Thank you, it is weird since the code looks ok. I will try a full rebuild just in case.
PS: after a full rebuild the behavior is the same, also using the AppVeyor generated AppImage.

thejockl:
Issue was: Properties of type QTime weren't handled (serialized / deserialized) yet, added this functionality
@amtriathlon, I hope I caught all required locations but am unsure. A review would for sure be beneficial

amtriathlon:
Hi Joachim, I think all required locations are included, thank you!

BTW, I think Calendar should be blocked for CloudDB upload like other non-programable charts in:
https://github.com/GoldenCheetah/GoldenCheetah/blob/4298eeca061bae818d78c13e341b3ddb3b50e958/src/Charts/GoldenCheetah.cpp#L1099-L1109

I will do that before the next build.

thejockl:
> Hi Joachim, I think all required locations are included, thank you!
> 
> BTW, I think Calendar should be blocked for CloudDB upload like other non-programable charts in:
> 
> https://github.com/GoldenCheetah/GoldenCheetah/blob/4298eeca061bae818d78c13e341b3ddb3b50e958/src/Charts/GoldenCheetah.cpp#L1099-L1109
> 
> I will do that before the next build.

Thanks, wasn't aware of this. Probably Agenda and PlanAdherence should also be added then

PR #4909, opened 2026-08-08
Optionally download default icons during upgrade

thejockl:
Offer to download the default icon set if last version is below 3.8RC1 and no icons are already installed


amtriathlon:
Hi Joachim, this works as expected for me, the only change I would do is to replace "Version 3.8 uses icons for sports and subsports" by "GoldenCheetah uses icons for sports and subsports", or "From v3.8 GoldenCheetah uses icons for sports and subsports" if you prefer, so this remains valid for next versions. Thank you!

amtriathlon:
Thank you!

PR #4912, opened 2026-08-19
Support for Kettler Golf E and other old Kettler trainers requiring uppercase serial commands

wangnick:
This pull request adds support for Kettler Golf E and other old Kettler trainers requiring uppercase serial commands. In order not to break the existing trainer base, it does so by sending the original commands and determining from the reply whether to keep or change the approach. It also robustifies the serial communication with the Kettler trainers during discovery and initialization in general.

The changes were discussed with the maintenance crew in https://groups.google.com/g/golden-cheetah-users/c/Hhsd5_bMPYE.


amtriathlon:
Thank you!

PR #4915, opened 2026-08-30
New Athlete - Use GlobalContext for Units

### LINKED ISSUE #4914: Creating a new athlete can change the global units
The following steps illustrate the issue, the code in NewAthleteWizard.cpp seems to be the problem:

i) Set GC to use imperial units
ii) Shutdown GC to ensure the imperial units (GC_UNIT) is written out to QT
iii) Open GC, create a new athlete with metric units for weight/height (this changes GC_UNIT, but leaves GC in imperial units)
iv) Shutdown and restart GC, it now read GC_UNIT on start-up, so GC is in metric units

I'm not sure why creating an athlete can change the global units: https://github.com/GoldenCheetah/GoldenCheetah/blob/cacc0d723b8267d33c39c2f4c6837db4e081a392/src/Gui/NewAthleteWizard.cpp#L108

amtriathlon:
- Honor metric/imperial setting when starting the wizard
- Notify metric/imperial setting for the (very) edge case of a user changing units while creating a new athlete while others are open. Fixes #4914


paulj49457:
Tested it, all good, thank you.

PR #4917, opened 2026-09-04
Always adding workouts-directory to gcplan bundles

thejockl:
Before: Import always required both planned and workouts directories to be present while export skipped workouts if empty.
Now: Export always adds a workouts directory, even if empty

See issue reported in https://groups.google.com/g/golden-cheetah-users/c/aYpDrwkO1eo


amtriathlon:
Thank you!

PR #4918, opened 2026-09-06
Global Named Searches file

paulj49457:
Currently named searches (filters) are defined per athlete (<athlete>/config/namedsearches.xml), but working with multiple athletes this means that searches have to be entered or amended multiple times to keep them consistent across all athletes.

This PR creates a single global named searches definitions that can be used by all athletes (in the same way as metadata).

To aid transition, at first start-up all the athlete's current filters are merged into the new single global definition file, and from this point only the global namedSearches.xml file is used. It is an easy one off activity for the user to prune out any duplicates using the manage filters dialog, if you have a single athlete there is nothing to do.

**Changes:**

- removed NamedSearches from the Athlete class, and related slots and signals
- NamedSearches is now a singleton
- namedsearches.xml file is now in gcroot directory, all existing athlete <athlete>/config/namedsearches.xml are left in place and unmodified.
- Adds a new signal to GlobalContext ( void namedSearchesChanged() ) to notify interested components the searches have changed.
- The NamedSearches class now manages writing and updating of the namedSearches.xml file as each change occurs.

**Remove unused items:**

removed unused _count_ from NamedSearch, as it was never written out to the namedsearches.xml file.
removed unused _loadcount_ in NamedSearchParser, as it never present in the namedsearches.xml file, so never parsed.

### DIFF
--- a/src/Core/Athlete.cpp
+++ b/src/Core/Athlete.cpp
@@ -147,9 +147,6 @@ Athlete::Athlete(Context *context, const QDir &homeDir)
     autoImportConfig = new RideAutoImportConfig(home->config());
     autoImport = NULL;
 
-    // Search / filter
-    namedSearches = new NamedSearches(this); // must be before navigator
-
     // Metadata
     rideCache = NULL; // let metadata know we don't have a ridecache yet
 
@@ -253,7 +250,6 @@ Athlete::~Athlete()
     delete davCalendar;
 #endif
 
-    delete namedSearches;
     delete routes;
     delete seasons;
     delete measures;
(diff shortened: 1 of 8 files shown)


amtriathlon:
Hi Paul, it looks fine at first sight, but already existing filters should not be append when consolidating per athlete filters, otherwise a coach with just 5 athletes will end up having 15-25 filters without having created a single one, it seems annoying.

paulj49457:
Hi Ale, given it was a one off combination of the athlete files to the global file I thought it'd be ok, but no problem, I'll add some code to compare the searches and prevent any duplicates,

